PR #85, opened 2015-02-25
Allow additional conventions in CF Conventions attr, fixes #82

### LINKED ISSUE #82: CF checker doesn't allow additional conventions in Conventions attribute
The test `check_conventions_are_cf_16` in cf.py requires the **Conventions** global attribute to be exactly equal to the string "CF-1.6". This is indeed what the [CF conventions document](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#identification-of-conventions) _recommends_. However, the [NetCDF User's Guide](http://www.unidata.ucar.edu/software/netcdf/docs/netcdf.html#Attribute-Conventions) allows this attribute to be a blank- or comma-separated list of convention names. 

Could the check be modifed to look for "CF-1.6" as a sub-string of the **Conventions** attribute?

E.g. [IMOS](http://imos.org.au/) data will usually have Conventions = "CF-1.6,IMOS-1.3". We would like these files to pass the CF checker.

daf:
Split on comma and check if any match.  Adds tests/test nc files.


dpsnowden:
This is splitting on both commas and spaces, right? #82 explains CF recommendations.

daf:
Yeah it doesn't.  Will address.

PR #93, opened 2015-03-13
space-time none coordinate variable should not be checked in

fxmzb123:
check_independent_axis_dimensions.
The check has been changed to check:
1) whether variable dimension is a space-time dimension
2) whether variable dimension has associated coordinated variable.
Issue 72

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -1353,6 +1353,7 @@ class CFBaseCheck(BaseCheck):
                     valid =True
                 else:
                     valid = False
+                    dim_name = each
                     break
 
             if valid == False :
@@ -1364,6 +1365,8 @@ class CFBaseCheck(BaseCheck):
                 ret_val.append(Result(BaseCheck.MEDIUM, \
                                valid, \
                                ('var', name, 'check_independent_axis_dimensions')))
+        print ret_val
+        
         return ret_val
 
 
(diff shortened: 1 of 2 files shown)


compliance_checker/cf/cf.py
```diff
@@ -1340,7 +1346,7 @@ def check_independent_axis_dimensions(self, ds):
                 space_time_coord_var.append(each._name)
 
         #Find all all space-time variables that are not coordinate variables
-        space_time_non_coord_var=[]
+        ''' space_time_non_coord_var=[]
```
daf: Instead of commenting the block, just delete it.

daf:
Hi @fxmzb123, thanks for the patch.  There is a test failure (`NameError: global name 'dim_name' is not defined`) which is the the invalid handler path - you had removed the dim_name setting line from the else branch.  You can use `each` in the error message in place of it, but I think that will be invalid if we move the checker to Python 3.

Then, after fixing this error locally, the test still fails as I think it is depending on the old logic.  Can you look at the test and see if it needs to be updated too?  Thanks.

fxmzb123:
Hi Dave,

Thank you for the comment.

I have removed not used code, fixed the missing dim_name variable and changed related test. I have also changed the check logic a little bit as the logic was not quite right.

Cheers,
Ming

lukecampbell:
Is this ready to go?

lukecampbell:
@daf

daf:
@lukecampbell will review today

PR #94, opened 2015-03-13
Added support to allow auxiliary coordinate variable to have axis

fxmzb123:
Added support to allow auxiliary coordinate variable to have axis attribute.
Issue91


daf:
I'm reviewing this now.  There's a code path that will give an error (tests, nor the couple of files I have to test with don't trip it).  I have a commit building on top of this that I will put in a new PR (superseding this one).

Issue #90, opened 2015-03-03
CF checker doesn't allow valid variables without units attribute


mhidas:
The `check_units` test in cf.py returns a high-priority fail for variables without a _units_ attribute. Some exceptions are allowed for, but a couple of important ones are not:
 1) "Units are not required for dimensionless quantities" ([CF](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#units)), and
 2) It doesn't make sense for quality-control variables to have units, even if they are represented as numeric flags.

It's mainly the second case that is giving us errors for every IMOS file, where most data variables have an associated quality control flags variable of type byte.

Proposed solutions:
 1) Allow for more exceptions to this rule, e.g. exempt any variable with a _flag_values_ attribute. This will resolve the second issue above, but not the first.
 2) Make the 'check_units' test return a medium- or low-priority fail for variables that don't have a _units_ attribute. If the attribute does exist, the other checks on it should remain high-priority.

benjwadams:
I ran into this today running the CF checker.  I noticed that the CF checker attempts to look for "modifiers" in standard names -- see  http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#standard-name-modifiers

In the case of flags, it's a `status_flag` separated by a space at the end of the standard name.  You'll notice that `status_flag` has no units in the table provided in Appendix C.  The checker picks up the modifier, so yet another method would be to look for a `status_flag` modifier, at least in the case of flag variables.

mhidas:
@benjwadams Good point. The limitation of that approach is that it only works for variables that have an appropriate CF standard_name. Unfortunately for us this is not always the case. On the other hand, all quality flag variables should have a `flag_meanings` attribute, to explain what the flags mean. So I think the solution in #96  is good (we just need to get the unit test right...)

mhidas:
Fixed in #102.

PR #96, opened 2015-03-16
Added support to skip quality control variables in check units.

fxmzb123:
Added support to skip quality control variables in check units.
Set medium priority fail for variables that don't have a unit attribute in check units.
Issue-90


mhidas:
I just made a minor tweak (to quality flag variables that use either flag_masks or flag_values... both of which should have flag_meanings).
This fixes #90 for me. Please merge if you're happy with this solution.

mhidas:
The unit test is not quite right (need to add a quality control variable without a units attribute). @fxmzb123 will update it again soon, then this should be ready to merge.

compliance_checker/cf/cf.py
```diff
@@ -491,11 +491,15 @@ def check_units(self, ds):
...
+            
+            # skip quality control vars
+            if hasattr(v, 'flag_meanings'):
+                continue
+            
             units = getattr(v, 'units', None)
 
             # 1) "units" attribute must be present
-            presence = Result(BaseCheck.HIGH, units is not None, ('units', k, 'present'))
+            presence = Result(BaseCheck.MEDIUM, units is not None, ('units', k, 'present'))
```
daf: I don't think this priority reduction is appropriate.  The [CF document states](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#units):

> The units attribute is required for all variables that represent dimensional quantities...

The use of "required" in this phrase means it is high priority.

I do agree that we need to better identify the variables which this phrase applies to.

mhidas: @daf I guess that's reasonable. Reliably identifying variables which legitimately don't have units because they're dimensionless will be tricky... but maybe can be dealt with separately.

mhidas:
@daf I've changed the priority back to HIGH and also fixed the unit test. Hope this can be merged now.

lukecampbell:
Variables that have "status_flag" in the standard name do not require units. I got into a heated argument about this and emailed @rsignell-usgs and he confirmed it, quality flags do not require units.

PR #99, opened 2015-04-13
Added a simple plugin system for checkers. Issue-60

fxmzb123:


daf:
This is a great start.  I'd like @lukecampbell to take a look too, and I'll research a bit about the proper way to load from external packages (namespace packages is the key I think), extending from this technique.  Will be looking at it this week.  Thanks!

compliance_checker/suite.py
```diff
@@ -18,6 +18,36 @@ class CheckSuite(object):
...
+        else:
+            for sub_class in sub_classes:
+                if hasattr(sub_class, "name") and hasattr(sub_class, "register_checker"):
+                    if sub_class.register_checker:
+                        if sub_class.name not in cls.checkers:
+                            cls.checkers[sub_class.name] = sub_class
+                cls.get_all_sub_classes(sub_class)
+
+    @classmethod
+    def load_all_avaiable_checkers(cls):
```
dpsnowden: Typo here.  Available not avaiable.

compliance_checker/suite.py
```diff
@@ -18,6 +18,36 @@ class CheckSuite(object):
...
+    @classmethod
+    def get_all_sub_classes(cls, parent_calss):
+        """
+        Helper method to retrieve all sub checker classes. In order to register a
+        checker class, the class must have both name and register_checker attributes.
+        The name is used for the test option in the command. The register_checker is
+        a flag to indicate whether the checker is required to be registered.
+        """
+
+        sub_classes = parent_calss.__subclasses__()
```
dpsnowden: Typo here.  parent_class not parent_calss

danfruehauf:
Will it be possible to specify more directories to load checkers from?

@mhidas I think that's an important requirement for us.

daf:
@danfruehauf I've got a proof of concept plugin system working locally, I think this is the correct path.  You'll be able to build your own check suites living in their own python packages.  I need to continue to flesh this out and make sure it's proper, but it's looking good.

danfruehauf:
@daf I see. Ideally for us we'd like to have sporadic files which define checks and we'd like to load them dynamically. Going through the hoops and loops of python packages is pretty unnecessary for this feature I believe. Eventually it's just a python class which implements an interface - and we might have many of those.

daf:
@danfruehauf I'll see if I can find a shortcut that will help you in this case.  I plan on providing a sample package that you can just clone and adjust for the package approach (which absolutely has to happen anyway), but may be able to bend the mechanism to allow you to specify a path to an unpackaged file too.

mhidas:
@daf Any progress on this? If you have something that works, we'd be keen to test it.

@fxmzb123 @lbesnard @dnahodil

daf:
@mhidas I have a few kinks to work out still but will get to it early this week.  Sorry for the holdup.

daf:
Closing in favor of #173

PR #112, opened 2015-05-07
Fix travis.

ocefpaf:
This PR changes the test call from `py.test` to `python setup.py test`, somehow `py.test` was failing.  My guess is because it was trying to download all the dependencies from PyPI (and `cf_units` is not there yet).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,4 +14,4 @@ before_install:
     - source activate test
 
 script:
-    - py.test
+    - python setup.py test


daf:
Weird.  The errors yesterday ([build 2](https://travis-ci.org/ioos/compliance-checker/builds/61529710)) appeared to be the earlier `conda install -r requirements.txt` line, which was failing because it couldn't find the cf_units package.  Maybe a package for that became available between yesterday and today?

I don't see how this PR fixes the build from yesterday, but since Travis now seems to be happy...

ocefpaf:
The first error was because cf_units was not available on ioos yet. I
designed the Travis test to work with udunitspy (I was not expecting the
other PR to get merged so quickly ;-).

Then Travis complained about dateutils and Wicken. I guess that py.test was
trying to create an env for the tests instead of using the conda setup.
Apparently python setup.py test does not do so.
Em 07/05/2015 10:47, "Dave Foster" <notifications@github.com> escreveu:

PR #125, opened 2015-06-18
Fixes ACDD Checks and adds explicit ver of pb

lukecampbell:
- Petulant Bear still had the unicode issues if they are present in
  datasets, this commit sets the version explicitly to the (now) newest
  release.
- ACDD had a few bugs in it where the messages in the Results objects
  were strings instead of lists or iterables, which resulted in an
  exception being raised later.
- Basically, if you were checking ACDD, some test failures would result
  in checker crashing.

### DIFF
--- a/compliance_checker/acdd.py
+++ b/compliance_checker/acdd.py
@@ -44,7 +44,7 @@ class ACDDBaseCheck(BaseCheck):
             'institution',
             'project',
             'processing_level',
-            'acknowledgement',
+            'acknowledgment',
             'geospatial_lat_min',
             'geospatial_lat_max',
             'geospatial_lon_min',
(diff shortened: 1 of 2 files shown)


lukecampbell:
@daf

compliance_checker/acdd.py
```diff
@@ -69,7 +69,7 @@ def check_recommended(self, ds):
     def check_suggested(self, ds):
         return [
             'contributor_name',
-            ('contributor_role', ['principalInvestigator', 'author']),
+            'contributor_role',
```
daf: This former syntax was to say that not only must `contributer_role` need to exist, it needed to be one of those two values.  I got this from some document at one point as it was the requirements. Has it changed?

lukecampbell: http://www.unidata.ucar.edu/software/thredds/current/netcdf-java/metadata/DataDiscoveryAttConvention.html#contributor_role_Attribute

lukecampbell: It was changed, not sure when. There was a mailing list thing and @rsignell-usgs was on it and may have driven the change. Basically the discussion was about how everyone used it differently and it just needed to be comma separated.

daf: Looks like it came from http://wiki.esipfed.org/index.php/Attribute_Convention_for_Data_Discovery_1-1
I feel like it was in 1-2, which I also think this checker was based on.  It's cool.

abirger: @daf , @lukecampbell , Which ACDD version is checker currently checking against? 1.1, 1.2, or 1.3?

daf:
Need to put petulant-bear on binstar for travis to be happy.  Did @ocefpaf do that before?  Is it in ioos-recipes?

compliance_checker/acdd.py
lukecampbell: 1.2, I didn't know 1.3 was released yet, but we should update the compliance checker to support 1.3

daf:
Looks like petulantbear showed up in binstar, so Travis became happy after I re-ran.

ocefpaf:
Sorry @daf.  I forgot to ping you 10 days ago!

daf:
Not a problem @ocefpaf !

Issue #70, opened 2014-10-10
Amount of data within file affects run time


mbiddle-bcodmo:
This isn't an error but more of a question. What is actually causing the script to run soo slow when the data within the file is increased? For example, I have one file that is 325MB, which has been reduced down to 41KB by clearing out some (a lot) of the data points. All of the variables and attributes have stayed exactly the same, only the dimensions and amount of data has changed. When I run the CF check on the smaller (41KB) file, it completes in a matter of seconds. However, when I run the CF check on the larger (325MB) file, it takes forever to run (I'm on hour two of runtime right now). Does the checker actually need to pull in each individual data point within the file? Isn't it just checking variable attributes and global attributes? FYI, I'm running with both files on my local system...

danfruehauf:
We are encountering the same problem with gridded files. It takes a while to load a ~100MB file into memory. Depending on filesystem IO it can be ~1-2 minutes per file. In other words - too slow.

I would suggest to add a command line argument which will load only attributes and variables without the whole data. This can imply that checks that rely on testing all the points in a data set will not run. On the other hand, it'll be much quicker for 80% of the tests (at least for us) which just verify existence of attributes and variables.

@lbesnard

mhidas:
@fxmzb123 is looking into this issue.

mhidas:
@fxmzb123 The two files in [this catalog](http://thredds.aodn.org.au/thredds/catalog/IMOS/eMII/checker_test/SRS/SST/catalog.html) are an example of this issue.

The first is 66Mb gridded dataset. The CF checker takes almost 2 mintues to check it.
The other has exactly the same attributes and structure, just  much less data. CF checker runs in just over a second.

daf:
This is not exclusive to CF checks, ACDD confirms that time extent metadata matches the actual variable data which involves a min/max check on the time variable.

DanielJMaher:
@daf @lukecampbell @kwilcox 

Do we want to wrap this up into #202 and make a more intensive system for skipping tests and not accessing datasets during tests?

daf:
It could be a similar solution to what I proposed in #202 with a decorator:

```python
@accesses_remote_data()
def check_some_remote_data_stuff(self, ds):
...
```

and control the running of those via command line flags (`--allow-remote-data=True` or similar).  It should probably produce skips in the output a la unit tests.

lukecampbell:
yeah that would be good!

kwilcox:
[...]

PR #129, opened 2015-06-24
Reimplemented the logic in check_packed_data, fixes #128

### LINKED ISSUE #128: Unnecessary data shuffling in CF check_packed_data
While investigating the causes of issue #70 (slow-running checker for large files) @fxmzb123 found that the `check_packed_data` method (in CF suite) was responsible for a lot of the running time. The current implementation calls `numpy.reshape` on each variable 5 times, which can be pretty slow for large multi-dimensional arrays. This could certainly be optimised by doing the reshape only once for a variable.

Going further, I think that reshaping data arrays is actually unnecessary here. The CF convention only refers to the data types of the variable itself, and some of its attributes. Surely these can be determined without manipulating the entire array.

fxmzb123:
This is the branch against issue 128.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -2368,7 +2368,7 @@ class CFBaseCheck(BaseCheck):
             if not (add_offset or scale_factor):
                 continue
 
-            valid = False
+            valid = True
             reasoning = []
 
             # if only one of these attributes is defined, assume they
@@ -2380,17 +2380,19 @@ class CFBaseCheck(BaseCheck):
 
             if hasattr(add_offset, 'dtype') and hasattr(scale_factor, 'dtype'):
 
-                if var.dtype == add_offset.dtype == scale_factor.dtype:
-                    valid = True
-                elif add_offset.dtype == add_offset.dtype and scale_factor.dtype != var.dtype:
-                    # Check both attribute types are type float or double
-                    if add_offset.dtype in [np.float, np.float16, np.float32, np.float64, np.float128]:
-                        # Check variable types must be byte, short or int
-                        if var.dtype in [np.int, np.int8, np.int16, np.int32, np.int64]:
-                            valid = True
-
-            if not valid:
-                reasoning.append("'add_offset' and 'scale_factor' are not both of type float or int or the data variable is not of type byte, short, or int.")
+                if add_offset.dtype != scale_factor.dtype:
+                    valid = False
+                    reasoning.append("Attributes add_offset and scale_factor have different data type.")
+                elif scale_factor.dtype != var.dtype:
+                    # Check both attributes are type float or double
+                    if not scale_factor.dtype in [np.float, np.float16, np.float32, np.float64, np.float128]:
+                        valid = False
+                        reasoning.append("Attributes add_offset and scale_factor are not of type float or double.")
+                    else:
+                        # Check variable type is byte, short or int
+                        if not var.dtype in [np.int, np.int8, np.int16, np.int32, np.int64]:
+                            valid = False
+                            reasoning.append("Variable is not of type byte, short, or int.")
 
             result = Result(BaseCheck.MEDIUM, valid, ('var', name, 'packed_data'), reasoning)
             ret_val.append(result)
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -2363,62 +2363,74 @@ class CFBaseCheck(BaseCheck):
         ret_val = []
         reasoning = []
         for name, var in ds.dataset.variables.iteritems():
-            if hasattr(var, 'add_offset') and hasattr(var, 'scale_factor'):
-                valid = False
-                valid_missing_variable = False
 
-                if hasattr(var.add_offset, 'dtype') and hasattr(var.scale_factor, 'dtype'):
+            add_offset = getattr(var, 'add_offset', None)
+            scale_factor = getattr(var, 'scale_factor', None)
+            if not (add_offset or scale_factor):
+                continue
 
-                    if var.dtype == var.add_offset.dtype == var.scale_factor.dtype:
-                        valid = True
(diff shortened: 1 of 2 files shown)


compliance_checker/cf/cf.py
```diff
@@ -2360,67 +2360,68 @@ def check_packed_data(self, ds):
         When data to be packed contains missing values the attributes that indicate missing values (_FillValue, valid_min,
         valid_max, valid_range) must be of the same data type as the packed data.
         """
-        
         ret_val = []
         reasoning = []
-        names = list(ds.dataset.variables.iterkeys())
         for name, var in ds.dataset.variables.iteritems():
-            if hasattr(var, 'add_offset') or hasattr(var, 'scale_factor'):
+            if hasattr(var, 'add_offset') and hasattr(var, 'scale_factor'):
```
mhidas: It is valid to have just one of these attributes defined, so we need to handle that case as well. For the purpose of this check, if one of them is not present, we assume it has the same dtype as the other.

mhidas:
@fxmzb123 We also need to handle the case when `add_offset` and `scale_factor` have different types (which should fail the check) and when only one of these is present (which is ok, see my inline comment)

mhidas:
@fxmzb123 Sorry, I just realised your code does handle the case when the two attributes are different. Just took a while to get my head around it.

compliance_checker/cf/cf.py
```diff
@@ -2360,67 +2360,68 @@ def check_packed_data(self, ds):
...
-
-                    data_size = np.s_[ds.dataset.variables[name].size]
-                    data_type_check = np.reshape(ds.dataset.variables[name],data_size)[0]
-                    #, type(ds.dataset.variables[name][0,0,0]) ,type(scale), type(offset)
-                    for x in xrange(data_size):
-                        if type(np.reshape(ds.dataset.variables[name],data_size)[x]) != 'numpy.ma.core.MaskedConstant':
-                            data_type_check = np.reshape(ds.dataset.variables[name],data_size)[x]
-                            break
-                    if type(data_type_check) == type(scale) == type(offset):
+                valid_missing_variable = False
```
daf: I think this needs to be set to `True` here, as I don't see anything in the code below that does so.  Either that, or this variable doesn't need to exist at all.

compliance_checker/cf/cf.py
```diff
@@ -2360,67 +2360,68 @@ def check_packed_data(self, ds):
...
+                            if var.dtype in [np.int8, np.int16, np.int32, np.int64, int()]:
+                                valid = True
+
+                        if valid:
+                            valid_missing_variable = True
+                            # test further with  _FillValue , valid_min , valid_max , valid_range
+                            if hasattr(var, "_FillValue"):
+                                if var._FillValue.dtype !=  var.dtype:
+                                    valid = False
+                                    valid_missing_variable = False
```
daf: For each of these, it would be great to start filling out the `reasoning` list with specific messages instead of one message saying "one of these four is wrong".  Since it is a list, it can handle each one as an individual message: 
```python
reasoning.append("_FillValue is set ($dtype) but does not match var dtype ($dtype)")
reasoning.append("valid_min is set ($dtype) but does not match var dtype ($dtype)")
```

compliance_checker/cf/cf.py
```diff
@@ -2360,67 +2360,68 @@ def check_packed_data(self, ds):
...
+
+                if valid:
+                    valid = True
+                    result = Result(BaseCheck.MEDIUM,\
+                                    valid,\
+                                    ('var', name, 'packed_data'),\
+                                    reasoning)
+                    ret_val.append(result)
+                    reasoning = []
```
daf: No need to set reasoning blank here.

compliance_checker/cf/cf.py
```diff
@@ -2360,67 +2360,68 @@ def check_packed_data(self, ds):
...
-                        ret_val.append(result) 
-                        reasoning = []
-
 
+                    result = Result(BaseCheck.MEDIUM,\
+                                    valid,\
+                                    ('var', name, 'packed_data'),\
+                                    reasoning)
+                    ret_val.append(result)
+                    reasoning = []
```
daf: No need to set reasoning blank here.

mhidas:
@daf I've revised this PR to address your comments and (try to) further simplify the logic. Please let me know what you think & merge if appropriate.

By the way, I've realised that the checks comparing the types of `_FillValue`, `valid_range` etc... attributes to the variable type should really be a separate test, i.e. I don't think it is specific to packed data. (E.g. CF section on [missing data](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#missing-data) says: "The scalar attribute with the name _FillValue and **of the same type as its variable** is recognized by the netCDF library as the value used to pre-fill disk space allocated to the variable".) Perhaps this could be dealt with in a separate PR.

compliance_checker/cf/cf.py
```diff
@@ -2360,66 +2360,78 @@ def check_packed_data(self, ds):
...
+                    reasoning.append("Type of valid_max attribute (%s) does not match variable type (%s)" %\
+                                     (var.valid_max.dtype, var.dtype))
+            if hasattr(var, "valid_range"):
+                if var.valid_range.dtype != var.dtype:
+                    valid = False
+                    reasoning.append("Type of valid_range attribute (%s) does not match variable type (%s)" %\
+                                     (var.valid_range.dtype, var.dtype))
+
+            if not valid:
+                result = Result(BaseCheck.MEDIUM, False, ('var', name, 'packed_data'), reasoning)
```
daf: It would be good for the results from this whole section always be added, instead of only added on failures.  Basically, if you check anything, a `Result` should be produced.  I know we're not entirely consistent with that throughout the codebase but should be.

mhidas: @daf Done.

mhidas:
Based on my experiments this fixes at least a part of #70. Has anyone else been able to test it?

daf:
Thanks @mhidas!

PR #135, opened 2015-06-30
Skip cf_role attr'd vars in check_units, fixes #131

### LINKED ISSUE #131: CF check_units should skip "feature id" variables
A variable with the attribute *cf_role* is used to uniquely identify each feature in a discrete-sampling-geometry (DSG) file ([CF ref](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#coordinates-metadata)). It can have any type, but even if it's numeric, it may not make sense for it to have units (e.g. it could be a serial number for profiles taken on a cruise).

May I propose that any such variable be skipped in check_units (cf.py)?

daf:
@lukecampbell

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -499,6 +499,10 @@ class CFBaseCheck(BaseCheck):
             
             if hasattr(v, 'standard_name') and 'status_flag' in v.standard_name:
                 continue
+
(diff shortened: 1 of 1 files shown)


mhidas:
:+1: Thanks @daf, looks good to me.

PR #140, opened 2015-07-10
Adds support for outputting HTML

lukecampbell:
- Added two new command line arguments:
  - `-o --output` 
  - `-f --format` 

Note: I haven't implemented the REST of the functionality for the output, in other words, you can't use -o anywhere other than for HTML output.

### DIFF
--- a/compliance_checker/runner.py
+++ b/compliance_checker/runner.py
@@ -51,10 +51,10 @@ class ComplianceChecker(object):
         elif criteria == 'lenient':
             limit = 3
 
-        if output_filename == 'stdout' and output_format == 'stdout':
+        if output_filename == '-' and output_format == 'text':
             groups = cls.stdout_output(cs, score_groups, verbose, limit)
 
-        elif output_filename != 'stdout' and output_format == 'html':
+        elif output_filename != '-' and output_format == 'html':
             groups = cls.html_output(cs, score_groups, output_filename, ds_loc)
 
         else:
--- a/compliance_checker/runner.py
+++ b/compliance_checker/runner.py
@@ -1,5 +1,6 @@
 import traceback
 
+from StringIO import StringIO
 from compliance_checker.acdd import ACDDBaseCheck
 from compliance_checker.cf import CFBaseCheck
 from compliance_checker.ioos import IOOSBaseCheck
@@ -54,7 +55,7 @@ class ComplianceChecker(object):
         if output_filename == '-' and output_format == 'text':
             groups = cls.stdout_output(cs, score_groups, verbose, limit)
 
-        elif output_filename != '-' and output_format == 'html':
+        elif output_format == 'html':
             groups = cls.html_output(cs, score_groups, output_filename, ds_loc)
 
         else:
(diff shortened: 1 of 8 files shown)


lukecampbell:
@daf

lukecampbell:
screen shot 2015-07-10 at 1 34 08 pm

requirements.txt
```diff
@@ -5,3 +5,4 @@ lxml>=3.2.1
 cf_units==0.1.0
 requests>=2.2.1
 python-dateutil>=2.2
+Jinja2==2.7.3
```
daf: y u exact `==` a version

lukecampbell: Oh, that was because I ran this
```
pip freeze | grep -i jinja >> requirements.txt
```

I'll pop out the `==`

lukecampbell:
screen shot 2015-07-10 at 2 23 20 pm

daf:
Looks great.  Minor comment about the argument handling, it doesn't quite feel UNIX-y.  `-f` should probably be `text|html` (not `stdout`), and `-o` should accept `-` as UNIX convention to stdout - that way I can pipe the HTML to something if I want.  I also envision more output types (json anyone?).  This would decouple formatting from output instead of those if blocks.

Also, text output could be made a jinja template (right?)

lukecampbell:
Wait one sec, I got a minor refactor that will make it more flexible

lukecampbell:
ok now you can do something like
```
cchecker.py -t acdd -f html dataset.nc
```
And it will spit out a bunch of HTML text

Might be handy for a micro web service ** hint hint **

daf:
- glyphs should toggle collapse when clicked (not just heading)
- priority field is redundant

compliance_checker/runner.py
```diff
@@ -25,7 +26,7 @@ class ComplianceChecker(object):
     the command line or can be used via import.
     """
     @classmethod
-    def run_checker(cls, ds_loc, checker_names, verbose, criteria):
+    def run_checker(cls, ds_loc, checker_names, verbose, criteria, output_filename='stdout', output_format='stdout'):
```
daf: should these defaults be changed to `'-'` and `'text'` respectively?

daf:
Corrective action table only appears to have top level corrective items.  For instance, this text output:

```
Name                             Priority:     Score:Reasoning
conventions                            :3:     0/ 1 : Conventions field is not
                                                      present
std_name                               :3:     0/ 2 :  
    u                                  :3:     0/ 1 :  
        legal                          :3:     0/ 1 : The standard name
                                                      'eastward_surface_velocity
                                                      ' is not in standard name
                                                      table
    v                                  :3:     0/ 1 :  
        legal                          :3:     0/ 1 : The standard name
                                                      'northward_surface_velocit
                                                      y' is not in standard name
                                                      table
time                                   :3:     2/ 4 :  
    time                               :3:     2/ 4 :  
        has_calendar                   :1:     0/ 1 :  
        valid_calendar                 :1:     0/ 1 :  
```

Only shows this in HTML:

lukecampbell:
It should be everything that has a message.

lukecampbell:
What dataset are you using?

lukecampbell:
Oh, I didn't know children results could have even more children.

daf:
Yeah, a complexity that needs to be removed at some point.  Dataset is https://s3.amazonaws.com/paegan/resources/marcooshfradar20120331.nc just in case

lukecampbell:
screen shot 2015-07-10 at 3 08 12 pm

daf:
Looking good!  I have some minor ideas of where to take this but none that require a hold up here.

abirger:
@daf , it looks good already! While perfecting it, just don't forget about the perfect being the enemy of you-know-what...

kwilcox:
this is awesome!

PR #148, opened 2015-07-23
Update CF check fill value to modern style, fixes #147

### LINKED ISSUE #147: Exception during reasoning output
Running the CF checker on [this file](http://thredds.aodn.org.au/thredds/catalog/IMOS/eMII/checker_test/ANFOG/delayed/catalog.html?dataset=IMOS/eMII/checker_test/ANFOG/delayed/IMOS_ANFOG_BCEOPSTUV_20140915T003743Z_SL209_FV01_timeseries_END-20141009T212310Z.nc), I get the following exception during the printing of the reasoning messages:

```
Name                             Priority:     Score:Reasoning
  File "/home/mhidas/bin/cchecker.py", line 36, in <module>
    sys.exit(main())
  File "/home/mhidas/bin/cchecker.py", line 27, in main
    args.format)
  File "/home/mhidas/code/compliance-checker/compliance_checker/runner.py", line 56, in run_checker
    groups = cls.stdout_output(cs, score_groups, verbose, limit)
  File "/home/mhidas/code/compliance-checker/compliance_checker/runner.py", line 91, in stdout_output
    cs.non_verbose_output_generation(score_list, groups, limit, points, out_of)
  File "/home/mhidas/code/compliance-checker/compliance_checker/suite.py", line 280, in non_verbose_output_generation
    self.reasoning_routine(groups, 0)
  File "/home/mhidas/code/compliance-checker/compliance_checker/suite.py", line 369, in reasoning_routine
    print wrapper.fill('%-39s:%1s:%6s/%2s : %s' %(str(indent*'    '+res.name)[0:39], res.weight, str(res.value[0]), str(res.value[1]), str(", ".join(res.msgs))))
TypeError: sequence item 0: expected string, tuple found
```

daf:
Brings CF `check_value_outside_valid_range` in line with other checks.

```
fill_value                             :3:    47/48 :  
    PROFILE                            :3:     0/ 1 :  
        outside_valid_range            :3:     0/ 1 : 99999 must not be between
                                                      0 and 900000
```

instead of a single flat result with `(47/48)` and many messages (that weren't working properly anyway).

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -397,7 +397,7 @@ class CFBaseCheck(BaseCheck):
                 valid     = not (v._FillValue >= rmin and v._FillValue <= rmax)
                 reasoning = []
                 if not valid:
-                    reasoning = ["%s must not be between %s and %s as specified by %s" % (v._FillValue, rmin, rmax, spec_by)]
+                    reasoning = ["%s must not be in valid range (%s to %s) as specified by %s" % (v._FillValue, rmin, rmax, spec_by)]
 
                 ret.append(Result(BaseCheck.HIGH, valid, ('fill_value', k, 'outside_valid_range'), msgs=reasoning))
 
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -399,7 +399,7 @@ class CFBaseCheck(BaseCheck):
                 if not valid:
                     reasoning = ["%s must not be in valid range (%s to %s) as specified by %s" % (v._FillValue, rmin, rmax, spec_by)]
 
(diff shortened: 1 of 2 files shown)


mhidas:
Thanks @daf. That fixes the exception I was getting (#147).

In the result message, it would be even more helpful to say
 `_FillValue (99999) must not be within valid range (0 to 900000)`
or even
`_FillValue (99999) must not be within valid range (0 to 900000) specified by valid_range or valid_min & valid_max`

daf:
Good idea, @mhidas.  I've made it so it will tell which attribute the values actually came from.

mhidas:
Thanks @daf . That's better, though the message still doesn't list the actual name of the attribute being checked (`_FillValue`). The name of the result is `fill_value`, which is close, but why not make it the same? It would just make it easier to find the problem in the CDL using a simple text search.

daf:
Updated, @mhidas.  Will merge now

PR #149, opened 2015-07-28
test on featureType value should be case insensitive

ggalibert:
The list of strings dual_role should be lower case since we're trying to compare it against ```getattr(ds.dataset, 'featureType', '').lower()```.

At the moment this check is giving false result for the following featureType : timeSeriesProfile, trajectoryProfile.


daf:
Good catch, @ggalibert!  Thanks!

PR #161, opened 2015-09-14
Adds support for outputting JSON

lukecampbell:


cchecker.py
```diff
@@ -18,7 +18,8 @@ def main():
 
     return_values = []
     for dataset in args.dataset_location:
-        print "Running Compliance Checker on the dataset from: %s" % dataset
+        if args.format != 'json':
+            print "Running Compliance Checker on the dataset from: %s" % dataset
```
daf: This should probably be printed to stderr, I think that's standard practice.  (then you can lose the if)

PR #164, opened 2015-09-16
Fixes #163: Exception handling

### LINKED ISSUE #163: Checker returns overall "pass" result and no stderr output when exceptions encountered
[This file](http://thredds.aodn.org.au/thredds/catalog/IMOS/eMII/checker_test/misc/catalog.html?dataset=eMII/checker_test/misc/cf_checker_exception.nc) appears to pass all CF checks, but causes an exception in the `check_flags` method (separate issue). The checker output looks like this:
```
Running Compliance Checker on the dataset from: cf_checker_exception.nc
The following exceptions occured during the cf checker (possibly indicate compliance checker issues):
cf.check_flags: 'unicode' object has no attribute 'dtype'


                     The dataset scored 65 out of 65 points                     
                              during the cf check                               
All tests passed!
```
Of this, only the first line is sent to stderr. 

Obviously we can fix the code so this particular exception doesn't occur, but others may still crop up. In general, if any exceptions occur during checks, I think it would make sense to 
 * report these to stderr; and
 * return an overall "fail" result (i.e. non-zero exit value).

Probably best to use a specific exit value other than 1, which is used when the file has failed a correctly running check.

mhidas:
* Exceptions reported to stderr;
* Exceptions also reported for html and json output options, not just text;
* If any exceptions, cchecker.py exit value is 2, even if all successfully run checks passed.

### DIFF
--- a/cchecker.py
+++ b/cchecker.py
@@ -18,18 +18,21 @@ def main():
     args.test = args.test or ['acdd']
 
     return_values = []
+    had_errors = []
     for dataset in args.dataset_location:
         if args.format != 'json':
             print("Running Compliance Checker on the dataset from: %s" % dataset, file=sys.stderr)
-        return_value = ComplianceChecker.run_checker(args.dataset_location[0],
+        return_value, errors = ComplianceChecker.run_checker(args.dataset_location[0],
                                       args.test,
                                       args.verbose,
                                       args.criteria,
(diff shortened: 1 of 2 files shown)


daf:
:+1: thanks!

lukecampbell:
:+1:

Issue #154, opened 2015-08-19
Add references to checker output?


mhidas:
For a data provider trying to correct issues picked up by the checker, it would be helpful if the brief reasoning message for each failed test could be followed by a reference (URL or text) to the relevant section of the convention document. Could we add a `reference` attribute of the `Result` object, and list these separately in the checker output (perhaps at the end)?

daf:
@mhidas , @lukecampbell addressed this in #167 and made the ref part of the check name itself.  If you get a minute can you please evaluate it and see if it addresses this issue well?

mhidas:
@daf @lukecampbell Thanks. I've finally had a look at this. It's not quite what I had in mind, but it might just do the trick. However, breaking the grouping (as you pointed out https://github.com/ioos/compliance-checker/pull/167#issuecomment-144741447) is a possible drawback. I don't care about combining scores according to the grouping, but it does order the results in a meaningful way.

Having said that, I haven't actually managed to see what the new output looks like, because even though the unittests pass, whenever I try to run the cf checker on a file that doesn't pass (running code at db62cea9e3ba90843ea0961b734077628f444c70), I get
```
$ cchecker.py -t=cf compliance_checker/tests/data/bad_data_type.nc 
Running Compliance Checker on the dataset from: compliance_checker/tests/data/bad_data_type.nc
Traceback (most recent call last):
  File "/home/mhidas/bin/cchecker.py", line 47, in <module>
    sys.exit(main())
  File "/home/mhidas/bin/cchecker.py", line 36, in main
    args.format)
  File "/home/mhidas/work/code/compliance-checker/compliance_checker/runner.py", line 48, in run_checker
    score_groups = cs.run(ds, *checker_names)
  File "/home/mhidas/work/code/compliance-checker/compliance_checker/suite.py", line 100, in run
    groups = self.scores(vals)
  File "/home/mhidas/work/code/compliance-checker/compliance_checker/suite.py", line 490, in scores
    grouped = self._group_raw(raw_scores)
  File "/home/mhidas/work/code/compliance-checker/compliance_checker/suite.py", line 538, in _group_raw
    grouped = itertools.groupby(sorted(raw_scores, key=group_func), key=group_func)
UnicodeDecodeError: 'ascii' codec can't decode byte 0xc2 in position 0: ordinal not in range(128)
```

or
```
[...]

Issue #109, opened 2015-05-05
File Independent Scoring Concept


mbiddle-bcodmo:
Hi all,

This might not be worth talking about, but I wanted to at least get it out in the open if it's something other people could find useful.

One of my thoughts is to use this compliance checker for different groups of data providers to test if their netCDF files are correctly formatted for transmission to the archive. Reviewing my understanding of how the scoring system works, the current incarnation has a varying amount of total available points depending on the files and tests. I understand that comparing between tests should yield differing scores, however, providing some consistency between varying files (using the same test) would be useful. 

It would be beneficial, from the NCEI perspective, if there was some consistent number to represent the score the file receives for each specific test. This would facilitate the use of the compliance checker across a breadth of files and provide a consistent evaluation for all files. 

My thought is that NCEI could say something along the lines of 'Your files should have a compliance checker score larger than X.' Although, now that I think about it, maybe evaluating the score as a percentage would suffice for these needs.

lukecampbell:
Hey Matt, we're in the process of reevaluating the entire scoring process for compliance checker as it is. 
We're struggling with conflicting desires:

1. Users want a simple "pass/fail" answer to the question "is this dataset good enough".
   - The problem is, that answer is not an easy one to come up with.
2. Users want guidance for how to fix their datasets if it's not good enough.
3. There are varying degrees of compliance and for documents like CF, compliance is sometimes up to interpretation of the document, and therefore debatable.

mbiddle-nodc:
I understand the difficulty with trying to implement a scoring concept that covers both the simple "pass/fail" concept but also provides enough details to fix the datasets. As far as the interpretation of the guidance documents, as long as you describe how the documentation was interpreted, that's the best that can be done. 

Another resource that I was just made aware of, apparently JPL has developed an online compliance checker for CF, ACDD, and GDS2 (very specific guidance for GHRSST data sets). http://podaac-uat.jpl.nasa.gov/mcc/.

daf:
Nice, they made it available!  They had an intern develop a GUI on top of this same compliance-checker  in Summer 2014.  At the time, they hadn't planned on releasing anything though.

kknee:
+1 on being happy to see the JPL version released!

abirger:
Wow, that GUI looks nice! It is what I had in mind in the Trade Study recommendations. 
[...]

PR #167, opened 2015-09-22
Adds more detailed result names to CF

lukecampbell:
- Adds UTF-8 support for results
- Adds more detailed result names to CF checker
- Adds limit support for JSON and HTML output

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
 import re
 from functools import wraps
 from collections import defaultdict, OrderedDict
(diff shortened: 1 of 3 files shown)


compliance_checker/cf/cf.py
```diff
@@ -1398,12 +1400,12 @@ def check_independent_axis_dimensions(self, ds):
             if valid == False :
                 ret_val.append(Result(BaseCheck.MEDIUM, \
                                valid, \
-                               ('var', name, 'check_independent_axis_dimensions'),['The %s dimension for the variable %s does not have an associated coordinate variable, but is a Lat/Lon/Time/Height dimension.'%(dim_name,name)]))
+                               (u'§5.1 Geophysical variables contain valid dimensions', name, 'check_independent_axis_dimensions'),['The %s dimension for the variable %s does not have an associated coordinate variable, but is a Lat/Lon/Time/Height dimension.'%(dim_name,name)]))
```
daf: the var!

daf:
I love this, and does well to address #154.

One minor thing I've noticed is that it's breaking apart the grouping mechanism.  We've talked about it before (can't find a ref, of course, but related to #109) in that the groups is basically an incorrect mechanism for scoring and should eventually be abolished - maybe this is the time to do so, or at least start it down that path?  It breaks the overall score, but in a way, that's already broken anyway.

So yeah, let's do this.

PR #168, opened 2015-09-22
Updates standard names table to version 29

lukecampbell:


compliance_checker/runner.py
```diff
@@ -90,35 +91,37 @@ def stdout_output(cls, cs, score_groups, verbose, limit):
...
     @classmethod
-    def html_output(cls, cs, score_groups, output_filename, ds_loc):
+    def html_output(cls, cs, score_groups, output_filename, ds_loc, limit):
         '''
         Generates rendered HTML output for the compliance score(s)
         @param cs              Compliance Checker Suite
         @param score_groups    List of results
         @param output_filename The file path to output to
         @param ds_loc          Location of the source dataset
+        @param limit           Integer value for limiting output
```
daf: this doesn't really explain what a limit is, what is being limited?

compliance_checker/__init__.py
```diff
@@ -1 +1,2 @@
 __version__="1.1.1"
+__stdname_table__="v29"
```
daf: This should either have the string `cf` involved in the name, or moved to a cf specific file instead of the package init.

compliance_checker/runner.py
lukecampbell: I just copied what was from the other methods.

compliance_checker/__init__.py
lukecampbell: Ah, good call.

compliance_checker/runner.py
daf: lal

lukecampbell: Called out. I'll change it.

lukecampbell:
@daf Ok, moved, and fixed the docstrings

PR #174, opened 2015-10-27
ACDD - Removes unnecessary reasoning and adds exceptions to attribute checks

lukecampbell:


daf:
Ran locally, this actually fails a test:

```
======================================================= FAILURES ========================================================
______________________________________________ TestCF.test_check_calendar _______________________________________________

self = test_check_calendar (compliance_checker.tests.test_cf.TestCF)

    def test_check_calendar(self):
        dataset = self.get_pair(static_files['example-grid'])
        results = self.cf.check_calendar(dataset)
        for r in results:
            self.assertTrue(r.value)
    
        dataset = self.get_pair(static_files['bad'])
        results = self.cf.check_calendar(dataset)
        rd = {r.name[1:] : r.value for r in results }
>       self.assertFalse(rd[('bad_time_1', 'has_calendar')])
E       KeyError: ('bad_time_1', 'has_calendar')

compliance_checker/tests/test_cf.py:555: KeyError
```

lukecampbell:
Weird that a CF test would fail for a PR that only touches ACDD.

lukecampbell:
Oh wait, the lingering commit touched CF.

PR #176, opened 2015-10-27
Fix travis

ocefpaf:
- `sudo: false` moves from the old travis to the new and faster travis
- `conda update --yes --all` updates everything rather than just conda
- `cf_units==0.1.1` This is the real fix.  The previous version had a baked conda numpy string for NP19 and the rest of the packages were compiled against numpy > 1.10.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: python
 
+sudo: false
+
 env:
     - CONDA="python=2.7"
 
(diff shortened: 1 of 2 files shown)


daf:
THANK YOU! :fireworks:

ocefpaf:
No problem.  This is one of those idiosyncrasies of conda that results in a lot of frustration!

Issue #130, opened 2015-06-24
CF check_units should not skip scalar variables


mhidas:
In cf.py `check_units` requires variables to have a *units* attribute, with some exceptions. Currently one of the exceptions is any variable with `v.shape == ()`, i.e. scalar variables. This seems wrong to me. [CF](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#units) says "Units are not required for dimensionless *quantities*". This refers to the physical quantity being stored, not the number of values or their (lack of) association with any dimension in the file.

For example, in a file containing a single timeseries, we may store the location in dimensionless variables, and these should have units.

daf:
Fixed in #177

PR #177, opened 2015-10-27
Removes dimensionless from units exception list

lukecampbell:
- Adds some well known container variables, I expect this list to grow
   over time.
 - Adds new netCDF file (and CDL) to demonstrate
 - Adds test to verify


lukecampbell:
This fixes #130

PR #181, opened 2015-11-02
Fixes recursion bug in circular dependencies in attrs

### LINKED ISSUE #170: CF check_two_dimensional: maximum recursion depth exception
Running the CF checker on [this file](http://thredds.aodn.org.au/thredds/catalog/IMOS/eMII/checker_test/SOOP/ASF/catalog.html?dataset=eMII/checker_test/SOOP/ASF/IMOS_SOOP-ASF_FMT_20150205T000200Z_ZMFR_FV02.nc), it passes all checks, but there's this exception:
```
WARNING: The following exceptions occured during the cf checker (possibly indicate compliance checker issues):
cf.check_two_dimensional: maximum recursion depth exceeded while calling a Python object
```

I think this is due to self-referencing coordinates attributes, e.g. 
```
   LATITUDE:coordinates = "TIME LATITUDE LONGITUDE" ;
```
... which is not correct, but still the checker should handle it nicely. This was already fixed once (#107), but it seems not completely.

lukecampbell:
Fixes #170


compliance_checker/cf/util.py
```diff
@@ -155,38 +155,64 @@ def fromkeys(cls, seq, value=None):
...
+            return self.reference_map[coord]
+        return NCGraph(self.ds, coord, self.ds.variables[coord], self.reference_variables, self.reference_map)
+
+    def get_grid_mapping(self, gm):
+        if gm not in ds.variables:
+            return
+        if gm in self.reference_map:
+            return self.reference_map[gm]
+        return NCGraph(ds, gm, ds.variables[gm], self.reference_map)
+
```
lukecampbell: Essentially this utility is a depth first search, well DFS is vulnerable to recursive errors which is the source of the original bug. In keeping with the typical algorithm, I added a list that maintains a list of references, so as not to reference two leafs recursively.

Issue #171, opened 2015-10-27
CF check_flags: 'numpy.int8' object is not iterable


mhidas:
Running the CF checker on [this file](http://thredds.aodn.org.au/thredds/catalog/IMOS/eMII/checker_test/SOOP/ASF/catalog.html?dataset=eMII/checker_test/SOOP/ASF/IMOS_SOOP-ASF_FMT_20150205T000200Z_ZMFR_FV02.nc), we get this exception:

```
cf.check_flags: 'numpy.int8' object is not iterable
```

This is another special case that can probably be considered incorrect according to CF, but in that case it should give a fail result, rather than an exception. The file has flag variables with only a single value in the `flag_values` attribute, rather than a list. [CF](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#flags) says it should be a list.

daf:
Fixed in 90375ba

daf:
@mhidas after applying your fix and testing on this file, I get a `UnicodeEncodeError: 'ascii' codec can't encode character u'\xa7' in position 0: ordinal not in range(128)` when trying to format the results, but that's certainly a separate issue.

mhidas:
@daf I get the same exception from versions before my fix was merged, e.g. try running https://github.com/ioos/compliance-checker/commit/db62cea9e3ba90843ea0961b734077628f444c70 
on the test file compliance_checker/tests/data/bad_region.nc 

There's also several slightly different but similar issues I've been seeing, since #167 was merged (see my comments https://github.com/ioos/compliance-checker/issues/154#issuecomment-147624951)

PR #187, opened 2015-11-04
CF check flags exception fix + add unittest

mhidas:
Running the CF checker on e.g. [this file](http://thredds.aodn.org.au/thredds/catalog/IMOS/eMII/checker_test/SRS/ALT/catalog.html?dataset=eMII/checker_test/SRS/ALT/IMOS_SRS_TCPS_20130912T034200Z_SRSSTO_FV01_SBE37_15m_END-20141010T002300Z_C-20141105T064337Z.nc) results in the exception
```
cf.check_flags: 'unicode' object has no attribute 'dtype'
```
This is because some of the variables (incorrectly) have string-valued `flag_values` attributes and the code tries to access their `dtype`. 

I've recently added a check making sure the `flag_values` attribute is a list ( #171). Now I've moved this check to the top so it's the first thing checked (and also converts any other type to a numpy array so the other checks don't fail).

Also, there was no unittest for `check_flags`, so I've added one.


lukecampbell:
:+1:

PR #191, opened 2015-11-08
Free cf_units version

ocefpaf:
cf_units 1.0.0 is out.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -2,7 +2,7 @@ OWSLib>=0.8.3
 petulant-bear>=0.1.3
 Wicken>=0.1.2
 lxml>=3.2.1
-cf_units==0.1.1
+cf_units>=0.1.1
 requests>=2.2.1
 python-dateutil>=2.2
 Jinja2>=2.7.3


ocefpaf:
@lukecampbell I guess we need to add a better test here.  Travis passed with `cf_units 1.0.0`, but it is clearly not OK.

Do not merge this yet.  I will put some fixes and a test here soon...

ocefpaf:
Jen reported the following error:

```python
[Anaconda2] C:\Users\jennifer.bosch>compliance-checker -h
Could not load acdd = compliance_checker.acdd:ACDDBaseCheck : (cf-units 1.0.0 (c
:\users\jennifer.bosch\appdata\local\continuum\anaconda2\lib\site-packages), Req
uirement.parse('cf-units==0.1.1'))
Could not load cf = compliance_checker.cf.cf:CFBaseCheck : (cf-units 1.0.0 (c:\u
sers\jennifer.bosch\appdata\local\continuum\anaconda2\lib\site-packages), Requir
ement.parse('cf-units==0.1.1'))
Could not load ioos = compliance_checker.ioos:IOOSBaseCheck : (cf-units 1.0.0 (c
:\users\jennifer.bosch\appdata\local\continuum\anaconda2\lib\site-packages), Req
uirement.parse('cf-units==0.1.1'))
usage: compliance-checker-script.py [-h] [--test {}]
                                    [--criteria [{lenient,normal,strict}]]
                                    [--verbose] [-f {text,html,json}]
                                    [-o OUTPUT] [-V]

positional arguments:
  dataset_location      Defines the location of the dataset to be checked.

optional arguments:
  -h, --help            show this help message and exit
  --test {}, -t {}, --test= {}, -t= {}
                        Select the Checks you want to perform.
  --criteria [{lenient,normal,strict}], -c [{lenient,normal,strict}]
                        Define the criteria for the checks. Either Strict,
                        Normal, or Lenient. Defaults to Normal.
  --verbose, -v         Increase output. May be specified up to three times.
  -f {text,html,json}, --format {text,html,json}
                        Output format
  -o OUTPUT, --output OUTPUT
                        Output filename
  -V, --version         Display the IOOS Compliance Checker version
                        information.
```


@lukecampbell I thought about adding a call to `compliance-checker -h` in Travis-CI, but that won't raise a non-zero `$?`.  This message is not seen as an error :unamused: inside Travis-CI.

Anyway, the version information is from the `requirements.txt `.  That means this PR fixes the issue.

lukecampbell:
If cf_units fails to appear correctly and causes the various plugins not to be loaded, the tests SHOULD fail on Travis. If this is not the case, we should investigate as to how they are able to be run and pass.

ocefpaf:
In the case above they passed because the cf_units version was passed correctly.

What I meant is that I wanted to add a test to avoid seeing this resurfacing,  but the CLI failure ends returns a `$? = 0` meaning that Travis won't fail by just running `compliance-checker --help`.

So I guess that the CLI script could be updated to do something like `sys.exit(1)` in that case; or we could be smarter and test that logic in travis.  I am :+1: for the former.

rsignell-usgs:
Is there a "dash" vs. "underscore" problem here?
The package is `cf_units`, not `cf-units`

ocefpaf:
Not sure.  It is underscore in the PR,

https://github.com/ioos/compliance-checker/pull/191/files#diff-b4ef698db8ca845e5845c4618278f29aR5

rsignell-usgs:
@lukecampbell ??

Issue #192, opened 2015-11-11
3.5 Flags and flag attributes - flag_values attr does not have same type as var


daf:
Ran the new compliance-checker-web on the file referenced in #171 and got a number of "Corrective Actions" with the following text:

```
3.5 Flags and flag attributes	3	flag_values attr does not have same type as var (fv: int8, v: int16)
```

This is partially useful as it tells me what types were wrong (cryptically, though), but it does not tell me what variable.  Especially as there's a ton of these messages, all the same.

daf:
Actually, this may be an issue with compliance-checker-web's output formatting, @lukecampbell ?

mhidas:
@daf I've come across this as well and thought it should be fixed. I think it's a result of the new naming system for CF checks. For a lot of them the variable name (and sometimes useful - IMHO - third-level check names) were removed, leading to many check results being combined into one. Some other checks have the same issue:
 * 2.4 Dimension order
 * 3.1 Variables contain valid CF Units
 * 3.3 Standard Names

Sure, sometimes the variable name is included in the reasoning message, but that's not very easy to read when there's a whole bunch of them with the same error. e.g.
```
3.1 Variables contain valid CF Units   :3:     0/ 6 : unknown units type
                                                      (Degrees clockwise from
                                                      true North) for CDIR,
                                                      unknown units type
                                                      (Decibels) for ABSI1,
                                                      unknown units type
                                                      (Decibels) for ABSI2,
                                                      unknown units type
                                                      (Decibels) for ABSI3,
                                                      ... etc ...
```

DanielJMaher:
"Someone needs to revisit the 3.5 test and make sure the variable is being passed " - Luke

benjwadams:
I can generate this myself, but it'd be preferable to attach an .nc file or minimal example to reproduce.

benjwadams:
@daf file referenced in link is currently 404.  Can make a minimal example to reproduce but would prefer using the original file to reproduce.

benjwadams:
Never mind, there's an example contained within the test data that accomplishes something similar.

PR #194, opened 2015-11-16
Suggestion: include variable names to CF result names 

mhidas:
This is not yet complete! I'm just putting it up as a suggestion. I think it would solve #192, and make the results easier to read (in plain text output -- I haven't looked at the HTML yet, sorry). Please let me know if you think it's a good idea, and I'll make these changes the other check methods (some of them will also requite changes to the unit tests).

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -530,14 +530,14 @@ class CFBaseCheck(BaseCheck):
             units = getattr(v, u'units', None)
 
             # 1) "units" attribute must be present
-            presence = Result(BaseCheck.HIGH, units is not None, u'§3.1 Variables contain units')
+            presence = Result(BaseCheck.HIGH, units is not None, (u'§3.1 Variables contain units', k))
             if not presence.value:
                 presence.msgs = [u'units attribute is required for %s' % k]
                 ret_val.append(presence)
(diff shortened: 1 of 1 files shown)


lukecampbell:
I wish I wouldn't have let this grow stale :(

I have a plan for 2.3.0 which will create a new attribute in the results so software clients can determine a few context things:

- variable or global
- attribute
- check_kind: a flag that indicates the kind of check. 

So each result will have something like

variable=global
attribute=platform
check_kind=exists

variable=temperature
attribute=standard_name
check_kind=standard_name

variable=temperature
attribute=units
check_kind=valid_units

something like that. When 2.3 gets rolling I'll detail it more.

mhidas:
@lukecampbell Thanks, that seems sensible, but it will only resolve my issue if these properties are then used to consistently group and order the final text output. Is that part of the plan?

lukecampbell:
Unfortunately the grouping code breaks the JSON and HTML outputs. I'm slowly moving away from the grouping logic with the tradeoff that we'll have the ability to identify variables and metadata about the checks so that applications like user interfaces or JSON parsers could group correctly. Right now groups just create sub-documents which mask checks for everything but the plain-text output which isn't very machine parseable.

Issue #206, opened 2016-02-04
Conventions field is "CF-1.x (x in 0-6)"


rsignell-usgs:
I used the web service to run a CF-1.6 check on this OpenDAP Data URL
http://geoport-dev.whoi.edu/thredds/dodsC/usgs/data2/emontgomery/stellwagen/CF-1.6/CAPE_COD_BAY/3051B-A.cdf
and got back a perfect 12/12, but also this corrective action:
![2016-02-04_7-57-14](https://cloud.githubusercontent.com/assets/1872600/12815708/032c021a-cb16-11e5-917d-431300207f16.png)

I'm not sure what this is supposed to mean.   The dataset *does* have the "Conventions = CF-1.6" attribute assigned.

lukecampbell:
Looks like a bug. The corrective action would suggest that it's missing, but I just checked and it's not.

rsignell-usgs:
Since you already have an API, it would be easy to make a RESTful service also, no?  
I still think it would be a nice addition.    Some people don't like to code.   :grinning:

daf:
Early in development we used the messages field for both "correct" and "incorrect" things, before we settled on the notion that messages are only for issues.  Basically, just remove the string [here](https://github.com/ioos/compliance-checker/blob/master/compliance_checker/cf/cf.py#L475).

PR #208, opened 2016-02-04
Fixes misleading messages for valid checks

lukecampbell:
There were a few lingering messages that get kept even though the checks
passed.

Fixes https://github.com/ioos/compliance-checker/issues/206


lukecampbell:
@daf would you care to merge?

PR #209, opened 2016-02-04
Doesn't require units attr for metadata vars like platform

lukecampbell:
This plus #208 will ensure that 

http://geoport-dev.whoi.edu/thredds/dodsC/usgs/data2/emontgomery/stellwagen/CF-1.6/CAPE_COD_BAY/3051B-A.cdf

gets a perfect score.

screen shot 2016-02-04 at 10 36 16 am

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -543,12 +543,14 @@ class CFBaseCheck(BaseCheck):
         clim_vars = self._find_clim_vars(ds)
         boundary_vars = self._find_boundary_vars(ds).itervalues()
         container_vars = self._find_container_variables(ds)
+        metadata_vars = self._find_metadata_vars(ds)
 
         for k, v in ds.dataset.variables.iteritems():
 
             # skip climatological vars, boundary vars
             if v in clim_vars or \
                v in boundary_vars or \
+               v in metadata_vars or \
                k in container_vars:
                continue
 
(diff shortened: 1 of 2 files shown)


lukecampbell:
@rsignell-usgs @daf

lukecampbell:
CF 1.6 doesn't explicitly say not for metadata variables/auxiliary like platform_id but it says only variables with dimensional quantities.

> The units attribute is required for all variables that represent dimensional
> quantities (except for boundary variables defined in Section 7.1, "Cell
> Boundaries" and climatology variables defined in Section 7.4, "Climatological
> Statistics" ).

lukecampbell:
@daf care to review?

compliance_checker/cf/cf.py
```diff
@@ -203,7 +203,7 @@ def _find_metadata_vars(self, ds, refresh=False):
             elif getattr(var, 'cf_role', None) == 'timeseries_id':
                 self._metadata_vars[ds].append(var)
 
-            elif len(var.dimensions) == 0:
+            elif getattr(var, 'standard_name', None) is None and len(var.dimensions) == 0:
```
lukecampbell: This has implications. It's hard to know whether a variable is a metadata variable or part of a coordinates system without inspecting feature types.

I'm eliminating some false negatives here by saying if it's dimensionless and it doesn't have a standard_name attribute, it's likely to be an auxiliary, ancillary or metadata variable.

compliance_checker/tests/test_cf.py
```diff
@@ -290,7 +290,7 @@ def test_check_flags(self):
         self.assertEqual(messages.count(u'flag_values attr does not have same type as var (fv: <U1, v: int16)'), 1)
 
 
-    def test_check_units(self):
+    def test_check_bad_units(self):
```
lukecampbell: We had two tests with the same name...

daf: haha good

PR #218, opened 2016-02-08
Closes #86 . Reverts back to only accept CF 1.6

### LINKED ISSUE #86: CF Conventions check allows pre 1.6 strings
Added in #68 is a list of new convention strings allowed from CF-1.0 to CF-1.6.  As noted by @DanielJMaher in the comments of that PR, this checker is only designed for CF-1.6.  Allowing other checks gives a false impression that other CF standards will be rigorously checked (they won't be).

I'm fine with the other changes of that PR but this list needs to be reverted back to a single entity, CF-1.6.

DanielJMaher:
Remove the list of conventions allowed.  Keeps the logic that supports multiple conventions as it does not harm and we might want to leverage it later.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -471,8 +471,8 @@ class CFBaseCheck(BaseCheck):
         """
         2.6.1 the NUG defined global attribute Conventions to the string value "CF-1.6"
         """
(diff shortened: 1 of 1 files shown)


lukecampbell:
@daf just a heads up this was requested a while back since we don't have valid checkers for the other conventions.

daf:
yeah this is good :+1:

Issue #211, opened 2016-02-04
Self Referencing Coordinates Bug 


lukecampbell:
http://geoport-dev.whoi.edu/thredds/dodsC/usgs/data2/emontgomery/stellwagen/CF-1.6/CAPE_COD_BAY/3051B-A.cdf

Running CF on this dataset results in a several `coordinates_reference_itself` for the §5.2 Check(s). Looking at the actual dataset, I cannot see any coordinates that reference themselves in a coordinates attribute. After looking and debugging minimally, I think the issue is in how we are building a graph structure for checking coordinates and references for §5.2

lukecampbell:
Discovered from https://github.com/ioos/compliance-checker/issues/200

kwilcox:
@lukecampbell I'm seeing either this bug, or another in the `TestCF.test_check_two_dimensional` test. It inconsistently passes and fails for me on the same dataset. @ggalibert

kwilcox:
I fixed the bug in `check_two_dimensional` here: https://github.com/ioos/compliance-checker/commit/1eac74f591a5ae50ce600203bd565cd4b7026e6b

lukecampbell:
![screen shot 2016-02-08 at 3 53 41 pm](https://cloud.githubusercontent.com/assets/1155163/12898728/288bd63e-ce7c-11e5-804b-d7cb909ee2aa.png)

PR #219, opened 2016-02-08
Adds better checking for self-referencing coords

lukecampbell:
Fixes and now tests #211


compliance_checker/cf/util.py
```diff
@@ -218,7 +218,8 @@ def get_coordinate(self, coord):
         if coord not in self.ds.variables:
             return
         if coord in self.reference_map:
-            self.reference_variables.add(self.obj.name)
+            if self.obj.name == coord:
+                self.reference_variables.add(self.obj.name)
```
lukecampbell: This fixes the graph recursive/self-referencing issue but it changes the nature of the test and how it reports. We no longer report variables that list coordinates that list themselves: we only list coordinates that are self referencing. Which in my opinion is MUCH better. Too many times people ask "why is temperature saying it is self referencing coordinates"... Well, it's not but latitude is...

PR #231, opened 2016-03-01
Fixes #151

### LINKED ISSUE #151: ACDD - Dimensionless variables are not required to have units attribute

DanielJMaher:
Adds in logic to check for dimensions and units.  Error messages designed to catch if units exist, but no dimensions and if dimensions exist, but no units.  

I could not run the test because i do not have the file.

### DIFF
--- a/compliance_checker/acdd.py
+++ b/compliance_checker/acdd.py
@@ -146,7 +146,7 @@ class ACDDBaseCheck(BaseCheck):
             results.append(Result(BaseCheck.HIGH, check, (variable, "var_std_name"), msgs))
 
         return results
-
+    
     @score_group('varattr')
     def check_var_units(self, ds):
         results = []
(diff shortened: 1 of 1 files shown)


compliance_checker/acdd.py
```diff
@@ -161,9 +161,17 @@ def check_var_units(self, ds):
...
+            unit_check = hasattr(ds.variables[variable], 'units')
+            dim_check = (len(getattr(ds.variables[variable], 'dimensions')) > 0)
+            # Check if we have dimensions and no units
+            if not unit_check and dim_check:
+                msgs.append("Var %s missing attr units, but has dimensions" % variable)
+            # Check if we have units and no dimensions
+            if not dim_check and unit_check:
+                msgs.append("Var %s has attr units, but no dimensions" % variable)
+            # Final check for pass/fail.  If we have units and dimensions or no units and dimensions we pass.
+            check = (unit_check == dim_check)
```
lukecampbell: Actually, I think we only need to see if it has units if and only if it is not dimensionless...

so something like

```python
if ds.dimensions == tuple():
    continue
...
```

DanielJMaher: Is there already a check for the case where they have an attributes unit and not dimensions?

DanielJMaher:
@lukecampbell I took your advice, but i left the units_check and the no_dim_check variables because i thought making one line for the "skipping if no dimensions" check was too hard to read.  Same thing though, just tidier IMO.

PR #234, opened 2016-03-11
References CDL now. Added resource.py. Fixed Test. Left off modified …

### LINKED ISSUE #230: ACDD Test NetCDF file is not included in this repo
The tests aren't very robust (read: only one test), but never the less it shouldn't be hardcoded like this.  That file should be included in the compliance_checker/tests/data folder and referenced from there.

DanielJMaher:
…CDLs (might be needed later) Fixes #230 for CDL.  Still need file for ACDD


DanielJMaher:
@lukecampbell I passed!

lukecampbell:
NICE!

@daf what do you think?

daf:
Looks like a fine idea.  I'm hesistant to want to add a new dep, especially one that doesn't look like it adds very much over [`subprocess.call()`](https://docs.python.org/2/library/subprocess.html#subprocess.call) - willing to make that swap?

DanielJMaher:
@daf @lukecampbell Removed dependency per Dave's guidance.  Can someone please pull down and run the CF test with a clean director (i.e. No NetCDF files for testing).  It worked for me, but i want to be safe.

compliance_checker/tests/data/rhgrid.cdl
```diff
@@ -0,0 +1,1048 @@
...
+		lon:units = "degrees_east" ;
+	float lat(rgrid) ;
+		lat:long_name = "latitude" ;
+		lat:units = "degrees_north" ;
+	float rgrid(rgrid) ;
+		rgrid:compress = "latdim londim" ;
+data:
+
+ PS = _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, 
+    _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _,
```
lukecampbell: Please don't do full ncdump for this, there's no data in this file, just use ncdump -h.

compliance_checker/tests/data/rhgrid.cdl
```diff
@@ -0,0 +1,1048 @@
...
+    _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, 
+    _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, 
+    _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, 
+    _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, 
+    _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, 
+    _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, 
+    _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, 
+    _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, 
+    _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, _, 
+
```
lukecampbell: because two hours later of scrolling, I can finally see the next file.

daf:
I pulled a clean copy, did a fresh conda env, and things worked well.

DanielJMaher:
I'll fix the files with no data in them.

lukecampbell:
Also, trim mapping.nc, it's 20MB. Change the dimensions or subset it or something so that's it's WAY smaller. This commit shouldn't have more than a few hundreds of LOC changes, not 240k+

DanielJMaher:
@lukecampbell You are just upset because i would become the top contributor.... But seriously, i'll cut that down.

DanielJMaher:
@lukecampbell Better?  I left compliance_checker/tests/data/index_ragged.cdl and compliance_checker/tests/data/ru07-20130824T170228_rt0.cdl as their full files because they're not too big,  I didn't want to get too nitpicky and spend more time than is needed.

lukecampbell:
That and 20MB is a rather large file to manage in a revision control system.

mhidas:
@DanielJMaher @lukecampbell  Thanks for this. I had also thought this would be the best way to include test files. However, I have two comments:
* Shouldn't we now mention `ncgen` as a dependency (for running tests at least)
* Shouldn't we now delete all those test netCDF files from the repo?

lukecampbell:
Yes and yes.

lukecampbell:
https://github.com/ioos/compliance-checker/issues/275

PR #235, opened 2016-03-14
Ancillary

DanielJMaher:
Must go in after #234 because of https://github.com/DanielJMaher/compliance-checker/blob/cdl/compliance_checker/cf/util.py#L222

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -725,9 +725,9 @@ class CFBaseCheck(BaseCheck):
             anc_result = Result(BaseCheck.HIGH, name=('§3.4 Ancillary Variables', k))
             msgs = []
 
-            if not isinstance(anc, str) and not isinstance(anc, unicode):
+            if not isinstance(anc, basestring):
                 anc_result.value = False
-                anc_result.msgs = ["ancillary_variable(s) %s is not a string or unicode" % anc]
+                anc_result.msgs = ["ancillary_variable(s) {} is not a string or unicode".format(anc)]
                 ret_val.append(anc_result)
                 continue
 
@@ -738,7 +738,7 @@ class CFBaseCheck(BaseCheck):
                 if a in ds.variables:
                     existing += 1
                 else:
-                    msgs.append("ancillary var %s does not exist" % a)
+                    msgs.append("ancillary var {} does not exist".format(a))
 
             anc_result.value = (existing, len(ancs))
             anc_result.msgs = msgs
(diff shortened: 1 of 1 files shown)


DanielJMaher:
@lukecampbell

compliance_checker/cf/cf.py
```diff
@@ -723,9 +725,9 @@ def check_ancillary_data(self, ds):
             anc_result = Result(BaseCheck.HIGH, name=('§3.4 Ancillary Variables', k))
             msgs = []
 
-            if not isinstance(anc, str):
+            if not isinstance(anc, str) and not isinstance(anc, unicode):
```
lukecampbell: @daf @kwilcox isn't there some special python3 magic that has to be done here?

kwilcox: This is fairly common (without the use of `six`) to support both py2 and py3:

```
try:
    basestring
except NameError:
    basestring = str
```

Then just always do `isinstance(foo, basestring)`

lukecampbell: Thanks, I thought so. I don't do a lot of py2 and py3 support together.
@DanielJMaher would you please?

DanielJMaher: @lukecampbell Im on it

kwilcox: you can add that test in a utils function and always import it from there... i know its used other places in CC already.

compliance_checker/cf/cf.py
```diff
@@ -723,9 +725,9 @@ def check_ancillary_data(self, ds):
...
                 anc_result.value = False
-                anc_result.msgs = ["ancillary_variables is not a string"]
+                anc_result.msgs = ["ancillary_variable(s) %s is not a string or unicode" % anc]
```
kwilcox: You should use new style string formatting... this style is deprecated.

lukecampbell: agreed.

compliance_checker/cf/cf.py
DanielJMaher: https://github.com/DanielJMaher/compliance-checker/blob/ancillary/compliance_checker/cf/cf.py#L12-L15

Already there,

DanielJMaher:
@lukecampbell @kwilcox

compliance_checker/cf/cf.py
```diff
@@ -723,9 +725,9 @@ def check_ancillary_data(self, ds):
             anc_result = Result(BaseCheck.HIGH, name=('§3.4 Ancillary Variables', k))
             msgs = []
 
-            if not isinstance(anc, str):
+            if not isinstance(anc, basestring):
```
lukecampbell: I think basestring is invalid in py3

daf: https://github.com/DanielJMaher/compliance-checker/blob/ancillary/compliance_checker/cf/cf.py#L12-L15 defines it

DanielJMaher: If you look at https://github.com/DanielJMaher/compliance-checker/blob/ancillary/compliance_checker/cf/cf.py#L12-L15 it catches if it is invalid and then defaults it to str  

This is per Kyle's guidance.  It should be fine in Py3

PR #239, opened 2016-03-17
str format conversion

brianmckenna:
@kwilcox !d causing me problems, think you were after !s or !r in the format conversion


daf:
Yep, this is correct.

kwilcox:
Meant these to be `{:d}`, but `{!s}` or just `{}` works too!

Issue #133, opened 2015-06-25
Come up with checker versioning scheme


daf:
So that other versions of say ACDD (1.2, 1.3) or CF (1.6, 2.x?!) can be checked using this software.  Extends #127

mhidas:
This would also be very useful for IMOS.

lukecampbell:
Any ideas? Like in the modules or something we'll need checkers for different versions. I'm frankly ok with like "acdd11.py" That's how OWSLib did it.

mhidas:
@lukecampbell Given that a new version of a given convention will likely only update a fraction of its rules, it might be easiest to have the checker for the newer version to be subclass of the older one?

In terms of command-line interface, it would be good to be nice to have options, e.g.
 * `cchecker.py -t=cf-16 file.nc` to specify CF-1.6
 * `cchecker.py -t=cf file.nc` to use the latest version by default, or
 * choose the version specified in the `Conventions` global attribute.

lukecampbell:
:+1:

kknee:
Now that we're supporting plug-ins, do we need to worry about this?  Can we not just have an ACDD 1.2 plug-in, ACDD 1.3 plug-in, etc?

daf:
That would mean we'd have to move the "core" checkers (ACDD/CF) out of the main package, which feels wrong - should come with some built in.

The only mechanism we'd have to distinguish with now is plugin name.  A "version" key would be good to have too as another first-order property.  (Version is likely not a good name to use as it implies software version, not spec version)

lukecampbell:
for acdd and cf, I propose we just have modules named "cf17" and "acdd13".

kknee:
that's what I was thinking, and was not trying to indicate that they would be coming out of the main CC package.

daf:
Perhaps rename `cf -> cf16` and `acdd -> acdd12`, with an alias `cf` meaning latest?  Maybe a docker-like label scheme?  `cf:1.6`, `cf:1.7`, `cf:latest` with `:latest` implied if omitted?

mhidas:
:+1:

daf:
@kwilcox has suggested a way to run more than one from a "family", maybe a special token like `cf:all` or `cf:*`?  Maybe comma separated?  `cf:1.6,1.7`?

kwilcox:
[...]

Issue #127, opened 2015-06-19
ACDD 1.3


lukecampbell:
Update compliance checker to support ACDD 1.3
http://wiki.esipfed.org/index.php/Attribute_Convention_for_Data_Discovery_1-3

dpsnowden:
Please ensure that all implemented versions of ACDD are supported.  Do not migrate away from the ACDD 1.2, keep both.  i.e. 

```bash
> compliance-checker input.nc --acdd-1.2
> compliance-checker input.nc --acdd-1.3 
```
both work.

benjwadams:
I have a branch I was working on for this some time ago here: 
https://github.com/benjwadams/compliance-checker/tree/acdd_1.3
Still WIP and does not allow for swapping of versions currently, but probably better than starting from scratch.

benjwadams:
Should be implemented now that #262 is merged. Closing.

PR #240, opened 2016-03-21
Implementation of ACDD 1.3

benjwadams:
Implements #127.

Also adds several new features including the ability to specify version numbers delimited by a colon after the standard name for `compliance-checker` command line tool.  This implements #133 
I.e.
`compliance-checker -t acdd:1.3 file.nc`

Also allows for a 2-tuple with an attribute name and function to be passed into highly recommended, recommended, or suggested attributes, which offers several advantages.  See commit message on   ad716dd


benjwadams:
Why's Travis having issues?  Can't check due to blocking :(

daf:
Looks like some missing dependencies.

benjwadams:
Hmm.   Just checked via other means and looks like `conda` is erroring due to not finding the `pygeoif` package.  Wasn't aware we were using it for the CI process.

daf:
Likely a dependency of a dependency?

benjwadams:
Looks like the `pygeoif` dependency.  I wanted something that could parse and verify WKT that I didn't need to write by hand, and something that was more lightweight than shapely.  If anyone can think of a more suitable library, I'm all ears.

daf:
Ah sorry, hadn't started reviewing your changes so wasn't aware it was a new dep :)

benjwadams:
@DanielJMaher is actually integrating a number of additional checks into this, so I guess I'll close this PR for the time being.

daf:
Can you maybe extract out the command line additions from the ACDD related stuff?

benjwadams:
Going to rebase a few commits to clean up and create a new branch.

daf:
Since pygeoif showed up on conda, I re-ran the builds.

https://travis-ci.org/ioos/compliance-checker/jobs/117552195

@DanielJMaher I think you should've added a new project wide util file in addition to CF's utils, some of which are pretty specific to CF only stuff.

Issue #186, opened 2015-11-03
Pretty names for Checkers


lukecampbell:
As a web developer, I would like to be able to have a list of pretty names for each checker and a detailed description so that I can include the names of the checker in the user interface(s). 

Example:
![screen shot 2015-11-03 at 2 41 56 pm](https://cloud.githubusercontent.com/assets/1155163/10919473/1331ec2c-8239-11e5-82d5-ff97c4b9ce93.png)

lukecampbell:
This is sooooo low priority, it may even just get closed. This is really a presentation issue and as such, it's currently trivial to have the presentation layer (UI) deal with mapping pretty names to the tests, but in the longer scheme of things, it would be nice to be able to dynamically include new plugins, and have the UI automagically update.

daf:
Could see this as additional `__` attributes on each Checker class.

Wants:
- Name
- Description
- URL
- Author
- Version

mhidas:
:+1: :+1: I've been thinking about adding such attributes too. In particular
 * Name and version to be displayed by the `--version` option, and
 * URL (and description) could be added as a footnote to the checker output to point to further information about the conventions being checked.

kwilcox:
How about these for optional attributes on any `Checker` class (with a fallback to something sane implemented in CC).

* `_cc_title`
* `_cc_description`
* `_cc_url`
* `_cc_author`
* `_cc_version`

daf:
Version really has two components - one for spec version (CF **1.6**, ACDD **1.3**) etc, also one for checker internal version (semantic, of course @dpsnowden).  Probably two separate fields.

- `_cc_spec_version`
- `_cc_???_version` ?

kwilcox:
Yes yes yes. I plan to release the `cc-plugin-sgrid` package with the same versions as the spec, but that won't always be the case!

* `_cc_checker_version`
* `_cc_spec_version`

?

daf:
Those names are good.

See also #133 and #127 for similar discussions.  My proposal for a docker-inspired selection mechanism is [here](https://github.com/ioos/compliance-checker/issues/133#issuecomment-171378700).

lukecampbell:
:+1:

kwilcox:
Updated list of specials:

[...]

PR #241, opened 2016-03-22
Allow versioning of standards via command line

benjwadams:
Allows for #133 in specifying version from command line.

Allows versioning of command line specs via `compliance-checker`
command via a colon-separated version if the version is contained in the `_supported_versions`.  I.e. one can specify
 `compliance-checker -t acdd:1.3 file` to run an ACDD-1.3 check.
Versionless specs still work and should revert to the most sensible
fallback version.

### DIFF
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -35,8 +35,15 @@ class CheckSuite(object):
         for x in working_set.iter_entry_points('compliance_checker.suites'):
             try:
                 xl = x.load()
-
                 cls.checkers[xl.name] = xl
+                # set any specific version numbers or pass if we don't have
+                # the _supported_versions attribute
+                try:
+                    for ver in sorted(xl._supported_versions):
+                        cls.checkers["{}:{}".format(xl.name, ver)] = xl
+                except AttributeError:
+                    pass
+
             except Exception as e:
                 print("Could not load", x, ":", e, file=sys.stderr)
 
@@ -66,10 +73,13 @@ class CheckSuite(object):
         if len(checker_names) == 0:
             checker_names = list(self.checkers.keys())
 
-        args = [(name, self.checkers[name]) for name in checker_names if name in self.checkers]
+        # Get the base name of the checker here.  Do not worry about version
+        # handling until later
+        args = [(name, self.checkers[name]) for name in checker_names
+                if name.rsplit(':', 1)[0] in self.checkers]
         valid = []
 
-        all_checked = set([a[1] for a in args])  # only class types
+        all_checked = {a[1] for a in args}  # only class types
         checker_queue = set(args)
 
         while len(checker_queue):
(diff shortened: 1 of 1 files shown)


compliance_checker/suite.py
```diff
@@ -35,8 +35,15 @@ def load_all_available_checkers(cls):
         for x in working_set.iter_entry_points('compliance_checker.suites'):
             try:
                 xl = x.load()
-
                 cls.checkers[xl.name] = xl
+                # set any specific version numbers or pass if we don't have
+                # the _supported_versions attribute
+                try:
+                    for ver in sorted(xl._supported_versions):
```
daf: What is `_supported_versions`?

daf:
Can you write some tests for this please?

Help messages need to be more informative:

```
~/dev/compliance-checker -> ./cchecker.py -t=acdd:1.2 ~/Downloads/Copy.nc
usage: cchecker.py [-h] [--test {gliderdac,acdd,cf,ioos}]
                   [--criteria [{lenient,normal,strict}]] [--verbose]
                   [-f {text,html,json}] [-o OUTPUT] [-V]
cchecker.py: error: argument --test/-t/--test=/-t=: invalid choice: 'acdd:1.2' (choose from 'gliderdac', 'acdd', 'cf', 'ioos')
```

aka, list the versions that are known.  This might involve screwing with argparser a little bit, if it's looking too much, we'll have to punt to another issue.

I think we need to implement https://github.com/ioos/compliance-checker/issues/186#issuecomment-179310390 for the built-in checkers before this can really be tested - need those versions.

compliance_checker/suite.py
```diff
@@ -100,7 +110,22 @@ def run(self, ds, *checker_names):
...
+            except IndexError:
+                # is this assignment even needed?
+                standard_version = None
+                checker = checker_class()
+            else:
+                # Otherwise use the constructor with the version number
+                # specified.  If an invalid version is specified, leave that
+                # up to the class being called to determine that and throw
+                # an exception if necessary
+                checker = checker_class(version=standard_version)
```
daf: Do we have support for this already?

benjwadams: I'm not entirely clear what you meant here.  This tries to pull the standard version by splitting along a delimiter (:).  If it's not there, it uses the default constructor for that specified class.

daf: Passing `version=standard_version`, do any checkers implement this?

benjwadams:
I can modify the help message to display available versions.  Right now, specifying a bad version will spit out the valid options for versions.  Since we haven't actually defined versioned parsers, this will still be the defaults now.  Once we have #127 implemented, there should be actual versioned parsers, but it was requested in #240 that command line additions should have their own separate PR.

compliance_checker/suite.py
benjwadams: Right now it's how I'm enumerating which versions we're supporting.  Since I broke this off from #240, it was the mechanism used.  Basically there are two major choices: create a new class for each new version of a particular standard, or pass along a version argument to the checker constructor.  I opted for the latter since most changes are fairly minor between versions currently.  Perhaps the former is what should have been done, or delegate the base constructor to the various versions.  I wouldn't be averse to a refactor at some point.

benjwadams: Right now it's how I'm enumerating which versions we're supporting.  Since I broke this off from #240, it was the mechanism used.  Basically there are two major choices: create a new class for each new version of a particular standard, or pass along a version argument to the checker constructor.  I opted for the latter since most changes are fairly minor between versions currently.  Perhaps the former is what should have been done, or delegate the base constructor to the various versions.  I wouldn't be averse to a refactor at some point.

benjwadams: Thinking it over, I'll probably refactor before submitting the PR.  It could get messy quick if we need to support many versions.

PR #242, opened 2016-03-23
Adds ACDD 1.3 support

benjwadams:
Implements ACDD 1.3 support per #127.  Running this PR to see if Travis tests pass; may want to squash a number of the commits.


benjwadams:
Py3 compatibility errors. Maybe we should use `tox`?  But then it won't work with conda.  Is there any good automated way to run tests locally against multiple Python versions?

benjwadams:
:clap:  I just read the note about the linter under `py.test`. Shall I ensure the code passes linter specs as well?

benjwadams:
*Do not merge*, there are issues with the command line invocation of compliance checker.

PR #243, opened 2016-03-24
Adds tox and test requirements file

benjwadams:
Adds tox so multiple versions of Python can be tested on local machines.
Does not work with `conda`.  Also added test_requirements.txt file
for testing dependencies.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,0 +1,9 @@
+[tox]
+envlist = py27,py33
+[testenv]
+deps=
+     numpy==1.10.1
+     netCDF4==1.2.1
+     -rrequirements.txt
+     -rtest_requirements.txt
+commands=py.test {toxinidir}/compliance_checker/tests
--- a/README.md
+++ b/README.md
@@ -486,6 +486,17 @@ $ py.test --pylama
 Take a look at the failed tests and fix accordingly. Travis does not run with the `--pylama` flag so you MUSt do this yourself!
 
 
+To run tests for both Python 2 and 3, install `tox` globally via
+```
+pip install tox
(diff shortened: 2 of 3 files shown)


tox.ini
```diff
@@ -0,0 +1,9 @@
+[tox]
+envlist = py27,py33
```
daf: should align this with travis-ci, which uses 3.4 and 3.5, no 3.3.  i looked briefly at tox docs and didn't see a `py35` entry?

benjwadams: Will check this out and modify as needed.

benjwadams: So it looks possible.  However, supporting multiple minor versions of Python with the same major version (i.e. 3.4 and 3.5) might prove to be a little more complicated and have some OS specific steps, assuming one hasn't compiled from source to a directory such as `/opt/`

daf: ok, err to the future (3.5) if that makes this easier.  what are the issues (summary level)?

daf: poke

lukecampbell:
@daf were you able to run tox? I saw a ton of issues last I tried.

daf:
I modified the `tox.ini` locally to do `py27`, `py34`, and `py35`.  `py34` required I installed ubuntu package `libpython3.4-dev` before it worked.  My current ubuntu release does not have an equivalent 3.5 variety, so that part is broken for now, but the new ubuntu just dropped and I think that has it.  Not sure what that means for Travis, though.

lukecampbell:
closed in favor of https://github.com/ioos/compliance-checker/issues/279

PR #253, opened 2016-04-20
ACDD 1.3 implementation

benjwadams:
Implements ACDD 1.3 support (#127), versioned checkers (#186) with option to invoke multiple checkers (#133)


benjwadams:
Hmm, thought the `ordered-set` requirement was in Conda.  That's causing Travis to fail.

https://anaconda.org/auto/ordered-set

benjwadams:
OK, also created a PR to conda-recipes to fix conda install issues: https://github.com/ioos/conda-recipes/pull/821.  Tests should hopefully work after that.

daf:
I don't see why the OrderedSet business/changes to argparse are necessary.  `nargs='+'` is standard and works fine by specifying multiple `--test <something>` arguments on the command line:

```
$ python cchecker.py --test cf --test acdd compliance_checker/tests/data/2dim-grid.nc
```

runs both acdd and cf.  I'd prefer not to add a new dep and nonstandard behavior if possible.

benjwadams:
It isn't strictly necessary.  I merely wanted to avoid repeated tests if they were specified.   Also, I thought we were going by comma separated args per your comment in https://github.com/ioos/compliance-checker/issues/133.

Also running:
```
python cchecker.py --test cf --test acdd compliance_checker/tests/data/2dim-grid.nc
``` 
yields
```
usage: cchecker.py [-h] --test TEST [TEST ...]
                   [--criteria [{lenient,normal,strict}]] [--verbose]
                   [-f {text,html,json}] [-o OUTPUT] [-V]
cchecker.py: error: argument --test/-t/--test=/-t=: Test arguments ['compliance_checker/tests/data/2dim-grid.nc'] not in valid values of ['acdd', 'acdd:1.1', 'acdd:1.3', 'acdd:latest', 'cf', 'cf:1.6', 'cf:latest', 'ioos', 'ioos:0.1', 'ioos:latest']
```

when `nargs='+'`.  Running `python cchecker.py --test cf --test acdd -- compliance_checker/tests/data/2dim-grid.nc` has the desired effect, but it might not be obvious to end users.  What do you think the best way forward is?

daf:
Ah sorry, didn't mean to distract with my comment in #133.

I'm not getting that behavior on the command line (I run the same command and it works ok for me, python 2.7.10 and 3.4.3 w/ zsh), but the `--` syntax is not uncommon.  I'm not concerned with repetition, in fact I think `nargs='+'` already handles it?  (If i do `--test acdd` twice, it only runs it once).

I'd prefer leaving it as `nargs='+'` if possible.

benjwadams:
Ahhh.  Just saw that behavior with nargs and repetition now.  Had I known that, I would have excluded OrderedSet.  Ah well, all's well that ends well, I suppose.

benjwadams:
FYI, multiple `-t`/`--test` invocations appear to only override the last one.

daf:
> FYI, multiple -t/--test invocations appear to only override the last one.

What do you mean?

benjwadams:
Invoking `python cchecker.py --test acdd --test cf -- compliance_checker/tests/data/2dim-grid.nc` will result in only the CF test being run.

daf:
uh... works for me

```
python cchecker.py --test acdd --test cf  compliance_checker/tests/data/2dim-grid.nc
...
                     The dataset scored 13 out of 51 points                     
                             during the acdd check                              
...

                     The dataset scored 33 out of 34 points                     
                              during the cf check                               
...
```

benjwadams:
*Shrug* maybe it's some behavior of the builtins on newer Python versions.  I tested against 2.7.5 and 3.4.3 and got different behavior from you.  In any case, someone can just invoke the tests space separated to get similar behavior.

cchecker.py
```diff
@@ -18,23 +17,21 @@ class SupportedTests(argparse.Action):
 A custom argparse Action to ensure only the supported tests are passed in
         """
         def __call__(self, parser, namespace, values, option_string=None):
```
daf: is this custom handler needed anymore rather than the built in `nargs='+'` / `choices=` behavior?

benjwadams: It appears not.  I'll update that.

benjwadams:
Closing in favor of #262

PR #260, opened 2016-04-22
Add limit positional parameter

benjwadams:
Adds 'limit' positional parameter, which  is required by the `json_output` function.


benjwadams:
Meant to merge into @lukecampbell's branch, closing.

PR #264, opened 2016-04-26
Follows redirects before looking at status code, refactors code

### LINKED ISSUE #263: Redirect Issue
Some THREDD instances are behind a reverse-proxy and do some 301 redirecting when copy-pasting the OPeNDAP URL into input for Compliance-Checker. It would be good to have CChecker recognize this and get the forwarded address. I.e. follow-redirects.

benjwadams:
Attempts to fully follow redirects prior to opening a dataset, avoiding
erroneous errors on valid URLs.  Also refactors some code in `suite.py`.
Fixes #263.

### DIFF
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -269,11 +269,10 @@ class CheckSuite(object):
         score_list = []
         score_only_list = []
 
-        for v in range(len(groups)):
-            score_list.append([groups[v].name, groups[v].weight, groups[v].value,
-                               groups[v].children])
-            if groups[v].weight >= limit:
-                score_only_list.append(groups[v].value)
+        for g in groups:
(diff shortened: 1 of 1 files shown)


compliance_checker/suite.py
```diff
@@ -269,11 +269,10 @@ def get_points(self, groups, limit):
         score_list = []
         score_only_list = []
 
-        for v in range(len(groups)):
-            score_list.append([groups[v].name, groups[v].weight, groups[v].value,
-                               groups[v].children])
-            if groups[v].weight >= limit:
-                score_only_list.append(groups[v].value)
+        for g in groups:
```
daf: there was probably a reason this was indexed (v was used somehow) but can't remember why.

PR #266, opened 2016-05-01
Fixed 'Usage (from Python code)' section

MBARIMike:
This reflects the code I needed to write to use compliance-checker from Python.


benjwadams:
Looks good :+1:

PR #270, opened 2016-05-13
Force Results.value to be bool or 2-tuple

mhidas:
The value of a Result object is expected to be either of type `tuple` or `bool`, but there is nothing enforcing this. While this is true most of the time, in the IMOS check suite we've created some Results having a value of type `numpy.bool_` (e.g. the result of `numpy.all()`). This didn't cause a problem earlier, but since da674bcc5dcc5745394e7dea1eafd9b5a376a54c it breaks because `_translate_value()` (in suite.py) returns such a value unchanged, while its output is expected to always be a tuple.

I've added a unittest to demonstrate this issue, and suggested a change to the `Result` object to ensure that the `value` attribute is always one of the expected types.

### DIFF
--- a/compliance_checker/base.py
+++ b/compliance_checker/base.py
@@ -89,7 +89,8 @@ class Result(object):
     Holds the result of a check method.
 
     Stores such information as the check's value (True, False, a 2-tuple of (pass, total) or None for a skip),
-    weight of the check, any granular messages, or a hierarchy of results.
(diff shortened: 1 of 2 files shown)


mhidas:
Thanks @lukecampbell !
However, I just realised I forgot about the case where the result value is `None`, indicating a skipped check according to the comment in the code. Is this actually used anywhere?

mhidas:
See #271

PR #271, opened 2016-05-17
Allow Result.value = None

mhidas:
Comment in the definition of the `Result` object (in base.py) suggest a value of `None` can be used to signify a skipped test. I missed this in #270. Here is an update to allow for that, if it is deemed relevant.


lukecampbell:
Hmmm... I guess I would say returning None as a value would be undefined behavior, I'm not sure what the expectation would be if the value is None, but I like the error handling.

Issue #245, opened 2016-04-06
Strange encodings error with windows and python 3


rsignell-usgs:
When I try this test on win64, I get some strange "encodings" error:
```
[TEST] C:\Users\rsignell>conda create --name TEST --channel conda-forge python=3.4 compliance-checker

[TEST] C:\Users\rsignell>compliance-checker --test cf http://geoport.whoi.edu/thredds/dodsC/clay/usgs/users/jcwarner/Projects/Sandy/triple_nest/00_dir_NYB05.ncml

...
indexed_ragged_array                    :2:     0/0
missing_data                            :2:     0/0
orthogonal_multidim_array               :2:     0/0
                                 High Priority

--------------------------------------------------------------------------------

    Name                            :Priority: Score
Traceback (most recent call last):
  File "C:\Users\rsignell\AppData\Local\Continuum\Miniconda2\envs\TEST\Scripts\c
ompliance-checker-script.py", line 5, in <module>
    sys.exit(main())
  File "C:\Users\rsignell\AppData\Local\Continuum\Miniconda2\envs\TEST\Scripts\c
checker.py", line 41, in main
    args.format)
  File "C:\Users\rsignell\AppData\Local\Continuum\Miniconda2\envs\TEST\lib\site-
packages\compliance_checker\runner.py", line 44, in run_checker
    groups = cls.stdout_output(cs, score_groups, verbose, limit)
  File "C:\Users\rsignell\AppData\Local\Continuum\Miniconda2\envs\TEST\lib\site-
packages\compliance_checker\runner.py", line 74, in stdout_output
    cs.non_verbose_output_generation(score_list, groups, limit, points, out_of)
  File "C:\Users\rsignell\AppData\Local\Continuum\Miniconda2\envs\TEST\lib\site-
packages\compliance_checker\suite.py", line 311, in non_verbose_output_generatio
n
    print('%-40s:%s:%6s/%1s'  % (score_list[x][0][0:39], score_list[x][1], score
_list[x][2][0], score_list[x][2][1]))
  File "C:\Users\rsignell\AppData\Local\Continuum\Miniconda2\envs\TEST\lib\encod
ings\cp437.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_map)[0]
UnicodeEncodeError: 'charmap' codec can't encode character '\xa7' in position 0:
 character maps to <undefined>

[TEST] C:\Users\rsignell>
```

benjwadams:
[...]

PR #274, opened 2016-06-02
Avoid UnicodeEncodeError when checker output is redirected

mhidas:
The checker is running fine in my (Linux) terminal, but when I try to redirect the output to a file, this happens:
```
$ compliance-checker -t=cf file.nc >/tmp/cc.out
Running Compliance Checker on the dataset from: file.nc
  File "/sw/anaconda/envs/cc/bin/compliance-checker", line 9, in <module>
    load_entry_point('compliance-checker', 'console_scripts', 'compliance-checker')()
  File "/sw/chef/src/compliance-checker/cchecker.py", line 56, in main
    args.format)
  File "/sw/chef/src/compliance-checker/compliance_checker/runner.py", line 44, in run_checker
    groups = cls.stdout_output(cs, score_groups, verbose, limit)
  File "/sw/chef/src/compliance-checker/compliance_checker/runner.py", line 74, in stdout_output
    cs.non_verbose_output_generation(score_list, groups, limit, points, out_of)
  File "/sw/chef/src/compliance-checker/compliance_checker/suite.py", line 340, in non_verbose_output_generation
    print('%-40s:%s:%6s/%1s'  % (score_list[x][0][0:39], score_list[x][1], score_list[x][2][0], score_list[x][2][1]))
UnicodeEncodeError: 'ascii' codec can't encode character u'\xa7' in position 0: ordinal not in range(128)
```

Apparently the problem is that when I redirect output, Python sets the output encoding to `None` (which presumably defaults to ascii). Several solutions are suggested (e.g. [here](http://stackoverflow.com/questions/492483/setting-the-correct-encoding-when-piping-stdout-in-python), and [here](http://chase-seibert.github.io/blog/2014/01/12/python-unicode-console-output.html)). One of them is to force `stdout` and `stderr` encoding to unicode, which is what this PR does. Since we're explicitly trying to print unicode, this seems sensible.

Alternatively, this can be done using an environment variable (`PYTHONIOENCODING='UTF-8'`), so if you prefer not to set this in the code, we can just use that.

### DIFF
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -21,6 +21,13 @@ except:
 from datetime import datetime
 import requests
 import textwrap
+import codecs
+
+# Ensure output is encoded as Unicode when checker output is redirected or piped
+if sys.stdout.encoding is None:
+    sys.stdout = codecs.getwriter('utf8')(sys.stdout)
+if sys.stderr.encoding is None:
+    sys.stderr = codecs.getwriter('utf8')(sys.stderr)
 
 
 class CheckSuite(object):


mhidas:
Also related to #245

lukecampbell:
@benjwadams I'm in DC this week, could you look at this? I think you also dealt with a bunch of encoding issues recently, so it's probably fresher in your mind than mine at this point.

benjwadams:
Sure thing. @mhidas, I also encountered this.  Nice to see you found a fix.  I'll review it and merge it shortly.

daf:
:+1: thanks for the fix!

PR #276, opened 2016-06-10
Makes the variable detection more hardened

lukecampbell:
- cf_role doesn't have to be just timeseries_id, it can be profile_id
  etc.
- This just removes variables that define cf_role from the potential of
  being identified as geophysical_variables


lukecampbell:
This was in response to a GCOOS dataset that I was working with, and I noticed the checker was trying to perform attribute checks on a cf_role variable.

PR #278, opened 2016-06-13
Fixes default behavior for test selection

lukecampbell:
- Attempting to select one or more tests would yield a tuple append
  error.


lukecampbell:
Ok, this fixes it for sure. I can only speculate that stale files or a stale build on my end led to the multiple test selection working before merging into master...

PR #279, opened 2016-06-23
Fixes python3 setup errors

lukecampbell:
- Adds tox

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -32,13 +32,14 @@ setup(
     ],
     include_package_data = True,
     scripts=['cchecker.py'],
+    # Note: Do not use colons in the entry-point keys. Python 3 reserves
+    # portions of the key after a colon for special use.
     entry_points         = {
         'console_scripts': [
             'compliance-checker = cchecker:main'
         ],
         'compliance_checker.suites': [
             'cf = compliance_checker.cf.cf:CFBaseCheck',
-            'acdd = compliance_checker.acdd:ACDD1_1Check',
             'acdd-1.1 = compliance_checker.acdd:ACDD1_1Check',
             'acdd-1.3 = compliance_checker.acdd:ACDD1_3Check',
             'ioos = compliance_checker.ioos:IOOSBaseCheck',
--- a/setup.py
+++ b/setup.py
@@ -39,8 +39,8 @@ setup(
         'compliance_checker.suites': [
             'cf = compliance_checker.cf.cf:CFBaseCheck',
             'acdd = compliance_checker.acdd:ACDD1_1Check',
-            'acdd:1.1 = compliance_checker.acdd:ACDD1_1Check',
-            'acdd:1.3 = compliance_checker.acdd:ACDD1_3Check',
+            'acdd-1.1 = compliance_checker.acdd:ACDD1_1Check',
+            'acdd-1.3 = compliance_checker.acdd:ACDD1_3Check',
             'ioos = compliance_checker.ioos:IOOSBaseCheck',
         ]
     },
--- a/setup.py
+++ b/setup.py
@@ -32,8 +32,15 @@ setup(
     ],
     include_package_data = True,
     scripts=['cchecker.py'],
+
     # Note: Do not use colons in the entry-point keys. Python 3 reserves
     # portions of the key after a colon for special use.
+
+    # Note: The entry point names are not used at all. All methods in the
+    # compliance checker use class attributes to determine the checker's name
+    # and version. But, an entry point must be defined for each plugin to be
+    # loaded.
+
     entry_points         = {
         'console_scripts': [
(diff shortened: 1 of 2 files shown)


lukecampbell:
If you currently tried to pip install compliance-checker using the latest version of pip and python3, you will get:

```
Exception:
  File "/Users/lcampbell/.venvburrito/lib/python2.7/site-packages/pip-8.1.2-py2.7.egg/pip/basecommand.py", line 215, in main
    status = self.run(options, args)
  File "/Users/lcampbell/.venvburrito/lib/python2.7/site-packages/pip-8.1.2-py2.7.egg/pip/commands/install.py", line 317, in run
    prefix=options.prefix_path,
  File "/Users/lcampbell/.venvburrito/lib/python2.7/site-packages/pip-8.1.2-py2.7.egg/pip/req/req_set.py", line 742, in install
    **kwargs
  File "/Users/lcampbell/.venvburrito/lib/python2.7/site-packages/pip-8.1.2-py2.7.egg/pip/req/req_install.py", line 831, in install
    self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
  File "/Users/lcampbell/.venvburrito/lib/python2.7/site-packages/pip-8.1.2-py2.7.egg/pip/req/req_install.py", line 1032, in move_wheel_files
    isolated=self.isolated,
  File "/Users/lcampbell/.venvburrito/lib/python2.7/site-packages/pip-8.1.2-py2.7.egg/pip/wheel.py", line 352, in move_wheel_files
    console, gui = get_entrypoints(ep_file)
  File "/Users/lcampbell/.venvburrito/lib/python2.7/site-packages/pip-8.1.2-py2.7.egg/pip/wheel.py", line 229, in get_entrypoints
    cp.readfp(data)
  File "/usr/local/Cellar/python3/3.5.0/Frameworks/Python.framework/Versions/3.5/lib/python3.5/configparser.py", line 761, in readfp
    self.read_file(fp, source=filename)
  File "/usr/local/Cellar/python3/3.5.0/Frameworks/Python.framework/Versions/3.5/lib/python3.5/configparser.py", line 716, in read_file
    self._read(f, source)
  File "/usr/local/Cellar/python3/3.5.0/Frameworks/Python.framework/Versions/3.5/lib/python3.5/configparser.py", line 1090, in _read
    fpname, lineno)
configparser.DuplicateOptionError: While reading from '<???>' [line  3]: option 'acdd' in section 'compliance_checker.suites' already exists
```

the configparser treats the ':' as an optional argument but anything to the left of it, is the name of the entry, so using a colon to name unique entires won't work any more.

lukecampbell:
@kwilcox @daf @RSully @ocefpaf

RSully:
Hm I've been using this under Python 3 without changes, is this due to a dependency being bumped maybe?

setup.py
```diff
@@ -32,13 +32,14 @@ def readme():
     ],
     include_package_data = True,
     scripts=['cchecker.py'],
+    # Note: Do not use colons in the entry-point keys. Python 3 reserves
+    # portions of the key after a colon for special use.
```
daf: also plz add something like "the strings on the left below *DO NOT MATTER*, the real checker names are read from class attributes"

lukecampbell:
I've had python3 virtualenvs where this wasn't a problem too. I'm not entirely sure which versions are and aren't affected.

RSully:
If I have time I'll experiment and see if any package updates break it, but that might not be until Monday. I'd hold off merging this until we can track down the cause at least. The other thing that stands out to me is the mention of python2.7 in your traceback.

lukecampbell:
That's just that traceback, if I make a fresh virtualenv without virtualenvburrito it's still there, but different traceback paths. Will paste tomorrow

kwilcox:
Explanation: https://setuptools.readthedocs.io/en/latest/setuptools.html#dynamic-discovery-of-services-and-plugins

It mentions the colon is only reserved in the "value" portion but this effects the "name" portion. Bummer. Colons look better.

lukecampbell:
@RSully was right about virtualenvburrito mucking up python3. I would still like to remove the colons though, it doesn't negatively impact anyone who was using the compliance-checker before and all the code from the command line to the web interface relies on class attributes not entry_point names

daf:
As long as a comment to this affect also gets added to the comment you added, because without that piece of info, I'd be scratching my head a while at how plugin discovery is done.

RSully:
Fine by me then. I can't comment much on Tox; it uses virtualenv/pip and I've been using conda. Would be nice if there was something that could just run off of the existing travis file instead.

lukecampbell:
That PR had grown stale and it didn't get patched to work with 3.5

I can add in 3.4 to this tox file, but I don't have a 3.4 interpreter

Issue #289, opened 2016-07-27
CF check_alternative_coordinates seems to do nothing useful?


mhidas:
The `check_alternative_coordinates` method in [cf.py](https://github.com/ioos/compliance-checker/blob/master/compliance_checker/cf/cf.py#L1917) returns either a True result, or nothing. This may seem a bit of a pessimistic view, but what good is a compliance check that nothing can fail?

Looking at the related [CF text](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#alternative-coordinates), there's not much there to check, i.e. nothing that must or must not be. Maybe this whole method can be removed (along with its unittest)?

mhidas:
Fixed in #294

PR #293, opened 2016-09-01
Remove check alternative coordinates

Bobfrat:
@lukecampbell removes the test for alternative coordinates as described here https://github.com/ioos/compliance-checker/issues/289


daf:
uh this just removed the test and not the actual check in cf checker.

also just delete, it can be found again in history.

Issue #247, opened 2016-04-07
Mechanism for dynamically updating standard name table


lukecampbell:
CF Standard Names table updates much faster than releases for compliance checker so we need a way to dynamically update the standard name table for the checker.

mbiddle-nodc:
Not sure if this helps, but I parse the standard_name_table attribute for the table version. Then I use
`url = "http://cfconventions.org/Data/cf-standard-names/%s/src/cf-standard-name-table.xml"% std_version`
to get the xml document with the CF standard_name table for whatever version the file is using. For example, if the global attribute is
`                :standard_name_vocabulary = "CF Standard Name Table v30" ;`
Then, the url composed would be
`http://cfconventions.org/Data/cf-standard-names/30/src/cf-standard-name-table.xml`

This allows the CF standard_name testing to be a little bit more dynamic.

lukecampbell:
That's an awesome idea!

daf:
I'd recommend a base version be packaged, with a mechanism to download/cache this live one to `$XDG_HOME_CACHE` or similar (not sure what that is on Windows, expecting python to have this covered for us).  We should not have a connectivity requirement for testing local files.

benjwadams:
This might come in handy, it generates a CSV from a standard name table version (v29 in the gist).  Since it uses pandas, we could easily output it to a Python data structure instead.

https://gist.github.com/benjwadams/2f9c4db1aa8db60128e7

Bobfrat:
So is the consensus here to try to download the standard name table that's defined in ```:standard_name_vocabulary``` and if that fails, fall to the packaged version? 

It sounds like @lukecampbell wants the ability to update the packaged version as well. I could add a command line argument to update to the latest version.

kwilcox:
You could just update the packaged file with a new version periodically

lukecampbell:
Ok, the code is ready, now we just need to update the README and we can close this.

daf:
I'd like it to print the current standard name table version (and if it came from local) on run if possible.  Will help future debugging sessions.

lukecampbell:
actually yeah, that's pretty critical I think. Because it affects output results somewhat dramatically.

PR #296, opened 2016-09-08
Dynamically download specified cf standard name table

Bobfrat:
@lukecampbell @daf @kwilcox Can one of you guys review please?

This PR addresses https://github.com/ioos/compliance-checker/issues/247.

During the CF test, if a file has a particular version of the cf standard name table specified in the global attributes (i.e. ```:standard_name_vocabulary = "CF Standard Name Table v30" ;```) that doesn't match the packaged version, it will try to download the specified version. If it fails, it will fall back to packaged version.

This PR also adds ability to update the packaged version by using the 'd' switch and passing in the version you want to fetch.
```
python cchecker.py -d 34
```

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -9,7 +9,7 @@ import os
 
 from compliance_checker.base import BaseCheck, BaseNCCheck, score_group, Result
 from compliance_checker.cf.appendix_d import dimless_vertical_coordinates
-from compliance_checker.cf.util import NCGraph, StandardNameTable, units_known, units_convertible, units_temporal, map_axes, find_coord_vars, is_time_variable, is_vertical_coordinate, download_cf_standard_name_table, _possiblet, _possiblez, _possiblex, _possibley, _possibleaxis, _possibleaxisunits
+from compliance_checker.cf.util import NCGraph, StandardNameTable, units_known, units_convertible, units_temporal, map_axes, find_coord_vars, is_time_variable, is_vertical_coordinate, create_cached_data_dir, download_cf_standard_name_table, _possiblet, _possiblez, _possiblex, _possibley, _possibleaxis, _possibleaxisunits
 
 try:
     basestring
@@ -133,7 +133,7 @@ class CFBaseCheck(BaseCheck):
         self._metadata_vars  = defaultdict(list)
         self._boundary_vars  = defaultdict(dict)
 
-        self._std_names      = StandardNameTable('data/cf-standard-name-table.xml')
+        self._std_names      = StandardNameTable()
 
     ################################################################################
     #
@@ -159,7 +159,7 @@ class CFBaseCheck(BaseCheck):
 
         # Try to parse this attribute to get version
         version = None
-        if 'CF Standard Name Table' in standard_name_vocabulary:
+        if 'cf standard name table' in standard_name_vocabulary.lower():
             version = standard_name_vocabulary.split()[-1]
         else:
             # Can't parse the attribute, use the packaged version
@@ -174,12 +174,12 @@ class CFBaseCheck(BaseCheck):
 
         # Try to download the version specified
         try:
-            location = 'compliance_checker/data/cf-standard-name-table-{0}.xml'.format(version)  # (don't overwrite package location)
+            data_directory = create_cached_data_dir()
+            location = os.path.join(data_directory, 'cf-standard-name-table-test-{0}.xml'.format(version))
             # Did we already download this before?
             if not os.path.isfile(location):
                 download_cf_standard_name_table(version, location)
-            resource_name = location.split('compliance_checker/')[1]  # relative to package
-            self._std_names = StandardNameTable(resource_name)
+            self._std_names = StandardNameTable(location)
             return 1
         except Exception:
             # There was an error downloading the CF table. That's ok, we'll just use the packaged version
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -5,10 +5,11 @@ import re
 from functools import wraps
 from collections import defaultdict
 import numpy as np
+import os
 
 from compliance_checker.base import BaseCheck, BaseNCCheck, score_group, Result
 from compliance_checker.cf.appendix_d import dimless_vertical_coordinates
-from compliance_checker.cf.util import NCGraph, StandardNameTable, units_known, units_convertible, units_temporal, map_axes, find_coord_vars, is_time_variable, is_vertical_coordinate, _possiblet, _possiblez, _possiblex, _possibley, _possibleaxis, _possibleaxisunits
+from compliance_checker.cf.util import NCGraph, StandardNameTable, units_known, units_convertible, units_temporal, map_axes, find_coord_vars, is_time_variable, is_vertical_coordinate, download_cf_standard_name_table, _possiblet, _possiblez, _possiblex, _possibley, _possibleaxis, _possibleaxisunits
 
 try:
     basestring
@@ -132,7 +133,7 @@ class CFBaseCheck(BaseCheck):
         self._metadata_vars  = defaultdict(list)
         self._boundary_vars  = defaultdict(dict)
 
-        self._std_names      = StandardNameTable('cf-standard-name-table.xml')
+        self._std_names      = StandardNameTable('data/cf-standard-name-table.xml')
 
     ################################################################################
     #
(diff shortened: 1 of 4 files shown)


daf:
I'll have some comments this afternoon.

compliance_checker/cf/cf.py
```diff
@@ -132,7 +133,7 @@ def __init__(self):
         self._metadata_vars  = defaultdict(list)
         self._boundary_vars  = defaultdict(dict)
 
-        self._std_names      = StandardNameTable('cf-standard-name-table.xml')
+        self._std_names      = StandardNameTable('data/cf-standard-name-table.xml')
```
lukecampbell: Could you use `pkg_resources` here? This path is only valid if you're running this from the project root, anywhere else, this path is very likely to be invalid.

compliance_checker/cf/cf.py
```diff
@@ -146,6 +147,43 @@ def setup(self, ds):
...
+        """
+        Parse out the :standard_name_vocabulary attribute and download that version of
+        the cf standard name table
+        """
+        # Get the standard name vocab
+        standard_name_vocabulary = getattr(ds, 'standard_name_vocabulary', '')
+
+        # Try to parse this attribute to get version
+        version = None
+        if 'CF Standard Name Table' in standard_name_vocabulary:
```
lukecampbell: could you check the strings in lowercase?

compliance_checker/cf/cf.py
```diff
@@ -146,6 +147,43 @@ def setup(self, ds):
...
+        if version.startswith('v'):  # i.e 'v34' -> '34' drop the v
+            version = version[1:]
+
+        # If the packaged version is what we're after, then we're good
+        if version == self._std_names._version:
+            return 0
+
+        # Try to download the version specified
+        try:
+            location = 'compliance_checker/data/cf-standard-name-table-{0}.xml'.format(version)  # (don't overwrite package location)
```
lukecampbell: Please use `pkg_resources`

compliance_checker/cf/util.py
```diff
@@ -1,4 +1,9 @@
 import itertools
+try:
+    from urllib import urlretrieve
+except:
+    from urllib.request import urlretrieve
```
lukecampbell: `six` has a solution for this.

daf: don't use urllib

compliance_checker/cf/util.py
```diff
@@ -294,6 +300,27 @@ def __iter__(self):
...
+    '''
+
+    if location is None:  # This case occurs when updating the packaged version from command line
+        location = 'compliance_checker/data/cf-standard-name-table.xml'
+
+    url = "http://cfconventions.org/Data/cf-standard-names/{0}/src/cf-standard-name-table.xml".format(version)
+    rhead = requests.head(url, allow_redirects=True)
+    if rhead.status_code == 200:
+        print("Downloading cf-standard-names table version {0} from: {1}".format(version, url))
+        urlretrieve(url, location)
```
daf: you use requests right above, why swithc to use urllib?  just use requests.  also don't make a head request, check its status and then continue - just `requests.get`

compliance_checker/cf/util.py
```diff
@@ -294,6 +300,27 @@ def __iter__(self):
...
+    else:
+        raise Exception("Could not find specified cf-standard-names table version {0} from: {1}".format(version, url))
```
daf: use requests `raise_for_status` or use a more specific Exception class

compliance_checker/cf/cf.py
daf: eh see below

daf:
`pkg_requests` is unlikely to help with this scenario - consider a system-wide installed compliance-checker python package - you'd use `sudo pip` or similar, so the user won't have write access to wherever the file is.

Instead, use [XDG directories](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html), specifically `$XDG_DATA_DIRS`, to create a compliance-checker directory where local data can be cached for a user.  Something like [pyxdg](https://freedesktop.org/wiki/Software/pyxdg/) (i have no idea if this supports windows, i know the concept is there for that platform).  Typically this is `~/.local/share/...`

Then, the cc logic should check that location for an updated standard name file - if not, go to regular pkg_resources checking.

daf:
Also looks like you could [write a custom importer for pkg_resources](http://setuptools.readthedocs.io/en/latest/pkg_resources.html#supporting-custom-importers) to look really slick but probably unnecessary.

lukecampbell:
XDG does seem heavy compared to

```python
if not os.path.exists(os.path.expand('~/.cc-cache')):
    os.makedir(os.path.expanduser('~/.cc-cache'))
    
# Download standard name table definitions here
# Load standard name table definitions from here
```

daf:
At least use the same path if possible: `~/.local/share/compliance-checker` ([quickly googled reasons](https://ploum.net/207-modify-your-application-to-use-xdg-folders/))

It's well supported:

```
~ -> ls -l .local/share | wc -l
45
```

ocefpaf:
> XDG does seem heavy compared to

But standard across multiple platforms :smile: 

I would use `XDG_DATA_HOME` though. Something like:

```python
writable_directory = os.path.join(os.path.expanduser('~'), '.local', 'share')
data_directory = os.path.join(os.environ.get("XDG_DATA_HOME", writable_directory),
                                             'compliance-checker')
```

BTW that expands to AppData on Windows if I am not mistaken.

daf:
at the very least respect the (usually not set) os environment var `$XDG_DATA_HOME`:

`os.path.expanduser(os.environ.get('XDG_DATA_HOME', '~/.local/share'))`

lukecampbell:
yeah that's all fine, but I don't want to go down the rabbit hole that the spec outlines.

> There is a single base directory relative to which user-specific data files should be written. This directory is defined by the environment variable $XDG_DATA_HOME.

`XDG_DATA_HOME` is sparsely defined.

> $XDG_RUNTIME_DIR defines the base directory relative to which user-specific non-essential runtime files and other file objects (such as sockets, named pipes, ...) should be stored. The directory MUST be owned by the user, and he MUST be the only one having read and write access to it. Its Unix access mode MUST be 0700.

I don't want to have to deal with use-case permissions and special handling of files... just to download an XML file that compliance checker needs.

> The directory MUST be on a local file system and not shared with any other system. The directory MUST by fully-featured by the standards of the operating system. More specifically, on Unix-like operating systems AF_UNIX sockets, symbolic links, hard links, proper permissions, file locking, sparse files, memory mapping, file change notifications, a reliable hard link count must be supported, and no restrictions on the file name character set should be imposed. Files in this directory MAY be subjected to periodic clean-up. To ensure that your files are not removed, they should have their access time timestamp modified at least once every 6 hours of monotonic time or the 'sticky' bit should be set on the file.

Again, more checks and stuff that seem extremely heavy handed for what we're trying to do.

daf:
So use a library then?

- https://pypi.org/project/xdg/
- https://launchpad.net/dirspec
- https://pypi.org/project/pyxdg/

etc

ocefpaf:
@lukecampbell you do not need to enforce/check any of the `XDG` rules. In fact `writable_directory` is all of the above it `XDG_DATA_HOME` is not defined.

lukecampbell:
But is it really necessary to introduce a new dependency, go down this mess of integrating when, all we want to do is download a file? I like @ocefpaf loose implementation, it's like 3 lines.

ocefpaf:
> I like @ocefpaf loose implementation, it's like 3 lines.

Just to be fair with some unknown author that is not my implementation but a pattern I have copied-and-pasted in many places that I forgot the original source.

compliance_checker/cf/util.py
```diff
@@ -309,17 +315,30 @@ def download_cf_standard_name_table(version, location=None):
...
+def create_cached_data_dir():
+    '''
+    Returns the path to the data directory to download CF standard names.
+    Use $XDG_DATA_HOME.
+    '''
+    writable_directory = os.path.join(os.path.expanduser('~'), '.local', 'share')
+    data_directory = os.path.join(os.environ.get("XDG_DATA_HOME", writable_directory),
+                                  'compliance-checker')
+    if not os.path.isdir(data_directory):
+        os.mkdir(data_directory)
```
daf: use `os.makedirs` to recursively make them (this will fix travis)

Bobfrat:
Thanks all for the comments and tips. I implemented @ocefpaf 's suggestion. Let me know what you guys think. Also, there is a new version of CF standard name table (v35). Want me to push the new version in this PR?

Issue #283, opened 2016-06-29
Metadata Only Checks


lukecampbell:
We need to add a feature where we can check a file's compliance by only looking at the CDL. 

We can easily accomplish this by using ncgen to generate a netCDF file and then running the checker on that. But we need to add some small checks and stuff to make sure the dimension sizes get changed before the file is generated, otherwise we can wind up with files that are easily gigabytes or larger.

PR #297, opened 2016-09-13
Add ability to run compliance checker on .cdl files

Bobfrat:
@lukecampbell please review. 
This addresses https://github.com/ioos/compliance-checker/issues/283 but without the dimension size checks. That's implemented on cc-web.


compliance_checker/suite.py
```diff
@@ -528,11 +543,38 @@ def is_binary_string(bts):
...
+                            lat:standard_name = "latitude" ;
+                            lat:long_name = "station latitude" ;
+                    etc...
+
+                :param str filename: Absolute path of file to check
+                :param str data: First chuck of data from file to check
+                '''
+                if os.path.splitext(filename)[-1] == '.cdl':
+                    return True
+                if data.startswith('netcdf') or 'dimensions' in data:
```
lukecampbell: There's no guarantee that the file doesn't start with whitespace or a newline.

lukecampbell: We'll leave it for now, we don't have to handle all the cases of CDL formatting. Just the ones where ncdump generates the CDL.

Issue #202, opened 2016-02-03
Create a way to skip a check


DanielJMaher:
We could use some way to pass a test to prevent one failed test from having a trickle down effect causing more failures that aren't real.

Not time critical, but cam up in conversation.

kwilcox:
Already ran into the need for this. My temp solution:

```python
def check1(self, ds):
    assert ds.Conventions.lower() == "cf-2.0"

def check2(self, ds):
    got, want = self.check1(ds).value
    if got != want:
        return None

    ... go about your test
```

Obviously not ideal if `check1` is intensive... but works. Without a dependency graph, this won't be easy.

daf:
Dependency graph idea:

```python
@depends('check1')
def check2(self, ds):
    ....
```

CheckSuite would have to scan for checks as usual, find these deps, make a graph, figure out if the graph is traversable, then execute in correct order.  Also give warning/error out if unknown dep shows up.

DanielJMaher:
The proper way to handle this is to split this into two tasks.  The first is to create a decorator for skipping tests.  The second is to create the dependency graph for each check/subcheck.  The second step is a very time intensive task and requires a certain level of domain knowledge.  The first is rather simple and can be set to a command line prompt for now.  After the dependency graph is created the skip decorator can be modified to work with the dependency graph.

For now we should just create the command line skip test functionality and table the dependency graph for another milestone.  Having the command line skip will provide a way to deal with data intensive checks (e.g. Missing _FillValue), which can be troublesome for some users.

kwilcox:
The verbiage in this issue is confusing. We want to skip a `check` not a `test! :confused:

How would the decorator for skipping checks work? Would we decorate each check with a list of tags and have the ability to skip certain tags when running? Would we supply the actual check function names to skip when running?

DanielJMaher:
something simple like:

compliance-checker -t cf --skip "3.1.2, 4.5, 2.2" <path to file>

Then the decorator just checks "is my check in the list of skips? yes, skip. no, check."

I made those check numbers up, so don't worry if those aren't checks.
also i edited my comment for the word test, but the content is still the same.

kwilcox:
How would the skips be presented in the final score? A new "skips" section? Would each skip just reduce the "total number" of the suite?

DanielJMaher:
I would lean towards reducing the total number.  The user knows they skipped the check and anyone reviewing the scoring should notice that some were skipped.

daf:
[...]

PR #301, opened 2016-09-16
Adds ability to skip checks

benjwadams:
Allows skipping of check methods (methods beginning with `check_`) via a
CLI flag of `-s` or `--skip-checks`. Addresses #202.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,8 @@ optional arguments:
   -s {check1} [check2...] --skip-checks {check1} [check2...]
                         Skips the any check functions contained within the
                         check suites selected which have the same function
-                        name(s) as the string(s) specified.
+                        name(s) as the string(s) specified.  May be specified
+                        multiple times in order to skip multiple checks.
   -d DOWNLOAD_STANDARD_NAMES, --download-standard-names DOWNLOAD_STANDARD_NAMES
                         Specify a version of the cf standard name table to
                         download as packaged version
--- a/README.md
+++ b/README.md
@@ -61,6 +61,10 @@ optional arguments:
                         Output format
   -o OUTPUT, --output OUTPUT
                         Output filename
+  -s {check1} [check2...] --skip-checks {check1} [check2...]
+                        Skips the any check functions contained within the
+                        check suites selected which have the same function
+                        name(s) as the string(s) specified.
   -d DOWNLOAD_STANDARD_NAMES, --download-standard-names DOWNLOAD_STANDARD_NAMES
                         Specify a version of the cf standard name table to
                         download as packaged version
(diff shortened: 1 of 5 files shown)


compliance_checker/runner.py
```diff
@@ -16,7 +16,8 @@ class ComplianceChecker(object):
     the command line or can be used via import.
     """
     @classmethod
-    def run_checker(cls, ds_loc, checker_names, verbose, criteria, output_filename='stdout', output_format='stdout'):
+    def run_checker(cls, ds_loc, checker_names, verbose, criteria,
+                    skip_checks=[], output_filename='stdout', output_format='stdout'):
```
daf: You can't put `[]` in a method signature.

benjwadams: I'm aware of the issue when mutating an array, but this code shouldn't be modifying the default arg.  I could do a check for `None` instead if you'd like.

benjwadams:
Oops, let me update the README with the new CLI flag.

compliance_checker/runner.py
```diff
@@ -25,13 +27,16 @@ def run_checker(cls, ds_loc, checker_names, verbose, criteria, output_filename='
...
         @param  output_filename Path to the file for output
+        @param  skip_checks     Names of checks to skip
         @param  output_format   Format of the output
 
         @returns                If the tests failed (based on the criteria)
         """
         cs = CheckSuite()
         ds = cs.load_dataset(ds_loc)
-        score_groups = cs.run(ds, *checker_names)
+
```
lukecampbell: Not required but I prefer to do
```python
skip_checks = skip_checks or []
```

lukecampbell:
Please change the CLI argument parser to use an array instead of `nargs=*`.  I would prefer users to have to write

```
--skip-checks <name> --skip-checks <name> --skip-checks <name>
```

rather than have the ambiguity of argument order because of the nargs.

README.md
```diff
@@ -52,6 +52,11 @@ optional arguments:
                         Output format
   -o OUTPUT, --output OUTPUT
                         Output filename
+  -s {check1} [check2...] --skip-checks {check1} [check2...]
+                        Skips the any check functions contained within the
+                        check suites selected which have the same function
+                        name(s) as the string(s) specified.  May be specified
+                        multiple times in order to skip multiple checks.
```
lukecampbell: Could you update the README?

benjwadams: It was updated; specifically I added "May be specified multiple times in order to skip multiple checks."

benjwadams: Oh, let me fix the invocation for the command.

lukecampbell:
@daf you all set?

lukecampbell:
@daf could you take a gander over this way, I'd rather not merge when you have requested changes.

kknee:
@daf can you take another look at this pull request?

PR #304, opened 2016-09-16
Allows output of text to file

### LINKED ISSUE #290: Save text output
@lukecampbell and @daf I am helping @jbosch-noaa with some notebooks in

https://github.com/ioos/notebooks_demos/pull/21

and I noticed that simple text outputs are only redirected to `stdout`. There is on API to save it to a file.
Correct me if I am wrong but I could not see it [here](https://github.com/ioos/compliance-checker/blob/7276abba102e2ef6cb390d49ba0c38726f625913/compliance_checker/runner.py#L43-L50
).

I do know that in CLI interface it is easy to redirect to a file but it would be nice to be able to do that from class method too. Not also that the docs give the idea that this is possible and the error message [here](https://github.com/ioos/compliance-checker/blob/7276abba102e2ef6cb390d49ba0c38726f625913/compliance_checker/runner.py#L53) is not very helpful.

benjwadams:
Allows 'text' format to be output to file with the `-o` option, i.e.
`compliance-checker -t acdd -f text -o cc_results.txt ncfile.nc`
Fixes #290.

### DIFF
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -315,10 +315,6 @@ class CheckSuite(object):
 
         return score_list, points, out_of
 
-    # TODO: These methods for output generation could use a refactor to simply
-    #       generate a string and handle output as necessary instead of a
-    #       large number of print functions.  Perhaps use something like a
-    #       Jinja2 template?
     def standard_output(self, limit, check_name, groups):
         """
         Generates the Terminal Output for Standard cases
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -315,6 +315,10 @@ class CheckSuite(object):
 
         return score_list, points, out_of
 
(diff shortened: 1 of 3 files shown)


benjwadams:
Ran tests against both Python 2 and 3 so should hopefully work across newer major versions.

compliance_checker/suite.py
```diff
@@ -315,6 +315,10 @@ def get_points(self, groups, limit):
 
         return score_list, points, out_of
 
+    # TODO: These methods for output generation could use a refactor to simply
```
ocefpaf: :+1:  to jinja templates
:-1: to `TODO` comments inside the code may get lost. Can you open an issue for that suggestion?

kwilcox:
Any release planned? Could use this!

kknee:
Yep, this should be in next week's release

PR #309, opened 2016-09-20
Fix missing reference to process_doc in suite

daf:


lukecampbell:
Is this ready to merge?

daf:
It doesn't clear the entire path through for a remote XML document but it fixes obvious errors to get us to the next step.  Your call

Issue #216, opened 2016-02-04
IOOS Metadata checker needs Wicken updating


daf:
There aren't any tests for it, and netcdf files work fine for it, so it wasn't caught during the Wicken removal.  The XML paths need updating.

lukecampbell:
Didn't we can wicken?

daf:
We did, but the IOOS Metadata checker isn't fully updated for that change, specifically in non-netcdf cases.

lukecampbell:
Even though I stuck it in the milestone, I'm probably going to use the resources to REMOVE wicken completely.

lukecampbell:
XPath should suffice, I don't need this weird XML ORM thing. Plus with all the work that @cmorse, @abirger and @kwilcox have done to manage the mappings between SOS and NetCDF, we can update the IOOS Metadata Checker.

daf:
^^ all of this

cmorse:
@lukecampbell Pretty sure I'm not the person you meant to @ mention.

lukecampbell:
Yeah sorry I meant @cheryldmorse

benjwadams:
I'm not seeing wicken when I run `pip freeze` on my virtualenv.

benjwadams:
I don't see Wicken when I run `pip freeze` in my virtualenv.  I'm also not entirely clear on the issue here -- is it that some checks on the IOOS suite (`-t ioos`) are not working properly on non-NetCDF datasets as there are no XPaths?

daf:
Right - the IOOS checker won't run on GetCapabilities/DescribeSensor documents as that codepath was never updated to work without Wicken.

benjwadams:
Do you know of a last known good commit WRT the IOOS SOS check?  Tried bisecting and there are just a lot of missing fields, not outright failures.

daf:
The last "working" commit would be pre-wicken removal, c4387985a15e578141bf15298067ab67e036af7d

I can't test how it works currently because I can't get past a bunch of random issues, see fixes in #309, and now I'm on some encoding issues with the XML document stuff.  I'm testing via:

```
~/dev/compliance-checker -> python cchecker.py --test ioos http://sos.maracoos.org/stable/sos/sldmb/43493/43493_2014-03-30.nc                             [fix-xml-urls]&cc-modern
Running Compliance Checker on the dataset from: http://sos.maracoos.org/stable/sos/sldmb/43493/43493_2014-03-30.nc
Traceback (most recent call last):
[...]

PR #311, opened 2016-09-23
Fix IOOS check suite

benjwadams:
Fixes the IOOS check suite for SOS DescribeSensor and GetCapabilities formats.  Also adds simple tests to invoke the checkers and see that they run.   Fixes an outstanding encoding issues with `process_doc` function.  Implements #216.


benjwadams:
Rebased to work with #312.  Please review when ready.

Issue #319, opened 2016-09-29
ACDD check in conflict with CF-1.6


castelao:
According to CF-1.6, coordinate variables shouldn't have the attribute _FillValue, but acdd.check_vertical_extents() requires it.

lukecampbell:
Currently the ACDD check doesn't have the logic necessary for differentiating between geophysical variables and coordinate variables. In our 3.0 milestone, I aim to integrate the logic needed to identify all the types of variables, and features into the core of the project to improve the checks across the board.

castelao:
Hi @lukecampbell , that sounds great. For now, is it necessary to explicitly create a masked array using the _FillValue? Otherwise, would my patch be fine for pre-3.0? Shall I open a pull request?

lukecampbell:
If you could make a pull-request to the master branch we could review it and test it for integration.

lukecampbell:
I have fix for this shortly, it'll probably be integrated early next week.

lukecampbell:
@castelao could you post a sample dataset that you were using where you identified the issue? I think I solved it with the refactor of ACDD.

PR #320, opened 2016-09-30
BUGFIX #319, coordinate variables shouldn't have _FillValue.

castelao:
According to CF-1.6, coordinate variables shouldn't have a _FillValue
attribute. But since acdd.check_vertical_extents() uses _FillValue, it
was not possible to check for ACDD while complying with CF-1.6.


lukecampbell:
@Bobfrat could you review this? It would be nice to get it in for 2.3.

PR #323, opened 2016-09-30
Fixes json output when using multiple score groups

kwilcox:
* Changed the CheckSuite's `json_output` to a `dict_output`, making it much more useful for future output formats (when that gets revamped).
* Multiple score groups now work with the `json` output format
* Cleaned up tests (there was a duplicate test name)
* Added a test that examines `stdout` contents. This should be mimicked as needed in other output formats to test the contents of `stdout` (its not tested at all right now).

### DIFF
--- a/compliance_checker/runner.py
+++ b/compliance_checker/runner.py
@@ -3,6 +3,7 @@ from __future__ import print_function
 import traceback
 import sys
 import io
+import json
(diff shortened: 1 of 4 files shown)


daf:
@lukecampbell i'd like your signoff as well.

PR #326, opened 2016-10-12
Feature Detection for Boundary Variables CF §7

lukecampbell:


compliance_checker/cfutil.py
```diff
@@ -45,34 +45,52 @@ def is_geophysical(ds, variable):
...
 
-    if getattr(ncvar, 'axis', None):
+    # Is it a §7.1 Cell Boundaries variable
+    if variable in get_cell_boundary_variables(ds):
         return False
 
     return True
 
 
+def get_cell_boundary_variables(ds):
```
mhidas: Note that the `CFBaseCheck` class already has a method `_find_boundary_vars` which also uses a cache of this list so that the search isn't repeated every time.

compliance_checker/cfutil.py
```diff
@@ -45,34 +45,52 @@ def is_geophysical(ds, variable):
...
+    '''
+    boundary_variables = []
+    has_bounds = ds.get_variables_by_attributes(bounds=lambda x: x is not None)
+    for var in has_bounds:
+        if var.bounds in ds.variables:
+            boundary_variables.append(var.bounds)
+    return boundary_variables
+
+
 def get_geophysical_variables(ds):
```
mhidas: Maybe this could also be done just once, as part of `CFBaseCheck.setup` and then cached?

mhidas:
Just as an aside, it's a bit confusing that there's a [compliance_checker/cfutil.py](https://github.com/ioos/compliance-checker/blob/master/compliance_checker/cfutil.py) and also a [compliance_checker/cf/util.py](https://github.com/ioos/compliance-checker/blob/master/compliance_checker/cf/util.py), with quite different code in them. Could they either be merged, or one of them renamed to make a clearer distinction?

lukecampbell:
@mhidas I'm going to get rid of all the util modules but one. This is just one step towards that ultimate goal.

lukecampbell:
I have to take this large refactor for 3.0 in small steps. Some of these changes will be used right away by like @Bobfrat who's working on the NCEI 2.0 plugin.

daf:
:+1:

PR #327, opened 2016-10-13
ACDD Cleanup

lukecampbell:
- General code cleanup to improve readability
- Adds unit tests for ACDD
- Checks ACDD against NCEI 1.1 Gold Standard
- Changes mechanism for variable selection in ACDD


compliance_checker/util.py
```diff
@@ -1,11 +1,27 @@
...
+        return False, ['Datetime provided is not in a valid ISO 8602 format']
+
+
+def dateparse(date_str):
+    '''
+    Returns a datetime string parsed from an ISO-8601 input
+
+    :param str date_str: An ISO-8601 string
+    '''
+    if isinstance(date_str, basestring):
```
lukecampbell: @daf @kwilcox what's the python2/3 way to do this again?

kwilcox: Try to import `basestring` and catch the `ImportError` to assign `basestring` to `str` ( no acces to computer atm).

IMO this is the best date parser: https://github.com/ioos/pyoos/blob/master/pyoos/utils/asatime.py

I've never heard of isodate package!

lukecampbell: I never heard of isodate either. I used to use dateutil but switched to arrow because of time zone offsets. dateutil will return a timezone aware object if a timezone offset is used, otherwise it will be naive. They don't compare using vanilla datetimes but they will if you're using arrow objects.

lukecampbell:
@Bobfrat @benjwadams want to look over the code? Please check for:
- readability
- obvious mistakes

No need to run tests, as travis already has.

PR #347, opened 2016-10-18
Refactors _find_coord_vars

lukecampbell:
- This is part of an effort to consolidate return types and mistaken
  assumptions in the code about types.


lukecampbell:
@DanielJMaher @daf @benjwadams I want to institute a set of mandates across the project that need to be obeyed going forward.

The first tenant is:

- If your function needs to return a list of variables, it **must** return a list of strings of the variable names. Not variable instances.

I found that all over the codebase clients of the functions made INCORRECT assumptions about the return type. I found several instances of 

```
if name in something_that_returns_instances(ds):
```

And I've found myself very frequently asking, _"crap, what does this one return? Oh that's right it's a dictionary of names to a list of instances with a special attribute to get the name..."_

Issue #344, opened 2016-10-18
Fix climatology variable detection


lukecampbell:
CF §7.4 describes climatologies and datasets describing climatologies.

After looking at `_find_clim_vars` it will identify just about all time coordinate variables as being a climatology variable. 

In practice, a dataset describing a climatology has the following properties:

- The `time` coordinate variable has `climatology` attribute which defines the variable describing the cell boundaries for the climatology.
- Geophysical variables define `cell_methods`
- There is a variable that defines the cell boundaries for the climatology

```
dimensions:
  time=4;
  nv=2;
variables:
  float temperature(time,lat,lon);
    temperature:long_name="surface air temperature";
    temperature:cell_methods="time: minimum within years time: mean over years";
    temperature:units="K";
  double time(time);
    time:climatology="climatology_bounds";
    time:units="days since 1960-1-1";
  double climatology_bounds(time,nv);
data:  // time coordinates translated to date/time format
  time="1960-4-16", "1960-7-16", "1960-10-16", "1961-1-16" ;
  climatology_bounds="1960-3-1",  "1990-6-1",
                     "1960-6-1",  "1990-9-1",
                     "1960-9-1",  "1990-12-1",
                     "1960-12-1", "1991-3-1" ;
```

PR #348, opened 2016-10-18
Fixes climatology variable detection and adds tests

lukecampbell:
- Detecting a climatology variable is now exactly as it is described in
  §7.4 of CF.
- There can only be one variable describing the climatology bounds in a
  CF compliant dataset.
- It's simply whatever the time variable points to with the
  `climatology` attribute.
- It is NOT a geophysical variable

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -326,45 +326,14 @@ class CFBaseCheck(BaseCheck):
                              invalidated.
         '''
 
-        if ds in self._clim_vars and not refresh:
+        if self._clim_vars.get(ds, None) and not refresh:
             return self._clim_vars[ds]
 
-        c_time = set()  # set of climatological time axes
-        c_vars = set()  # set of climatological variables
+        climatology_variable = cfutil.get_climatology_variable(ds)
+        if climatology_variable:
+            self._clim_vars[ds].append(climatology_variable)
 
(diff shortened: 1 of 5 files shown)


lukecampbell:
Fixes https://github.com/ioos/compliance-checker/issues/344

compliance_checker/cf/cf.py
```diff
@@ -688,73 +657,73 @@ def check_units(self, ds):
...
-            if v in clim_vars or \
-               v in boundary_vars or \
-               v in metadata_vars or \
-               v in ancillary_vars or \
-               k in container_vars:
+            if name in clim_vars or \
+               name in boundary_vars or \
+               name in metadata_vars or \
+               name in ancillary_vars or \
+               name in container_vars:
```
daf: have all of these methods been updated for the name/variable consistency thing?  boundary vars does not look it to me: https://github.com/lukecampbell/compliance-checker/blob/bf45f8bda048c680f0ac4c7ecaa644c529db6d8f/compliance_checker/cf/cf.py#L338-L358

lukecampbell: It's in progress. The tests still pass, so what's that say about our tests?

lukecampbell:
I need to update the rest of them like @daf said.

compliance_checker/cf/cf.py
daf: :laughing:

PR #350, opened 2016-10-18
Fixes variable detection method return types

lukecampbell:


compliance_checker/cf/cf.py
```diff
@@ -326,45 +327,14 @@ def _find_clim_vars(self, ds, refresh=False):
                              invalidated.
         '''
 
-        if ds in self._clim_vars and not refresh:
+        if self._clim_vars.get(ds, None) and not refresh:
```
daf: `not refresh` should be changed for consistency

PR #353, opened 2016-10-19
Refactor check_units

lukecampbell:
- Replaces `_find_data_vars` with `_find_geophysical_variables`
- Refactors `check_units` to correctly validate units according to section 3 of CF 1.6
- Splits the complicated logic of `check_units` into three functions
- Adds more detailed unit tests to check assertions of failure and success

### DIFF
--- a/compliance_checker/base.py
+++ b/compliance_checker/base.py
@@ -91,7 +91,15 @@ class Result(object):
     Stores the checker instance and the check method that produced this result.
     """
 
-    def __init__(self, weight=BaseCheck.MEDIUM, value=None, name=None, msgs=None, children=None, checker=None, check_method=None):
+    def __init__(self, 
+                 weight=BaseCheck.MEDIUM,
+                 value=None,
+                 name=None,
+                 msgs=None,
+                 children=None,
+                 checker=None,
+                 check_method=None,
+                 variable_name=None):
 
         self.weight = weight
 
@@ -109,6 +117,7 @@ class Result(object):
 
         self.checker = checker
         self.check_method = check_method
+        self.variable_name = variable_name
 
     def __repr__(self):
         ret = '{} (*{}): {}'.format(self.name, self.weight, self.value)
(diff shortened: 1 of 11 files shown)


lukecampbell:
@daf @benjwadams 

This is the start of the big changes.

I started the refactor by simply validating that check_units actually flagged instances where units were wrong, or the UDUnits were not valid or not appropriate for a standard name. And unfortunately it behaved somewhat erratically. 

I tried to clean it up but I couldn't understand what key parts were doing.

For starters, the units are converted to string:

https://github.com/ioos/compliance-checker/blob/b7be4a04f8d1891f037fc2714df82e1bfad565e8/compliance_checker/cf/cf.py#L680

Which means that this will ALWAYS be true
https://github.com/ioos/compliance-checker/blob/b7be4a04f8d1891f037fc2714df82e1bfad565e8/compliance_checker/cf/cf.py#L683

and so will this:
https://github.com/ioos/compliance-checker/blob/b7be4a04f8d1891f037fc2714df82e1bfad565e8/compliance_checker/cf/cf.py#L683

Then there was the filtering of all sorts of variables:
https://github.com/ioos/compliance-checker/blob/b7be4a04f8d1891f037fc2714df82e1bfad565e8/compliance_checker/cf/cf.py#L658-L662

I replaced that with the logic that units should be checked on:
- geophysical variables
- coordinate variables
- auxiliary coordinate variables

Nothing else fits the bill.

I also added the logic for the [standard name modifier units](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#standard-name-modifiers) which it turns out only `number_of_observations` changes the units we need to check. The others aren't checked at all or share the same units as the `standard_name`.

daf:
How does a `TestCtx` differ from a `Result`?

lukecampbell:
It's just a bit easier to use. The `assert_true` function basically does what we end up doing everywhere: apply this message if condition is not met.

daf:
Is there a way to merge the two into a single unit of functionality?

lukecampbell:
I may rename it or make it a little more correct after the refactor, for now I just moved it from NCEI to here.

PR #356, opened 2016-10-20
Refactors chapter 4 coordinate types checking

lukecampbell:
- CF allows for coordinate types to be things besides time,lat,lon etc.
  (See §4.5 Discrete Axis for more info)
- Most of Chapter 4's intro is recommended and optional
- The only required aspect is that axis be T, X, Y, or Z
- Coordinate types that are dimensionless are LOOSELY considered
  coordinate variables that have a dimension size of 1.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -396,7 +396,7 @@ class CFBaseCheck(BaseCheck):
                                'float32'
                                ]:
 
-                fails.append(('The variable %s failed because the datatype is %s' % (k, v.datatype)))
+                fails.append('The variable {} failed because the datatype is {}'.format(k, v.datatype))
         return Result(BaseCheck.HIGH, (total - len(fails), total), '§2.2 Valid netCDF data types', msgs=fails)
 
     def check_naming_conventions(self, ds):
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -412,13 +412,14 @@ class CFBaseCheck(BaseCheck):
         fails = []
         total = len(ds.variables)
 
-        rname = re.compile("[A-Za-z][A-Za-z0-9_]*")
+        rname = re.compile("^[A-Za-z][A-Za-z0-9_]*$")
 
         for k, v in ds.variables.items():
             if not rname.match(k):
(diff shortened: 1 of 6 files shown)


compliance_checker/cf/cf.py
```diff
@@ -412,13 +412,14 @@ def check_naming_conventions(self, ds):
...
+        rname = re.compile("^[A-Za-z][A-Za-z0-9_]*$")
 
         for k, v in ds.variables.items():
             if not rname.match(k):
-                fails.append('Variable %s failed because it does not start with a letter, digit, or an underscore' % k)
+                fails.append('Variable {} should begin with a letter and be composed of letters, '
+                             'digits and underscores'.format(k))
 
-        return Result(BaseCheck.HIGH, (total - len(fails), total), '§2.3 Legal variable names', fails)
+        return Result(BaseCheck.MEDIUM, (total - len(fails), total), '§2.3 Legal variable names', fails)
```
lukecampbell: I wish CF said this was a **must** but it actually says

> variable, dimension and attribute names _should_ begin with a letter and be composed of letters, digits, and underscores.

compliance_checker/cf/cf.py
```diff
@@ -409,16 +409,42 @@ def check_naming_conventions(self, ds):
...
         '''
-        fails = []
-        total = len(ds.variables)
+        ret_val = []
+        variable_naming = TestCtx(BaseCheck.MEDIUM, '§2.3 Naming Conventions for variables')
+        dimension_naming = TestCtx(BaseCheck.MEDIUM, '§2.3 Naming Conventions for dimensions')
+        attribute_naming = TestCtx(BaseCheck.MEDIUM, '§2.3 Naming Conventions for attributes')
 
-        rname = re.compile("[A-Za-z][A-Za-z0-9_]*")
+        rname = re.compile("^[A-Za-z][A-Za-z0-9_]*$")
```
lukecampbell: Without the `^` and `$` it won't capture spaces and whitespace as illegal characters.

benjwadams: https://docs.python.org/2/library/re.html#re.match

Match is implicitly anchored at the start of the string, although not the end.

PR #359, opened 2016-10-20
Fixes dimension checks

lukecampbell:
- Adds unit test to verify duplicate dims check
- Fixes dimension order check to deal with dimensions defined without coordinate variables


lukecampbell:
Force pushed to keep the 8MiB netCDF file out of the commit history.

Issue #371, opened 2016-10-24
Refactor check_calendar


lukecampbell:
- iterate over time coordinates not every variable.
- Include variable name in result title.
- `calendar` attribute is optional, so only check it's contents if it exists. If it doesn't exist, there is no check. CF assumes that if `calendar` is omitted, then it is `standard/gregorian` according to §4.4.1

PR #387, opened 2016-10-26
Fixes calendar check

benjwadams:
Addresses #371

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -1862,10 +1862,14 @@ class CFBaseCheck(BaseCheck):
             if not util.is_time_variable(k, v):
                 continue
             reasoning = None
+            has_calendar = hasattr(v, 'calendar')
+            result = Result(BaseCheck.LOW,
+                            has_calendar,
+                            '§4.4.1 Time and calendar',
+                            reasoning)
             # if has a calendar, check that it is within the valid values
             # otherwise no calendar is valid
-
-            if hasattr(v, 'calendar'):
+            if has_calendar:
                 reasoning = None
                 valid_calendar = v.calendar in valid_calendars
                 if not valid_calendar:
@@ -1874,7 +1878,7 @@ class CFBaseCheck(BaseCheck):
                                     valid_calendar,
                                     '§4.4.1 Time and calendar',
                                     reasoning)
-                    ret_val.append(result)
+                ret_val.append(result)
 
         return ret_val
 
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -1861,16 +1861,10 @@ class CFBaseCheck(BaseCheck):
             if not util.is_time_variable(k, v):
                 continue
             reasoning = None
-            has_calendar = hasattr(v, 'calendar')
-            if not has_calendar:
-                reasoning = ['Variable %s should have a calendar attribute' % k]
-            result = Result(BaseCheck.LOW,
-                            has_calendar,
-                            '§4.4.1 Time and calendar',
-                            reasoning)
-            ret_val.append(result)
+            # if has a calendar, check that it is within the valid values
+            # otherwise no calendar is valid
 
-            if has_calendar:
+            if hasattr(v, 'calendar'):
                 reasoning = None
                 valid_calendar = v.calendar in valid_calendars
                 if not valid_calendar:
@@ -1879,7 +1873,7 @@ class CFBaseCheck(BaseCheck):
                                     valid_calendar,
                                     '§4.4.1 Time and calendar',
                                     reasoning)
-                ret_val.append(result)
+                    ret_val.append(result)
 
         return ret_val
 
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -1780,7 +1780,8 @@ class CFBaseCheck(BaseCheck):
             if not has_units:
                 result = Result(BaseCheck.HIGH,
                                 False,
-                                '§4.4 Time coordinate variable and attributes', ['%s does not have units' % k])
+                                '§4.4 Time coordinate variable and attributes',
+                                ['%s does not have units' % k])
                 ret_val.append(result)
                 continue
             # Correct and identifiable units
@@ -1791,7 +1792,7 @@ class CFBaseCheck(BaseCheck):
             correct_units = util.units_temporal(v.units)
             reasoning = None
             if not correct_units:
-                reasoning = ['%s doesn not have correct time units' % k]
+                reasoning = ['%s does not have correct time units' % k]
             result = Result(BaseCheck.HIGH,
                             correct_units,
                             '§4.4 Time coordinate variable and attributes',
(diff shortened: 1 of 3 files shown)


compliance_checker/tests/data/non-comp/bad.cdl
```diff
@@ -45,10 +45,7 @@ variables:
 		lev2:standard_name = "atmosphere_hybrid_height_coordinate" ;
 		lev2:formula_terms = "a: var1 b: var2 orog: var3" ;
 		lev2:description = "Missing variables from formula_terms" ;
-	double bad_time_1(time) ;
-		bad_time_1:standard_name = "time" ;
-		bad_time_1:description = "lacks units" ;
-	double bad_time_2(time) ;
+	double time(time) ;
 		bad_time_2:units = "seconds" ;
```
daf: i think you need to rename these prefixes as well.  also this likely breaks the test changed below (the assert will not have the test `bad_time_2` in it anymore)

compliance_checker/cf/cf.py
```diff
@@ -1910,7 +1904,7 @@ def check_calendar(self, ds):
                                     valid_calendar,
                                     '§4.4.1 Time and calendar',
                                     reasoning)
-                ret_val.append(result)
+                    ret_val.append(result)
```
daf: While this does fix the code so it only appends an existing result, I think both the ret_val line and the result line should be un-indented, so there's always a Result object returned, not just when the calendar is invalid.

I know this goes a bit against the (possible) future philosophy of only report issues, but this would keep the check in line with the current philosophy. 

@lukecampbell please weigh in.

lukecampbell: According to CF, `calendar` is actually optional. So the only checks that it recommends is that if it exists, that it be a valid value according to the list in 4.4.1.

I'm ok with returning results that pass if the attribute doesn't exist.

daf: At this point in logic, we've returned already if the attribute doesn't exist.  I agree there's no Result for that, because we didn't check anything.  Specificaly, at this code juncture, it only is returning a Result if `not v.calendar in valid_calendars`, and that's incorrect.  We've found the calendar attribute, tell me if it's valid or not.

compliance_checker/tests/data/non-comp/bad.cdl
benjwadams: Ooops, thought I had changed over the prefixes.  I need to change the variable name and/or the dimension name, as the `bad_time_1(time)` is not a coordinate variable as the var name isn't the same as the coordinate name.

compliance_checker/cf/cf.py
lukecampbell: In other words, please remove 4 spaces.

lukecampbell:
Can you rebase to fix the conflicts?

benjwadams:
Rebased against master and force pushed to resolve merge conflicts.

lukecampbell:
@daf could you review the latest?

compliance_checker/cf/cf.py
```diff
@@ -1893,14 +1894,12 @@ def check_calendar(self, ds):
...
-            if not has_calendar:
-                reasoning = ['Variable %s should have a calendar attribute' % k]
             result = Result(BaseCheck.LOW,
                             has_calendar,
                             '§4.4.1 Time and calendar',
                             reasoning)
-            ret_val.append(result)
-
+            # if has a calendar, check that it is within the valid values
+            # otherwise no calendar is valid
```
daf: this can be interpreted two ways (thanks, english) but not pressing enough to correct this PR

compliance_checker/cf/cf.py
```diff
@@ -1893,14 +1894,12 @@ def check_calendar(self, ds):
                 continue
```
lukecampbell: @benjwadams Could you change the for loop to loop over `cfutil.get_time_variables(ds)` instead?

lukecampbell: Actually the easiest way to fix everything is to change the for loop to iterate over

```python
for timevar in ds.get_variables_by_attributes(calendar=lambda x : x is not None):
```

compliance_checker/cf/cf.py
```diff
@@ -1893,14 +1894,12 @@ def check_calendar(self, ds):
...
             reasoning = None
             has_calendar = hasattr(v, 'calendar')
-            if not has_calendar:
-                reasoning = ['Variable %s should have a calendar attribute' % k]
             result = Result(BaseCheck.LOW,
                             has_calendar,
                             '§4.4.1 Time and calendar',
                             reasoning)
```
lukecampbell: We don't need a check that checks if something has a `calendar` attribute, only that the value of `calendar` is in a list of acceptable values.

compliance_checker/cf/cf.py
```diff
@@ -1893,14 +1894,12 @@ def check_calendar(self, ds):
...
-            ret_val.append(result)
-
+            # if has a calendar, check that it is within the valid values
+            # otherwise no calendar is valid
             if has_calendar:
```
lukecampbell: If you change the for loop to only loop over variables with a calendar-attribute this condition is not needed and you can just check if the calendar is valid.

compliance_checker/cf/cf.py
benjwadams: This was removed, it looks like you're looking at the old changes.

compliance_checker/cf/cf.py
lukecampbell: nope. There's no requirement for a CF compliant dataset to only have one time coordinate.

daf:
Looking good (conflicts need to be addressed, assuming the util PR just goofed it).

A future improvement could be to list the calendars as CC understands them - they are hardcoded with no explicit reference as to where those values come from.  If the CF standard changes or the way those values are formatted differ slightly, it could lead to frustration.  The most clear way CC can communicate this issue is:

> Variable "time" does not have a valid calendar ("made_up_calendar").  Valid calendars are "gregorian", "standard", "proleptic_gregorian", ...  See http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#calendar

The challenge with that is the length of the list of valid calendars and having reasonable output.  Hopefully the link locks the values with the CF version CC is coded to, etc etc.

benjwadams:
Re force pushing again to fix merge conflicts.

lukecampbell:
I don't know how this is possible but f17e07aff66bfd373997101e9d3fd33a21c2f892 broke the build for the project and travis didn't catch it.

lukecampbell:
```
tests/test_cf.py .....F............................F.........F...
AssertionError: assert "Variable time should have a valid calendar: 'nope' is not a valid calendar" in ["Variable bad_time_2 should have a valid calendar: 'nope' is not a valid calendar"]
E       AssertionError: assert 'time does not have correct time units' in ['bad_time_1 does not have units', 'bad_time_2 does not have correct time units']
E       AssertionError: assert (14, 15) == (13, 14)
E         At index 0 diff: 14 != 13
E         Use -v to get the full diff
```

lukecampbell:
no, I take it back, the netCDF files were changed. which means I need to clear the cache of netCDF files

lukecampbell:
Yup, that was it, cached netCDF files.  We need to figure out a way to clear netCDF files before tests run.

Issue #375, opened 2016-10-24
Add check for grids containing coordinates


lukecampbell:
CF Chapter 5 Paragraph 8

> If the coordinate variables for a horizontal grid are not longitude and latitude, it is recommended that they be supplied in addition to the required coordinates. For example, the Cartesian coordinates of a map projection should be supplied as coordinate variables in addition to the required two-dimensional latitude and longitude variables that are identified via the coordinates attribute. The use of the axis attribute with values X and Y is recommended for the coordinate variables

Basically, if it's a grid, check that latitude/longitude exist and are associated via `coordinates`

PR #389, opened 2016-10-27
Refactors the coordinate checks for grids

lukecampbell:
- Adds feature detection for grids that don't vary with time.
- Refactors the grid checks for coordinates to use more consistent checking
  - Ensure that for every grid feature defined, it is associated with true lat and true lon


lukecampbell:
This fixes #375

Issue #364, opened 2016-10-24
Fix imports


lukecampbell:
The imports are too long, they should be broken up.

benjwadams:
I'm assuming you mean the from `x import y, ... z100` imports here?

lukecampbell:
I would actually prefer to import the module and use `.` to access the part of the module accessed.

```python
from compliance_checker.cf import util as legacy_util
...

legacy_util.units_known(...)
```

PR #391, opened 2016-10-27
Refactor util imports in cf.py

benjwadams:
Refactors util imports to use namespaced imports.  Implements #364


daf:
:+1: :+1: :+1:

PR #399, opened 2016-10-28
Adds check for instance variables being refd as coords

### LINKED ISSUE #396: Coords dont define cf_role
Update an existing check or add a new one to make sure that any referenced coordinates, either coordinate variables, or auxiliary coordinate variables directly associated with a variable don't define `cf_role`.

lukecampbell:
Fixes #396


lukecampbell:
Fixes #396

PR #405, opened 2016-10-28
Refactors feature type check

lukecampbell:


lukecampbell:
@benjwadams @daf @ocefpaf 

The travis build engine is using an old (not sure how old, but old) version of netCDF. So, the ncgen command that it uses can't correctly convert [the CDL](https://github.com/lukecampbell/compliance-checker/blob/57274b54655f88a40aedea88a6d0cb9224bc520e/compliance_checker/tests/data/bad_data_type.cdl) to use int64 types. 

In the interest of time and sanity, I'm going to commit the file but it would be super awesome if the netCDF version could get bumped, I'm not even sure what controls that, I suspect conda.

benjwadams:
We're using conda-forge for Conda, and it looks like there's the latest version of libnetcdf (4.4.1 as of the date of this writing).  So maybe it's something else?

PR #407, opened 2016-10-31
Fixes assumption about coordinates

lukecampbell:
- Updates label in test

My assumption up until this point was that coordinate types where
physical data representing a coordinate of the variable. This is
supported in every chapter right up until chapter 6. §6.1, Labels,
throws a huge wrench into this assumption.

Chapter 6 allows for auxiliary coordinates to be labels, like regions or
station names or whatever. It was my assumption, until I reread it, that
labels just shared dimensions with features, not were directly related
through coordinates.

I personally disagree with this, labels aren't conceptually coordinates
they would be better suited as ancillary data. They don't define where,
or when in the physical world or a mapped coordinate system.

In any case this refactor corrects my incorrect assumption.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -779,7 +779,7 @@ class CFBaseCheck(BaseCheck):
         standard_name, standard_name_modifier = self._split_standard_name(standard_name)
         unitless_standard_names = cfutil.get_unitless_standard_names()
 
-        should_be_unitless = standard_name in unitless_standard_names
+        should_be_unitless = standard_name in unitless_standard_names or variable.ndim == 0 or variable.dtype.char == 'S'
 
         # 1) Units must exist
         valid_units = TestCtx(BaseCheck.HIGH, '§3.1 Variable {} contains valid CF units'.format(variable_name))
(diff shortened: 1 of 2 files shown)


lukecampbell:
@daf @benjwadams @mbiddle-nodc 

This was a big eye opener for me and it invalidated several very important assumptions, so I want to make sure I share this with you guys and we're all aware.

Coordinates are generally physical (although sometimes dimensionless) quantities that tell the "when" and "where" about feature types. The huge use case of CF is to help software authors make accurate reasoning about arbitrary netCDF datasets, and the likely biggest use case is finding features and finding out where they are in the physical world. 

Unfortunately, Chapter 6 throws this out the window with labels. Which instead of being ancillary variables, which I think they should be, they allow for labels to be associated with variables through the `coordinates` attribute. 

This is the fun bit, Chapter 3 and 4 both outline recommendations and requirements for variables and coordinates. Specifically things like units being required, either `standard_name` or `long_name` to be used. The usual exceptions to those rules outlined in Chapter 3 and 4 are `dimensionless quantities` which labels are not dimensionless... 

So this is an example of not necessarily contradiction but certainly a lack of clarification in CF 1.6. 

For now the CF checker will ignore coordinates that are of type `string` as described in §6.1 and coordinates that define the `cf_role` attribute. §9.5 has several examples of `cf_role` and none of them are listed as coordinates. 

If anyone is on a CF review board, could someone clarify the reasoning behind using labels as `coordinates` instead of `ancillary_variables` ?

lukecampbell:
@rsignell-usgs Including Rich as well, he might have some insight.

PR #408, opened 2016-10-31
Start to add the Appendix H examples

lukecampbell:
This also fixes the parsing of dimension order to allow for unlimited dimensions to precede other dimensions according to §9.3 p2

And fixes bug with duplicate vertical axes showing up in axis check.


lukecampbell:
I will be adding unit tests as part of the Chapter 9 stuff for all these examples.

PR #411, opened 2016-11-01
Fixes chapter 4 assumptions

lukecampbell:
- Chapter 4 mainly applies to gridded features.
- There is no requirement for non-spatial coordinates to have
  units/axis/standard_name
- There is no requirement for an axis to have an explicit relationship
  to a spatial coordinate in chapter 4.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -1245,9 +1245,9 @@ class CFBaseCheck(BaseCheck):
         :return: List of results
         '''
         ret_val = []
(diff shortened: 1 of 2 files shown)


lukecampbell:
Basically, I refactored the §4.0 checks to be just checking if axis is one of T, Z, Y, or X

I've reread chapter 4 several times by now. There is no mention of any requirement for all of the stuff we were checking. There are several requirements for coordinates of time, latitude, longitude and vertical coordinates though. These are individually checked in separate checker functions.

Issue #394, opened 2016-10-28
Add 8.1 checks on compression


lukecampbell:
- `compress` attribute must be a blank-separated string
- compression coordinates must be coordinate variables
- compression coordinates must be an integer type.

benjwadams:
From the bullet points here, I think you mean 8.2?

benjwadams:
OK to close?

PR #420, opened 2016-11-04
Implements 8.2, "Compression by gathering"

benjwadams:
Implements section 8.2 of the CF manual, "Compression by
gathering"/issue #394 on GitHub.

Checks that compression variable containing compress attribute has
exactly one dimension, the compression variable is of integer type, and
the referenced dimensions from the blank-separated 'compress' attribute
exist within the NetCDF dataset.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -2473,16 +2473,16 @@ class CFBaseCheck(BaseCheck):
                                "snow",
                                "trees"
                                "vegetation"]
-        methods = [ 'point',
-                    'sum',
-                    'mean',
-                    'maximum',
-                    'minimum',
-                    'mid_range',
-                    'standard_deviation',
-                    'variance',
-                    'mode',
-                    'median']
+        methods = ['point',
+                   'sum',
+                   'mean',
+                   'maximum',
+                   'minimum',
+                   'mid_range',
+                   'standard_deviation',
+                   'variance',
+                   'mode',
+                   'median']
 
         ret_val = []
         reasoning = []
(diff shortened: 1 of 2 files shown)


compliance_checker/cf/cf.py
```diff
@@ -2809,6 +2809,70 @@ def check_packed_data(self, ds):
...
+        offset.
+        """
+        ret_val = []
+        for compress_var in ds.get_variables_by_attributes(compress=lambda s: s is not None):
+            valid = True
+            reasoning = []
+            # puts the referenced variable being compressed into a set
+            compress_set = set(compress_var.compress.split(' '))
+            if compress_var.ndim != 1:
+                valid=False
```
lukecampbell: Could you change this line to read `valid = False` ?

compliance_checker/cf/cf.py
```diff
@@ -2809,6 +2809,70 @@ def check_packed_data(self, ds):
...
+                reasoning.append("Compression variable {} may only have one dimension".format(compress_var.name))
+            # ensure compression variable is a proper index, and thus is an
+            # signed or unsigned integer type of some sort
+            if compress_var.dtype.kind not in {'i', 'u'}:
+                valid=False
```
lukecampbell: Same here. By the way vim has a super awesome Python linter

compliance_checker/cf/cf.py
benjwadams: ups

benjwadams: I do have a linter, but somehow missed this.  Also, a side note: a number of lines exceed 80 columns and I'm not sure how to easily break some of them up.

PR #422, opened 2016-11-07
Fix handling of numpy arrays when no mask defined

benjwadams:
Fixes missing data method for arrays where there are no masked values, i.e.
 NetCDF variables with no _FillValue attribute defined.


benjwadams:
Please don't merge for the time being, this accesses data and I'd like to find a more elegant way to go about this if possible.

Issue #325, opened 2016-10-11
CF: "boundary variables" don't need coordinates


mhidas:
I have received a file that has:

```
dimensions:
        TIME = 21 ;
        nv = 2 ;
variables:
        double TIME(TIME) ;
                ...
                TIME:bounds = "TIME_bnds" ;
        double TIME_bnds(TIME, nv) ;
```

According to [CF section 7.1](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#cell-boundaries) this is correct, and the `TIME_bnds` variable doesn't need to have "attributes such as `long_name` and `units`". Presumably that includes the `coordinates` attribute too.
However, the CF checker complains:

```
§9.5 Discrete Geometry                 :2:    10/11 :  
    TIME_bnds                          :2:     0/ 1 :  
        check_coordinates              :2:     0/ 1 : The variable TIME_bnds
                                                      does not have associated
                                                      coordinates
```

lukecampbell:
The CF Checker doesn't deal with gridded data very well just yet. This should be taken care of in the 3.0 release.

lukecampbell:
@benjwadams this should be fixes as part of #349

benjwadams:
Did we merge something to fix this?  With all the PR activity I've sort of lost track.

lukecampbell:
@mhidas do you have a sample dataset?

benjwadams:
This wasn't fixed as part of #349 as that dealt with implementation of the `bounds` attribute.   I will make a separate branch to fix this issue.

benjwadams:
@mhidas, I tried reproducing this by taking the `tests/data/grid-boundaries` CDL and removing units from the bounds variables, but was unable to reproduce the behavior you mentioned against the current master branch.  If you have another dataset I could test against it'd be helpful as well.

mhidas:
@benjwadams I'm a bit confused. I just re-ran the CF checker on the file in question and now it passes all checks! Even when I go back to version 2.2.0, which is what we were using at the time.

However, I did manage to construct a simple example that shows the original issue:
```
netcdf test {
dimensions:
    time = 2;
    nv = 2;
variables:
    double time(time);
        time:standard_name = "time";
        time:axis = "T";
        time:units = "seconds since 1970-01-01T00:00:00Z";
        time:calendar = "gregorian";
        time:bounds = "time_bnds";
    double time_bnds(time, nv);
    float lat;
[...]

PR #425, opened 2016-11-07
Skips coordinate order checks for climatology and cell bounds variables

benjwadams:
Skips dimension order checks for climatology and cell bounds variables.  Still
checks the parent variables, but it is unnecessary to check the former
variables as they already have their own set of checks which make the
generalized dimensional checks somewhat redundant.  Additionally, the
generalized checks were leading to erroneous bad results.  Should
address #325.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -505,8 +505,16 @@ class CFBaseCheck(BaseCheck):
             if dimension not in coord_axis_map:
                 coord_axis_map[dimension] = 'U'
 
-        # Check each variable's dimension order
+        # Check each variable's dimension order, excluding climatology and
+        # bounds variables
+        any_clim = cfutil.get_climatology_variable(ds)
+        any_bounds = cfutil.get_cell_boundary_variables(ds)
         for name, variable in ds.variables.items():
+            # Skip bounds/climatology variables, as they should implicitly
+            # have the same order except for the bounds specific dimension.
+            # This is tested later in the respective checks
+            if name in any_bounds or name == any_clim:
+                continue
             if variable.dimensions:
                 dimension_order = self._get_dimension_order(ds, name, coord_axis_map)
                 valid_dimension_order.assert_true(self._dims_in_order(dimension_order),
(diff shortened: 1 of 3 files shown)


compliance_checker/cf/cf.py
```diff
@@ -505,8 +505,16 @@ def check_dimension_order(self, ds):
             if dimension not in coord_axis_map:
                 coord_axis_map[dimension] = 'U'
 
-        # Check each variable's dimension order
+        # Check each variable's dimension order, excluding climatology and
+        # bounds variables
+        any_clim = cfutil.get_climatology_variable(ds)
+        any_bounds = cfutil.get_cell_boundary_variables(ds)
         for name, variable in ds.variables.items():
```
lukecampbell: Could you change this line to be:
```python
for name in (any_clim + any_bounds):
    variable = nc.variables[name]
```

compliance_checker/cf/cf.py
```diff
@@ -505,8 +505,16 @@ def check_dimension_order(self, ds):
...
+            # Skip bounds/climatology variables, as they should implicitly
+            # have the same order except for the bounds specific dimension.
+            # This is tested later in the respective checks
+            if name in any_bounds or name == any_clim:
+                continue
```
lukecampbell: And remove this?

compliance_checker/tests/test_cf.py
```diff
@@ -177,10 +177,8 @@ def test_check_dimension_order(self):
         """
         dataset = self.load_dataset(STATIC_FILES['bad_data_type'])
         result = self.cf.check_dimension_order(dataset)
-        assert result.value == (5, 7)
-        assert result.msgs[0] == ("salinity's dimensions are not in the recommended order "
-                                  "T, Z, Y, X. They are latitude, longitude, time")
-        assert result.msgs[1] == ("really_bad's dimensions are not in the recommended order "
+        assert result.value == (5, 6)
```
benjwadams: These lines had to be changed due to the new bounds order checking.  Checks against any bounds variables are deferred until the `check_cell_boundaries` test.  It would likely also fail on account of the fact it has a one-dimensional bounds variable with only one instead of two elements.

Issue #418, opened 2016-11-04
Refactor Chapter 7


lukecampbell:

benjwadams:
Reopening for `check_cell_methods`.

PR #432, opened 2016-11-08
Refactors cell_measure checks

benjwadams:
Refactors cell_measure checks (§7.2).  Part of #418.


benjwadams:
Travis, y u no work?

benjwadams:
Looks like the Travis build broke due to some kind of race condition while fetching packages.  Is there a way to kick it off over again without updating the PR code?

PR #458, opened 2017-01-12
Enable ESGF Authentication scheme for compliance checker

### LINKED ISSUE #456: accessing OpenDAP on ESGF data archives
ESGF data archives provide OpenDAP to access climate data. But to access the data you need a certificate and an appropriate `.dodsrc` file. The python netcdf4 library is already able to handle this without modifications. See my notebook example:

https://github.com/cehbrecht/demo-notebooks/blob/master/esgf-opendap.ipynb

When i run the compliance checker on an ESGF OpenDAP url i get the following exception:

``
Could not understand response code 401 and content-type text/html;charset=ISO-8859-1
``

To make this work you probably just need to add the 401 error code to the following line (in method `CheckSuite.load_dataset`):

https://github.com/ioos/compliance-checker/blob/master/compliance_checker/suite.py#L523

lukecampbell:
Fixes #456

### DIFF
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -520,7 +520,7 @@ class CheckSuite(object):
         if pr.netloc:       # looks like a remote url
             rhead = requests.head(ds_str, allow_redirects=True)
             # if we get a 400 here, it's likely a Dataset openable OpenDAP url
-            if rhead.status_code == 400:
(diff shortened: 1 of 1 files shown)


lukecampbell:
I ran tox on it, and everything looks good. I don't have any tests that verify the full ESGF Authentication workflow though.

Issue #457, opened 2017-01-11
cmd_data_type values


jpotemra:
Hi all,

I was wondering if you can expand the acceptable list of cmd_data_type to include "Timeseries"?  The checker seems to accept the following: ([u'Grid', u'Image', u'Point', u'Radial', u'Station', u'Swath', u'Trajectory']).  However, ERDDAP throws an error with cmd_data_type = Station.  Instead it wants "TimeSeries".  The valid cmd_data_type values for ERDDAP are: Grid, MovingGrid, Other, Point, Profile, RadialSweep, TimeSeries, TimeSeriesProfile, Swath, Trajectory, TrajectoryProfile.

BTW, ncSOS seems to be happy with either Station or TimeSeries.

Thanks,

Jim

lukecampbell:
Unfortunately the only allowed values are those that we check:
https://github.com/Unidata/thredds/issues/72

As per the authors and the THREDDS source code. I wish it was a little bit more clear in documentation for ACDD and CF where these are defined but the values don't include the values you mention.

lukecampbell:
I would like to reach out to @kwilcox @daf @brianmckenna @mbiddle-nodc for their input on this. Should we change the ACDD and CF checkers that check the attribute `cdm_data_type` to include additional enumerated values that are leveraged by other community driven software like ERDDAP?

lukecampbell:
I'm at the Winter 2017 ESIP Conference now, I'll ask about this at the netCDF session.

lukecampbell:
I think I'll remove this from the ACDD checks.

lukecampbell:
Just an update, I need to do a release of compliance checker including this patch sometime this or next week.

lukecampbell:
Just a continuation, I didn't forget about this, but I wanted to batch all these changes together with some of the recent fixes and the Glider DAC changes.

daf:
This was [released in 3.0.1](https://github.com/ioos/compliance-checker/releases/tag/3.0.1)

PR #459, opened 2017-01-12
remove cdm_data_type from ACDD checks

lukecampbell:


daf:
See #457 (discussion)

lukecampbell:
I spoke with some people at ESIP about this, and I got the same answer as in the original unidata thread https://github.com/Unidata/thredds/issues/72

`cdm_data_type` is deprecated and was slated to be removed but got thrown in 1.3. It also doesn't affect discoverability. Even though ACDD 1.1 says it maps to `gmd:MD_SpatialRepresentationTypeCode` it actually doesn't in the latest version of ncISO:

https://github.com/Unidata/threddsIso/blob/a01a0204be97e906dd5ea4384dd7f9580e19482d/src/main/resources/xsl/nciso/UnidataDD2MI.xsl#L621-L654

Issue #451, opened 2016-12-02
Test errors using conda


kwilcox:
When trying to run the tests using a `conda` env there are 4 error that pass when using `pip`:
https://gist.github.com/kwilcox/1a42e95c178af529b9f468fac2050ee5

lukecampbell:
Is this linux python 3.5?

kwilcox:
Yup!

lukecampbell:
@ocefpaf have you seen any issues with the builds for linux 3.5 ? I remember when we were updating the release there were a few broken builds.

ocefpaf:
Issue with the build? No. Let me check @kwilcox gist to see if that is a packaging issue. Hold on...

lukecampbell:
@ocefpaf It's been a while since I followed up on this, are there any updates?

ocefpaf:
We do run the test suite in [Travis-CI with conda and Python 3.6](https://github.com/ioos/compliance-checker/blob/master/.travis.yml#L8) and everything is green. @kwilcox are you still seeing those errors?

kwilcox:
I still get errors. The tests are using `pip` to install the dependencies and I'm obviously using `conda`. That is most likely the difference.

ocefpaf:
I did not see the pip install below :-(
@lukecampbell I would start be using only conda in Travis-CI to debug this.

lukecampbell:
Ok, it's probably going to be a while before I can get to this, but I'll try to debug it when I can.

ocefpaf:
See #479

PR #479, opened 2017-04-10
Investigate Travis-CI failures

ocefpaf:
The `master` branch is failing at the moment due to the pinning. 

```
Traceback:
compliance_checker/tests/test_cf_integration.py:4: in <module>
    from compliance_checker.suite import CheckSuite
compliance_checker/suite.py:15: in <module>
    from lxml import etree as ET
E   ImportError: libicui18n.so.58: cannot open shared object file: No such file or directory
```

Full log: https://travis-ci.org/ioos/compliance-checker/jobs/197774152#L514

Hopefully this will fix https://github.com/ioos/compliance-checker/issues/451 too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,28 +2,41 @@ language: python
 
 sudo: false
 
-env:
-    - CONDA="python=2.7"
-    - CONDA="python=3.4"
-    - CONDA="python=3.5"
+matrix:
+  fast_finish: true
+  include:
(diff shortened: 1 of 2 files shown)


ocefpaf:
@lukecampbell can you take a look at this? It solved #451 and current `master` builds.
It also split the test context into `default` and `coding_standards` while allowing the latter to fail.

PR #480, opened 2017-04-10
Check for float without relying on np.float128

ocefpaf:
@lukecampbell it seems that @jbosch-noaa cannot use `np.float128` on some Windows machines.

This PR changes the logic to check if `scale_factor` is the primitive `float` or `np.floating` (which means any of those in the list there.)

PS: I incorporated #479 so we can run the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,28 +2,41 @@ language: python
 
 sudo: false
 
-env:
-    - CONDA="python=2.7"
-    - CONDA="python=3.4"
-    - CONDA="python=3.5"
+matrix:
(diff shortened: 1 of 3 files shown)


jbosch-noaa:
@lukecampbell - Would you please issue a bug fix release of the compliance-checker for this issue.  I  am probably not the only  one with this problem.

jbosch-noaa:
Thanks, @lukecampbell

lukecampbell:
https://github.com/conda-forge/compliance-checker-feedstock/pull/13

PR #481, opened 2017-04-11
Remove python-dateutil and workaround httpretty

ocefpaf:
`httpretty` is not Python 3 compliant and the packages in `defaults` should be used with care (`conda-forge`) does not package them BTW. (See https://github.com/gabrielfalcao/HTTPretty#disclaimer)

However the problem was being masked because the tests where running on Python 2.7 only due to a weird bug: the presence of `python-datetutil` in the requirements list. (Check the previous logs and [this](https://travis-ci.org/ocefpaf/compliance-checker/builds/220922447) without `python-datetutil`.)

So this PR removes `python-datetutil`, to enable Python 3 tests, and works around the lack of `httpretty` in `test_ioos.py`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,13 +20,16 @@ matrix:
     env: TEST_TARGET=coding_standards
 
 before_install:
-    - wget http://bit.ly/miniconda -O miniconda.sh
-    - bash miniconda.sh -b -p $HOME/miniconda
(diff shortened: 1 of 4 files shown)


ocefpaf:
@jbosch-noaa is this passes and with #480 merged we should probably issue a minor release to fix your problems there. I am working on adding Windows tests too to avoid any regressions but we don't need to wait for that to issue a release.

PR #482, opened 2017-04-11
Windows tests

ocefpaf:
In this PR:

- build the `conda-recipe` on Windows as a packaging test;
- enable Windows testing to avoid regression like #480;
- properly close files in the test framework to make them pass on  Windows.

Note that I enabled Windows test using my own account. Different from Travis-CI, where the members have access to the configuration and that is tied to the repo, with AppVeyor the configuration is tied to the user :unamused: 

Someone inside IOOS should take charge of that to avoid problems in the future.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,48 @@
+environment:
+  matrix:
+    - TARGET_ARCH: x86
+      CONDA_PY: 27
+      CONDA_INSTALL_LOCN: C:\\Miniconda35-x64
+
+    - TARGET_ARCH: x64
+      CONDA_PY: 27
+      CONDA_INSTALL_LOCN: C:\\Miniconda35-x64
+
+    - TARGET_ARCH: x86
+      CONDA_PY: 36
+      CONDA_INSTALL_LOCN: C:\\Miniconda36-x64
+
+    - TARGET_ARCH: x64
+      CONDA_PY: 36
+      CONDA_INSTALL_LOCN: C:\\Miniconda36-x64
+
+platform:
+    - x64
+
+install:
+    # If there is a newer build queued for the same PR, cancel this one.
+    # The AppVeyor 'rollout builds' option is supposed to serve the same
+    # purpose but it is problematic because it tends to cancel builds pushed
+    # directly to master instead of just PR builds (or the converse).
+    - ps: if ($env:APPVEYOR_PULL_REQUEST_NUMBER -and $env:APPVEYOR_BUILD_NUMBER -ne ((Invoke-RestMethod `
+         https://ci.appveyor.com/api/projects/$env:APPVEYOR_ACCOUNT_NAME/$env:APPVEYOR_PROJECT_SLUG/history?recordsNumber=50).builds | `
+         Where-Object pullRequestId -eq $env:APPVEYOR_PULL_REQUEST_NUMBER)[0].buildNumber) { `
+           throw "There are newer queued builds for this pull request, failing early." }
+
+    # Add path, activate `conda` and update conda.
+    - cmd: set "PATH=%CONDA_INSTALL_LOCN%\\Scripts;%CONDA_INSTALL_LOCN%\\Library\\bin;%PATH%"
+    - cmd: conda update --yes --quiet conda
+    - cmd: call %CONDA_INSTALL_LOCN%\Scripts\activate.bat
+
+    - cmd: set PYTHONUNBUFFERED=1
+
+    - cmd: conda config --set show_channel_urls true
+    - cmd: conda config --add channels conda-forge
+    - cmd: conda install --yes conda-build
+    - cmd: conda info
+
+# Skip .NET project specific build phase.
+build: off
+
+test_script:
+    - "conda build conda.recipe"
--- a/compliance_checker/cf/util.py
+++ b/compliance_checker/cf/util.py
@@ -1,3 +1,4 @@
+import io
 import itertools
 import requests
 import os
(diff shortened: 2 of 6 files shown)


appveyor.yml
```diff
@@ -0,0 +1,48 @@
...
+    - TARGET_ARCH: x64
+      CONDA_PY: 27
+      CONDA_INSTALL_LOCN: C:\\Miniconda35-x64
+
+    - TARGET_ARCH: x86
+      CONDA_PY: 36
+      CONDA_INSTALL_LOCN: C:\\Miniconda36-x64
+
+    - TARGET_ARCH: x64
+      CONDA_PY: 36
```
ocefpaf: I am only testing on Python 2.7 and 3.6 because `AppVeyor` is very slow.

appveyor.yml
```diff
@@ -0,0 +1,48 @@
...
+    - cmd: conda config --set show_channel_urls true
+    - cmd: conda config --add channels conda-forge
+    - cmd: conda install --yes conda-build
+    - cmd: conda info
+
+# Skip .NET project specific build phase.
+build: off
+
+test_script:
+    - "conda build conda.recipe"
```
ocefpaf: The test call is inside the recipe.

compliance_checker/tests/test_cli.py
```diff
@@ -23,10 +23,11 @@ class TestCLI(TestCase):
     '''
 
     def setUp(self):
-        _, self.path = tempfile.mkstemp()
+        self.fid, self.path = tempfile.mkstemp()
 
     def tearDown(self):
         if os.path.isfile(self.path):
+            os.close(self.fid)
```
ocefpaf: Windows file lock prevents us from removing open files.

compliance_checker/tests/test_suite.py
```diff
@@ -125,6 +126,7 @@ def test_cdl_file(self):
             score_list, nc_points, nc_out_of = cs.standard_output(limit, checker, groups)
             # This asserts that print is able to generate all of the unicode output
             cs.non_verbose_output_generation(score_list, groups, limit, nc_points, nc_out_of)
+        ds.close()
```
ocefpaf: Same as https://github.com/ioos/compliance-checker/pull/482/files#r110933010

conda.recipe/meta.yaml
```diff
@@ -0,0 +1,41 @@
...
+    - compliance_checker
+  requires:
+    - pytest
+    - httpretty  # [not py3k]
+  imports:
+    - compliance_checker
+    - compliance_checker.cf
+    - compliance_checker.tests
+  commands:
+    - py.test -s -rxs -v compliance_checker
```
ocefpaf: Here is the line that runs the tests using the conda built package.

ocefpaf:
The last failing test is:

```
self = test_download_standard_name_table ( compliance_checker.tests.test_cf:TestCF.test_download_standard_name_table )
    def test_download_standard_name_table(self):
        """
            Test that a user can download a specific standard name table
            """
        version = '35'
    
        data_directory = create_cached_data_dir()
        location = os.path.join(data_directory, 'cf-standard-name-table-test-{0}.xml'.format(version))
        download_cf_standard_name_table(version, location)
    
        # Test that the file now exists in location and is the right version
        self.assertTrue(os.path.isfile(location))
        std_names = StandardNameTable(location)
>       self.assertEqual(std_names._version, version)
E       AssertionError: '36' != '35'
E       - 36
E       + 35
compliance_checker\tests\test_cf.py:390: AssertionError
```

It looks like the version parsing might be broken on Windows with Python 3. I will try to get a Windows laptop to debug this.

compliance_checker/cf/util.py
```diff
@@ -261,7 +262,7 @@ def _get(self, entrynode, attrname, required=False):
     def __init__(self, cached_location=None):
         if cached_location:
             try:
-                with open(cached_location, 'r') as fp:
+                with io.open(cached_location, 'r', encoding='utf-8') as fp:
```
ocefpaf: @lukecampbell do we really want to catch a generic `Exception` below? That was masking the real bug on Windows here preventing the user from using downloaded table.

Hopefully the code change here will fix it but maybe we should remove the `try`/`Exception` case and let it fail when it cannot read from the file.

ocefpaf:
> It looks like the version parsing might be broken on Windows with Python 3. I will try to get a Windows laptop to debug this.

I was wrong. It was not the version parsing but the file reading itself. The error was masked by a `try`/`Except` case. See https://github.com/ioos/compliance-checker/pull/482/commits/e38d3962571c3c9661309644d17d09a8c37171cb#r110988217

I fixed the file reading part but maybe we should remove the `try`/`Except` case and let it fail when a file cannot be read.

compliance_checker/cf/util.py
lukecampbell: There's a long, and probably never ending debate about exception handling among software engineers and computer scientists. On the one hand, and I'm in this boat, I like the philosophy, "fail early and often, so we can fix it". The other boat is larger, and has far more people in it, and they all say "don't punish the users, hide the failures, fail gracefully and if you can, try to log it." 

I would prefer not catching all exceptions in all but very rare cases that are absolutely clear and documented.

ocefpaf: I don't really want to get into the "debate" :wink: 

My concern is: if the goal is to allow the user to use a custom downloaded table then we should not have a fallback or the user will never know which file is being used.

ocefpaf:
@lukecampbell I am not sure if there are any Windows users out there relying on the downloaded table feature of `compliance-checker`. If there are many they could use a new release.

PR #483, opened 2017-04-11
Fix lints

ocefpaf:
This PR reduces the lints from:

```shell
1     E124 closing bracket does not match visual indentation
1     E125 continuation line with same indent as next logical line
9     E126 continuation line over-indented for hanging indent
3     E127 continuation line over-indented for visual indent
7     E128 continuation line under-indented for visual indent
1     E129 visually indented line with same indent as next logical line
2     E225 missing whitespace around operator
1     E231 missing whitespace after ','
1     E265 block comment should start with '# '
9     E302 expected 2 blank lines, found 1
6     E303 too many blank lines (2)
4     E305 expected 2 blank lines after class or function definition, found 1
1     E713 test for membership should be 'not in'
9     F401 'numpy.ma' imported but unused
1     F403 'from compliance_checker.cf.cf import *' used; unable to detect undefined names
2     F601 dictionary key 'cf_example_cell_measures' repeated with different values
2     F821 undefined name 'boundary_variable'
2     F841 local variable 'valid' is assigned to but never used
2     W391 blank line at end of file
3     W503 line break before binary operator
```

to

```shell
1     F821 undefined name 'boundary_variable'
```

Please merge with caution as some apperently unused variables and import were removed! I guess that `./compliance_checker/cf/cf.py` deserves most of the attention.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -33,6 +33,7 @@ def print_exceptions(f):
             print_exc()
     return wrapper
 
+
 __stdname_table__ = "v29"
 
 
@@ -972,8 +973,11 @@ class CFBaseCheck(BaseCheck):
                                                                standard_name)
 
         # If the variable is supposed to be unitless, it automatically passes
-        should_be_unitless = (variable.ndim == 0 or variable.dtype.char == 'S'
-                              or std_name_unitless)
+        should_be_unitless = (
+            variable.ndim == 0 or
+            variable.dtype.char == 'S' or
+            std_name_unitless
+        )
 
         valid_udunits = TestCtx(BaseCheck.LOW,
                                 "§3.1 Variable {}'s units are contained in UDUnits".format(variable_name))
@@ -1918,9 +1922,8 @@ class CFBaseCheck(BaseCheck):
 
         # if has a calendar, check that it is within the valid values
         # otherwise no calendar is valid
-        for time_var in \
-            ds.get_variables_by_attributes(calendar=lambda c: c is not None):
-            reasoning=None
+        for time_var in ds.get_variables_by_attributes(calendar=lambda c: c is not None):
+            reasoning = None
             valid_calendar = time_var.calendar in valid_calendars
 
             if not valid_calendar:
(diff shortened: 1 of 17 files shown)


compliance_checker/cf/cf.py
```diff
@@ -2513,7 +2516,7 @@ def check_cell_boundaries(self, ds):
             if boundary_variable_name not in ds.variables:
                 valid = False
                 reasoning.append("Boundary variable {} referenced by {} not "
-                                 "found in dataset variables".format(boundary_variable.name,
+                                 "found in dataset variables".format(boundary_variable_name,
```
ocefpaf: `boundary_variable.name` does not exists and this never failed, so I am guessing this if-clause is never True.
I changed it to `boundary_variable_name` because that one exists but I am not sure that is correct!

ocefpaf: PS: there is another one at ``./compliance_checker/cf/cf.py:2820:25: F821 undefined name 'boundary_variable'``

compliance_checker/cf/cf.py
```diff
@@ -2784,7 +2794,6 @@ def check_climatological_statistics(self, ds):
                 return ret_val
             # handle 1-d and 2d coordinate bounds
             if (clim_var.ndim + 1 != ds.variables[clim_var.climatology].ndim):
-                valid = False
```
ocefpaf: This seems unused. Is it OK to remove it @lukecampbell? (There are a few more below.)

compliance_checker/cf/cf.py
lukecampbell: It's likely that the tests don't capture the bad case which would cause this block to be executed.

compliance_checker/cf/cf.py
lukecampbell: Yeah ok to remove, I think. I don't see how this variable can be used in any way if this `if` block is executed.

ocefpaf:
Skipped `AppVeyor` here until we merge #482.

Based on @lukecampbell comments it would be nice to add a test case that catches https://github.com/ioos/compliance-checker/pull/483#discussion_r110968431 but that is not the scope of this PR.

PR #486, opened 2017-04-17
Adds better support for detecting dataset protocol

lukecampbell:
We recently found out that Hyrax and ERDDAP did not work because the
server intitally returns a HTTP 200/Content-type: text/html when an HTTP
request is made to a DODS endpoint. This is so user's don't see the
error page that THREDDS shows when a user visits a pure DODS endpoint.

We relied on this behavior of THREDDS and the too literal interpretation
of DAPv2's specification to identify services implementing OPeNDAP.

This patch includes better identification of services implementing
OPeNDAP, and SOS. As well as a faster check for binary datasets like
netCDF.

Identification now works by checking if 'xdods-server' exists in the
response headers. In the case for SOS, if the response headers has
'text/xml' in the content-type header, then it is treated as an SOS
response.

Binary netCDF files are checked by the filename suffix or the magic
number (first four bytes of a file).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,6 +7,8 @@ matrix:
   include:
   - python: 2.7
     env: TEST_TARGET=default
+  - python: 2.7
+    env: TEST_TARGET=integration
   - python: 3.4
     env: TEST_TARGET=default
   - python: 3.5
@@ -15,6 +17,8 @@ matrix:
     env: TEST_TARGET=default
   - python: 3.6
     env: TEST_TARGET=coding_standards
+  - python: 3.6
+    env: TEST_TARGET=integration
   allow_failures:
   - python: 3.6
     env: TEST_TARGET=coding_standards
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,10 @@ matrix:
   allow_failures:
   - python: 3.6
     env: TEST_TARGET=coding_standards
+  - python: 2.7
+    env: TEST_TARGET=integration
+  - python: 3.6
+    env: TEST_TARGET=integration
 
 before_install:
   - wget http://bit.ly/miniconda -O miniconda.sh
@@ -42,4 +46,8 @@ script:
 
   - if [[ $TEST_TARGET == "coding_standards" ]]; then
       flake8 --ignore=E501,E251,E221,E201,E202,E203 -qq --statistics . ;
+    fi
+
+  - if [[ $TEST_TARGET == "integration" ]]; then
+      py.test -m "integration" -s -rxs -v ;
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,7 +37,7 @@ install:
 
 script:
   - if [[ $TEST_TARGET == "default" ]]; then
-      py.test -s -rxs -v ;
+      py.test -k "not integration" -s -rxs -v ;
     fi
 
   - if [[ $TEST_TARGET == "coding_standards" ]]; then
--- a/compliance_checker/protocols/cdl.py
+++ b/compliance_checker/protocols/cdl.py
@@ -24,8 +24,9 @@ def is_cdl(filename):
     :param str filename: Absolute path of file to check
     :param str data: First chuck of data from file to check
     '''
-    if os.path.splitext(filename)[-1] == '.cdl':
-        return True
+    if os.path.splitext(filename)[-1] != '.cdl':
+        return False
+
     with open(filename, 'rb') as f:
         data = f.read(32)
     if data.startswith(b'netcdf') or b'dimensions' in data:
--- a/compliance_checker/protocols/cdl.py
+++ b/compliance_checker/protocols/cdl.py
@@ -1,0 +1,33 @@
+#!/usr/bin/env python
+'''
+compliance_checker/protocols/cdl.py
+'''
+import os
+
+
+def is_cdl(filename):
+    '''
+    Quick check for .cdl ascii file
+
+    Example:
+        netcdf sample_file {
+        dimensions:
+            name_strlen = 7 ;
+            time = 96 ;
+        variables:
+            float lat ;
+                lat:units = "degrees_north" ;
+                lat:standard_name = "latitude" ;
+                lat:long_name = "station latitude" ;
+        etc...
+
+    :param str filename: Absolute path of file to check
+    :param str data: First chuck of data from file to check
+    '''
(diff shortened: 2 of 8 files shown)


lukecampbell:
@ocefpaf I don't have a windows environment to test on, so I can only hope that it doesn't break any windows builds. I also touched .travis.yml to change the tests so that it doesn't run tests marked with `integration`. If one of those servers goes offline I don't want that affecting our builds/releases.

compliance_checker/protocols/netcdf.py
```diff
@@ -0,0 +1,59 @@
...
+'''
+compliance_checker/protocols/netcdf.py
+
+Functions to assist in determining if the URL points to a netCDF file
+'''
+
+
+def is_netcdf(url):
+    '''
+    Returns True if the URL is a valid OPeNDAP URL
```
daf: this docstring looks copied from `protocols/opendap.py`

compliance_checker/protocols/cdl.py
```diff
@@ -0,0 +1,33 @@
...
+                lat:units = "degrees_north" ;
+                lat:standard_name = "latitude" ;
+                lat:long_name = "station latitude" ;
+        etc...
+
+    :param str filename: Absolute path of file to check
+    :param str data: First chuck of data from file to check
+    '''
+    if os.path.splitext(filename)[-1] == '.cdl':
+        return True
```
daf: i'm not stuck on this comment, but i would argue that checking a filename extension does not always mean it's true that it is one - i'd flip this logic and say if it doesn't end with cdl, return False, and check contents like you do below.  But like I said, I'm not hung up on this, this is certainly fine.

compliance_checker/suite.py
```diff
@@ -511,87 +512,46 @@ def generate_dataset(self, cdl_path):
...
+        :param str ds_str: URL to the remote resource
+        '''
+        if opendap.is_opendap(ds_str):
+            return Dataset(ds_str)
         else:
-            # no doc? try the dataset constructor
-            ds = Dataset(ds_str)
+            response = requests.get(ds_str, allow_redirects=True)
+            if 'text/xml' in response.headers['content-type']:
+                return self.process_doc(response.content)
```
daf: I was wondering why checking for SOS didn't mirror the same `protocols` approach but then realized you want to operate on the response in one shot rather than request twice, that's fine.  Maybe a comment here how this is SOS just for discoverability in code.

ocefpaf:
> @ocefpaf I don't have a windows environment to test on, so I can only hope that it doesn't break any windows builds.

Neither do I :smile: that is why enabled AppVeyor testing here.

> I also touched .travis.yml to change the tests so that it doesn't run tests marked with integration. If one of those servers goes offline I don't want that affecting our builds/releases.

I would not prevent them from running but rather run them as `allow_failure` instead.

lukecampbell:
@ocefpaf how can I do that?

daf:
> I would not prevent them from running but rather run them as allow_failure instead.

Strongly backed.  However, do those show up somehow from the PR?  Would be nice to know if it passed with allowed failures.

compliance_checker/suite.py
lukecampbell: Yeah, I rewrote this section several times. I had a protocols/sos.py at one point but I didn't want to make two requests or have one function return a tuple (true, content) or something weird. I will def add a comment

ocefpaf:
> @ocefpaf how can I do that?

I can do that later (clocking out now) but if you want to try take a look at the `matrix` section of the `.travis.yml` file. You can create a new category named `integration` and run the as `allow_failures`. travis-CI will run the test but report green regardless of the result. Giving is a tool to diagnose but at the same time easy to ignore if we are in a hurry.

.travis.yml
```diff
@@ -18,6 +18,10 @@ matrix:
   allow_failures:
   - python: 3.6
     env: TEST_TARGET=coding_standards
+  - python: 2.7
+    env: TEST_TARGET=integration
+  - python: 3.6
+    env: TEST_TARGET=integration
```
ocefpaf: These need to be in the include too.

.travis.yml
```diff
@@ -37,9 +45,13 @@ install:
...
+      py.test -k "not integration" -s -rxs -v ;
     fi
 
   - if [[ $TEST_TARGET == "coding_standards" ]]; then
       flake8 --ignore=E501,E251,E221,E201,E202,E203 -qq --statistics . ;
     fi
+
+  - if [[ $TEST_TARGET == "integration" ]]; then
+      py.test -m "integration" -s -rxs -v ;
+    fi
```
ocefpaf: :+1:

compliance_checker/protocols/cdl.py
```diff
@@ -0,0 +1,34 @@
...
+            float lat ;
+                lat:units = "degrees_north" ;
+                lat:standard_name = "latitude" ;
+                lat:long_name = "station latitude" ;
+        etc...
+
+    :param str filename: Absolute path of file to check
+    :param str data: First chuck of data from file to check
+    '''
+    if os.path.splitext(filename)[-1] != '.cdl':
```
ocefpaf: This will throw away valid `cdl` files that are not with the proper extension. I am OK with that behavior (actually I like it) but it should be documented to avoid confusion.

compliance_checker/protocols/netcdf.py
```diff
@@ -0,0 +1,59 @@
...
+    Returns True if the URL points to a valid local netCDF file
+
+    :param str url: Location of file on the file system
+    '''
+    # Try an obvious exclusion of remote resources
+    if url.startswith('http'):
+        return False
+
+    # If it's a known extension, give it a shot
+    if url.endswith('nc'):
```
ocefpaf: How slow is the brute force? A bad dump of a netcdf file may have the proper extension but it won't be a netcdf file...

mwengren:
Thanks for resolving this issue @lukecampbell!

compliance_checker/protocols/netcdf.py
lukecampbell: If an empty or bad file has '.nc' it will crash when Dataset(ds_str) is executed, OR when an attribute of the dataset is accessed depending on the nature of the corruption.

compliance_checker/protocols/cdl.py
lukecampbell: I'll merge this and update the README

PR #492, opened 2017-05-08
print the exception

ocefpaf:
@lukecampbell there is something wrong with the URL construction when `compliance-checker` tries to download the `cf-standard-name-table.xml`. It forms:

`http://cfconventions.org/Data/cf-standard-names/2015)/src/cf-standard-name-table.xml`

note the `2015)` there. (I am investigating and will fix that in another PR soo.)

Since the current design ignores that and uses the packaged version I changed:
- `print` to a `warning`;
- print the actual error so users can investigate what went wrong;
- changed `Error` to `Problem` to avoid scaring user.


compliance_checker/cf/cf.py
```diff
@@ -148,9 +149,11 @@ def _find_cf_standard_name_table(self, ds):
 
             self._std_names = util.StandardNameTable(location)
             return True
-        except Exception:
+        except Exception as e:
             # There was an error downloading the CF table. That's ok, we'll just use the packaged version
-            print("Error fetching standard name table. Using packaged v{0}".format(self._std_names._version))
+            msg = ("Problem fetching standard name table:\n{0}\n"
+                   "Using packaged v{1}").format
+            warn(msg(e, self._std_names._version))
```
daf: this is a curious approach - since you don't reuse msg anywhere, why store the ref to `format` and call it as opposed to making msg the template string and and calling `format` in the warn line?

ocefpaf: I initially though I would have to reuse it. But I can change that approach.

ocefpaf:
FYI the error is because the dataset,

https://data.nodc.noaa.gov/thredds/dodsC/ioos/sccoos/scripps_pier/scripps_pier-2016.nc.html

has the `standard_name_vocabulary` attribute set to

```
standard_name_vocabulary: CF Standard Name Table (v28, 07 January 2015)
```

I am not sure if that is bad practice or if we should improve the logic to obtain the version number.
We should be getting `v28` and not `2015)`. That is happening because of the `.split`.

compliance_checker/cf/cf.py
```diff
@@ -148,9 +155,10 @@ def _find_cf_standard_name_table(self, ds):
 
             self._std_names = util.StandardNameTable(location)
             return True
-        except Exception:
+        except Exception as e:
             # There was an error downloading the CF table. That's ok, we'll just use the packaged version
-            print("Error fetching standard name table. Using packaged v{0}".format(self._std_names._version))
+            warn("Problem fetching standard name table:\n{0}\n"
```
ocefpaf: @daf is this better?

compliance_checker/cf/cf.py
```diff
@@ -122,7 +123,13 @@ def _find_cf_standard_name_table(self, ds):
         # Try to parse this attribute to get version
         version = None
         if 'cf standard name table' in standard_name_vocabulary.lower():
-            version = standard_name_vocabulary.split()[-1]
+            version = [s.strip('(').strip(')').strip('v').strip(',') for s in standard_name_vocabulary.split()]
+            # This assumes that table version number won't start with 0.
+            version = [s for s in version if s.isdigit() and len(s) <=2 and not s.startswith('0')]
```
ocefpaf: This is ugly and assumes that the number we are looking for ([1-42 at the moment](http://cfconventions.org/standard-names.html)) won't start with `0` and is not longer than two digits. Still, we can find ambiguous digits for the version and I'd rather escape to the packaged version.

I am open to suggestions but I would prefer a more "parseable" standard for that property.

compliance_checker/cf/cf.py
daf: :+1:

ocefpaf:
PR #493 should fix `AppVeyor` failures. We could also "fix" the recipe by applying the software pinning, but I'd rather have a relaxed version of the requirements in place.

compliance_checker/cf/cf.py
```diff
@@ -125,7 +126,13 @@ def _find_cf_standard_name_table(self, ds):
...
         version = None
         if 'cf standard name table' in standard_name_vocabulary.lower():
-            version = standard_name_vocabulary.split()[-1]
+            version = [s.strip('(').strip(')').strip('v').strip(',') for s in standard_name_vocabulary.split()]
+            # This assumes that table version number won't start with 0.
+            version = [s for s in version if s.isdigit() and len(s) <= 2 and not s.startswith('0')]
+            if len(version) > 1:
+                return False
+            else:
+                version = version[0]
```
ocefpaf: @Bobfrat do you want me to do something here? Maybe re-factor this in 3 PRs?
We have here:

- AppVeyor updates;
- "improved" parsing of the `standard_name_vocabulary` version;
- issue a warning instead of a print on the screen when failing to get the version.

Bobfrat: Nah sorry, this is ok to me, I'll merge

PR #498, opened 2017-05-17
Print messages about CF standard table downloading to stderr, fixes #497

### LINKED ISSUE #497: JSON output issue in 3.0.3 with ERDDAP URLs
I've been testing the new Compliance Checker release with ERDDAP (on Windows), and I see some different results in the output when using ERDDAP URLs with the JSON output format than OPeNDAP.

I made a test script that illustrates the issue: https://gist.github.com/mwengren/68f61b0e70868a57336b81f863d7dbc9

Or, I think you can see the same issue (in my environment at least) by running these commands:
```
compliance-checker -t cf -f json http://coastwatch.pfeg.noaa.gov/erddap/griddap/osuSstAnom

compliance-checker -t cf -f json http://ona.coas.oregonstate.edu:8080/thredds/dodsC/NANOOS/OCOS
```

The issue is that when I try to hit the ERDDAP URL, it includes what looks like a debug line in stdout:
```
Using cached standard name table v29 from C:\Users\Micah.Wengren\.local\share\compliance-checker\cf-standard-name-table-test-29.xml
{
  "cf": {
    "scored_points": 168,
    "low_count": 0,
...
```

When I try to parse this into JSON using subprocess.Popen (see the Gist), if fails because it can't parse the first line. 

I'm using 3.0.3:
```
$ compliance-checker --version
IOOS compliance checker version 3.0.3
```

Also, not sure if it matters, but I'm using Python 2.7 for these tests.

daf:


lukecampbell:
Thanks for this!

This fixes #497

Issue #501, opened 2017-07-19
CF module: misleading and erroneous results


martinjuckes:
I've been looking at the IOOS compliance checker cf module, and comparing against the cfchecker (available from pypi) which is a reference implementation. The comparison, using files with known errors, turned up one problem with the cfchecker (which I've raised on their discussion list) and a number of issues with the IOOS implementation listed below. 

The original files I used are from the WCRP CMIP5 archive ... available to registered users. In order to make the discussion easier I've created some anonymised and reduced (by spatial and temporal subsetting) files which have the same metadata errors: 

[https://drive.google.com/open?id=0B6AWgiTrQ0xFMHBEUVNXYzc3YUE](sos_Omon_exAA03_historical_r3i1p1_185001-200512_box.nc)
(1) The IOOS checker records use of Conventions=CF-1.4 as an error: this was the current version of the conventions at the time the file was written, so should not be considered as an error. Rather, it is an indication that the compliance checker should check the metadata against version 1.4 of the standard.

[https://drive.google.com/open?id=0B6AWgiTrQ0xFQ1J0bFVDNlpyV1k](tnhus_cfSites_exAA04_amip_r3i1p1_1998_slice.nc)
(2) The variable “aps” is referred to in a formula_terms attribute but not defined: IOOS misses this;
(3) IOOS is distributed with an old version of the CF standard name table which results in spurious errors … this could probably be fixed by switching the file .. but there is no obvious warning.
(4) IOOS records the omission of both a standard name and long name for a variable as an error, the CF standard states that use of at least one of these as “highly recommended” (the IOOS error message give the wrong section for this: 3.3 instead of 3.2)

[https://drive.google.com/open?id=0B6AWgiTrQ0xFX2hQMW9VTUtCQ1k](zsatcalc_Omon_exAA07_abrupt4xCO2_r1i1p1_013601-014012_box.nc) 
(5) Cf-checker lists one error in the cell_methods string, IOOS has this as a priority 2 message -- which is usually associated with a warning.
(6) The IOOS checker has a message that “latitude variable 'rlat' should define valid units for latitude” which is incorrect: the ‘rlat’ variable has standard name grid_latitude, and the definition of grid_latitude specifies that “degree”, as used in the above file, is correct.

[https://drive.google.com/open?id=0B6AWgiTrQ0xFbXhXTlgzS0MwU3M](hurs_Amon_exAA05_decadal1981_r1i1p1_198201-199112_box.nc) 
[...]

PR #502, opened 2017-07-22
Adds better reasoning statement for Conventions check

lukecampbell:
The checker checks the Conventions global attribute but only allows
'CF-1.6' currently. This patch improves the reasoning message indicating
that the checker can only check CF-1.6 at the moment.

https://github.com/ioos/compliance-checker/issues/501


Bobfrat:
👍

lukecampbell:
@Bobfrat you can merge if you want to. I'm just a contributor at this point.

Issue #491, opened 2017-05-02
CF check on number of cf_role variables has been removed in v3


mhidas:
CF ([section 9.5](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#coordinates-metadata), second paragraph) says:
>CF files that contain timeSeries, profile or trajectory featureTypes, should include only a single occurrence of a `cf_role` attribute; CF files that contain timeSeriesProfile or trajectoryProfile may contain two occurrences

Until v2.3, the CF checker used to check this. In v3 this got refactored. It still checks the values of the `cf_role` attribute, but doesn't check the number of variables that have it.

Bobfrat:
Closing, addressed in #503

PR #503, opened 2017-07-22
Fixes cf_role checks

lukecampbell:
The cf_role checks was returning an empty set without any actual checks.
This patch fixes the check for cf_role and addresses #491.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -3099,14 +3099,20 @@ class CFBaseCheck(BaseCheck):
         """
         ret_val = []
         valid_roles = ['timeseries_id', 'profile_id', 'trajectory_id']
+        variable_count = 0
         for variable in ds.get_variables_by_attributes(cf_role=lambda x: x is not None):
+            variable_count += 1
             name = variable.name
             valid_cf_role = TestCtx(BaseCheck.HIGH, '§9.5 {} contains a valid cf_role attribute'.format(name))
             cf_role = variable.cf_role
             valid_cf_role.assert_true(cf_role in valid_roles,
                                       "{} is not a valid cf_role value. It must be one of {}"
                                       "".format(name, ', '.join(valid_roles)))
-        return ret_val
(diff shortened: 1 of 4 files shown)


compliance_checker/cf/cf.py
```diff
@@ -3099,14 +3099,20 @@ def check_cf_role(self, ds):
...
             cf_role = variable.cf_role
             valid_cf_role.assert_true(cf_role in valid_roles,
                                       "{} is not a valid cf_role value. It must be one of {}"
                                       "".format(name, ', '.join(valid_roles)))
-        return ret_val
+        valid_cf_role.assert_true(variable_count < 2,
+                                  ('§9.5 states that datasets should not '
+                                   'contain more than two variables defining a '
+                                   'cf_role attribute.'))
+        return valid_cf_role.to_result()
```
lukecampbell: It's worth pointing out that before it was returning `ret_val` which was initially set to `[]` and never reassigned.

Bobfrat: @lukecampbell Do we need to know the featureType to check the correct number of occurrences of cf_role (1 or 2)? 
```
CF files that contain timeSeries, profile or trajectory featureTypes, should include only a single occurrence of a cf_role attribute; CF files that contain timeSeriesProfile or trajectoryProfile may contain two occurrences
```

lukecampbell: The first iteration I wrote did that but what if the user incorrectly labeled the feature? I wrote this check from the perspective of, "this dataset is horribly formatted and we can only try to make certain guesses about how it _should_ be." If the checker always got perfectly formatted data, there would not be a need for a checker. 

I leave the rest of the feature detection stuff up to the individual feature checks.

Bobfrat: Ok that makes sense. I think you want to ```assert_true variable_count < 3``` right?

lukecampbell: ah, you're right.

PR #513, opened 2017-08-04
Update instruction for usage from Python code.

cpaulik:
The given instructions print out nothing with the latest code since it overwrites `skip_checks` and `output_filename`

### DIFF
--- a/README.md
+++ b/README.md
@@ -416,10 +416,14 @@ from compliance_checker.runner import ComplianceChecker, CheckSuite
 check_suite = CheckSuite()
 check_suite.load_all_available_checkers()
 
-# Run cf and adcc checks with normal strictness, verbose text format to stdout
+# Run cf and adcc checks with verbosity level 0, normal strictness,
+# do not skip checks and output text format to stdout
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -419,7 +419,7 @@ check_suite.load_all_available_checkers()
 # Run cf and adcc checks with normal strictness, verbose text format to stdout
 return_value, errors = ComplianceChecker.run_checker(
                                 '/path/or/url/to/your/dataset',
-                                ['cf', 'acdd'], 0, 'normal', '-', 'text')
+                                ['cf', 'acdd'], 0, 'normal')
```
Bobfrat: Good catch! We're going to be updating the documentation as part of our next milestone. I like the idea of showing all the arguments in the example, so can you change this line to ```['cf', 'acdd'], 0, 'normal', None, '-', 'text')```?

cpaulik:
I've update the example to use the keywords. This should make it even clearer.

jbosch-noaa:
Just saw this ... The Python instructions are much better. Thanks. :)  I will put it to the test of a few other users and see what they think.

Issue #509, opened 2017-07-24
CF checker not catching units error in cell_methods comment


Bobfrat:
Originally part of #501

https://drive.google.com/open?id=0B6AWgiTrQ0xFbXhXTlgzS0MwU3M
(7) There is an error in the units declared in the cell_methods comment string -- only picked up by the cf-checker.

PR #516, opened 2017-09-18
Cell methods

benjwadams:
Adds fixes for cell methods requested for #509.  Additional test coverage.

### DIFF
--- a/compliance_checker/tests/test_cf.py
+++ b/compliance_checker/tests/test_cf.py
@@ -8,19 +8,14 @@ from netCDF4 import Dataset
 from tempfile import gettempdir
 from compliance_checker.tests.resources import STATIC_FILES
 from compliance_checker.tests import BaseTestCase
+from compliance_checker.tests.helpers import MockTimeSeries, MockVariable
 
 import os
 import re
 import sys
 import pytest
 
-
-class MockVariable(object):
-    '''
-    For mocking a dataset variable
-    '''
-    pass
-
+from operator import sub
 
 class TestCF(BaseTestCase):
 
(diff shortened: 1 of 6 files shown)


benjwadams:
The package dependency I added doesn't have a corresponding installation candidate in Conda for Python 3.4

benjwadams:
@ocefpaf, it's been a while since I've run into one of these conda build failures.  I'd need to rebuild the package for the affected versions and upload to conda-forge, correct?

ocefpaf:
@benjwadams conda-forge dropped support for Python 3.4 and I suggest we do the same here.

See https://github.com/ioos/compliance-checker/pull/518

compliance_checker/tests/test_cf.py
```diff
@@ -967,6 +962,29 @@ def test_check_cell_methods(self):
...
+        nc_obj = MockTimeSeries()
+        nc_obj.createVariable('temperature', 'd', ('time',))
+
+        temp = nc_obj.variables['temperature']
+        temp.cell_methods = 'lat: lon: mean depth: mean (interval: 20 meters)'
+        results = self.cf.check_cell_methods(nc_obj)
+        scored, out_of, messages = self.get_results(results)
+        result_dict = {result.name: result for result in results}
+        modifier_results = result_dict[u'§7.3.3 temperature has valid cell_methods modifiers']
+        self.assertTrue(sub(*modifier_results.value[::-1]) == 0)
```
Bobfrat: @benjwadams can you add a comment here?

Bobfrat:
@benjwadams just to summarize, this PR adds a check of valid units in a cell_methods string.

For example, in the issue example file the following would be flagged due to the misspelling of 'minutes'.
```
hurs:cell_methods = "time: mean (interval: 20 mintues)" ;
```

benjwadams:
@Bobfrat, yes and it also checks that the number specified after the interval is in fact parsable to a numeric type.

compliance_checker/tests/test_cf.py
benjwadams: Superceded by a simpler scoring check.

Issue #476, opened 2017-02-06
Error getting time_coverage attributes as non-ISO


lukecampbell:
The attached dataset gets an error from ACDD 1.3
[b5km_baa-max-7d_20160101.nc.zip](https://github.com/ioos/compliance-checker/files/754780/b5km_baa-max-7d_20160101.nc.zip)

```
time_coverage_extents_match            :2:     0/ 1 : time_coverage variables
                                                      are not formatted
                                                      properly. Please ensure
                                                      they are valid ISO-8601
                                                      time strings

```

Bobfrat:
time_coverage_extents_match is no longer getting flagged as not valid ISO-8601 date format in this dataset.

PR #517, opened 2017-09-18
Use pendulum for parsing iso dates

Bobfrat:


Bobfrat:
Hold fire on this one, still need to add a test. Addresses https://github.com/ioos/compliance-checker/issues/476

compliance_checker/util.py
```diff
@@ -15,15 +15,25 @@ def isstring(obj):
...
+    '''
+    Returns a naive datetime. parsed from an ISO-8601 input string
+
+    :param str date_str: An ISO-8601 string
+    '''
+    if len(date_str) > 10:
+        return isodate.parse_datetime(date_str).replace(tzinfo=None)
+    # Must be just a date. Parse as a python date, then convert to naive datetime
+    date_iso = isodate.parse_date(date_str)
+    return datetime.combine(date_iso, datetime.min.time()).replace(tzinfo=None)
```
lukecampbell: I would strongly recommend using a more developed library for parsing dates like `arrow`.

Bobfrat: I recommend not using Arrow when dealing with ISO-8601. Arrow is what we were using and what was causing the problem (they still haven't released a 1.0 version yet). From the Arrow docs:

http://arrow.readthedocs.io/en/latest/
```
Partial ISO-8601 support
```

Example of the problem
```
date_str = '20170919T100000'
arrow_obj = arrow.get(date_str)
  File "<stdin>", line 1, in <module>
  File "/Users/bobfratantonio/Documents/Dev/virtenvs/work/lib/python2.7/site-packages/arrow/api.py", line 23, in get
    return _factory.get(*args, **kwargs)
  File "/Users/bobfratantonio/Documents/Dev/virtenvs/work/lib/python2.7/site-packages/arrow/factory.py", line 164, in get
    dt = parser.DateTimeParser(locale).parse_iso(arg)
  File "/Users/bobfratantonio/Documents/Dev/virtenvs/work/lib/python2.7/site-packages/arrow/parser.py", line 112, in parse_iso
    return self._parse_multiformat(string, formats)
  File "/Users/bobfratantonio/Documents/Dev/virtenvs/work/lib/python2.7/site-packages/arrow/parser.py", line 246, in _parse_multiformat
    raise ParserError('Could not match input to any of {0} on \'{1}\''.format(formats, string))
arrow.parser.ParserError: Could not match input to any of [u'YYYY-MM-DDTHH:mm'] on '20170919T100000'
```

ocefpaf: @Bobfrat arrow need to know the format. In that case I believe what you want is:


```python
arrow.get('20170919T100000', 'YYYYMMDDThhmmss').naive
datetime.datetime(2017, 9, 19, 10, 0)
```

lukecampbell: the `isodate` project is dead. It hasn't been worked on since 2015, there are outstanding bugs and as of 6 days ago there is a bug relating to python 3 support. arrow is an active project with extremely strong community support. You can do complete support for ISO dates using arrow, the partial support likely stems from supporting "[duration](https://en.wikipedia.org/wiki/ISO_8601#Durations)" which I haven't seen time libraries that DO support it.

ocefpaf: I am :+1: for `arrow` but I would specify a few expected formats to ensure the proper conversion.

lukecampbell: I stand corrected, the partial support refers to the date and week-date formats too, arrow supports it but it's gotta be explicitly specified. This creates a problem, how do you know what format a user elects from within the ISO-8601 standard?

Bobfrat: If we have to specify and try to parse all the possible iso formats, what good is arrow? Can do the same thing using datetime right?

ocefpaf: > If we have to specify and try to parse all the possible iso formats, what good is arrow? Can do the same thing using datetime right?

You can, but a try/except case with arrow would catch all the supported and then try the known/expected, in an elegant, short, and readable way.

lukecampbell: You can do the same thing with datetime but arrow is easier to deal with the time zones.

brianmckenna: Suggest [pendulum](https://github.com/sdispater/pendulum), covers some drawbacks of [arrow](https://github.com/crsmithdev/arrow)

https://github.com/sdispater/pendulum#why-not-arrow

3rd example in link shows format that is realistic for our case and could be problematic

Bobfrat: I'm sold on Pendulum. I went to the [ISO wikipedia page](https://en.wikipedia.org/wiki/ISO_8601) and tested all of the formats they had listed and Pendulum correctly parsed each one. @ocefpaf it looks like its on conda-forge as well https://github.com/conda-forge/pendulum-feedstock


```
>>> import pendulum

>>> pendulum.parse('2017-09-19')
<Pendulum [2017-09-19T00:00:00+00:00]>

>>> pendulum.parse('2017-09-19T23:06:17+00:00')
<Pendulum [2017-09-19T23:06:17+00:00]>

>>> pendulum.parse('2017-09-19T23:06:17Z')
<Pendulum [2017-09-19T23:06:17+00:00]>

>>> pendulum.parse('20170919T230617Z')
<Pendulum [2017-09-19T23:06:17+00:00]>

>>> pendulum.parse('2017-W38')
<Pendulum [2017-09-18T00:00:00+00:00]>

>>> pendulum.parse('2017-W38-2')
<Pendulum [2017-09-19T00:00:00+00:00]>

>>> pendulum.parse('--09-19')
<Pendulum [2017-09-19T00:00:00+00:00]>

>>> pendulum.parse('2017-262')
<Pendulum [2017-09-19T00:00:00+00:00]>
>>>
```

ocefpaf: @Bobfrat if you are OK with `pendulum` go for it. I have to say that I am against all parsers that are assuming things for me. With that said, `arrow` can be misused in the same way. I use arrow only for its simpler string data formatters, but I always specify the range of known/expected.

lukecampbell: This discussion has left me disappointed in the lack of consistency in the ISO-8601 standard. I don't think the CF community had intended datetimes to be used like this, it's more likely the CF standard was meant for the YYYY-MM-DD format only.

benjwadams:
Conda!

compliance_checker/util.py
Bobfrat: @lukecampbell these iso checks are specifically for [ACDD](http://wiki.esipfed.org/index.php/Attribute_Convention_for_Data_Discovery_1-3). In the attribute content guidance section of that link reads: 

> The ACDD specifies ISO 8601:2004 date and time formats for its temporal attributes. 
> ACDD strongly encourages the use of the 'extended' format date-time, in the form
> YYYY-MM-DDThh:mm:ss<zone>

So maybe the right thing to do in `check_time_extents` is to parse `time_coverage_start` and `time_coverage_end` attributes using pendulum to verify that the dates indeed correspond to the first and last time stamps in the dataset. If that passes, then we check if they follow the ACDD recommended format of `YYYY-MM-DDThh:mm:ss<zone>`

PR #518, opened 2017-09-18
Drop Python 3.4 tests

ocefpaf:
conda-forge dropped support for `Python 3.4` and we are testing against `Python 3.5` and `3.6`, so it is safe to drop it from here too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,6 @@ matrix:
     env: TEST_TARGET=default
   - python: 2.7
     env: TEST_TARGET=integration
-  - python: 3.4
-    env: TEST_TARGET=default
   - python: 3.5
     env: TEST_TARGET=default
   - python: 3.6


benjwadams:
I'll merge.  Mainly want to unbreak the tests.  Do note that a number of major Linux distros (CentOS/RHEL 7) ship with Python 3.4 in the package manager.

ocefpaf:
>  Do note that a number of major Linux distros (CentOS/RHEL 7) ship with Python 3.4 in the package manager.

That is not a problem. Testing with `Python 3.5` and `3.6` is already an overkill, and compliance-checker does not make use of any exclusive feature from those to justify testing `3.*`.

Also, the issue lies with some dependencies that are platform specific and are not available not maintained for `Python 3.4`, like `netCDF4`, so there is nothing wrong with installing compliance-checker on `Python 3.4` as long as the dependencies do not come from `conda-forge`.

PR #520, opened 2017-09-25
Cell methods improvements

benjwadams:
Improve checking of cell methods contents which appear inside parentheses.  Currently these include standalone comments (no colon), interval specifications, and comments coming after standard elements.  Also checks if the element with a colon even exists and throws an error if it does not.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -2673,9 +2673,9 @@ class CFBaseCheck(BaseCheck):
         }
 
         ret_val = []
-        psep = regex.compile(r'((?P<vars>\w+: )+(?P<method>\w+) ?(?P<where>where (?P<wtypevar>\w+) '
-                              '?(?P<over>over (?P<otypevar>\w+))?| ?)(?P<brace>\(((?P<brace_wunit>\w+): '
-                              '(?P<interval_val>\w+) (?P<interval_unit>\w+)|(?P<brace_opt>\w+): (\w+))\))?)*')
+        # use zero-width assertion here?
+        psep = regex.compile(r'(?P<vars>\w+: )+(?P<method>\w+) ?(?P<where>where (?P<wtypevar>\w+) '
+                             '?(?P<over>over (?P<otypevar>\w+))?| ?)(?:\((?P<paren_contents>[^)]*)\))?')
 
         for var in ds.get_variables_by_attributes(cell_methods=lambda x: x is not None):
             if not getattr(var, 'cell_methods', ''):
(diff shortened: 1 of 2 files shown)


benjwadams:
Hold on merging this, I need to fix for strings like "lat: lon: mean depth: variance (interval 0.02 degrees interval: 2 m comment: using spherical ellipsoid)"

compliance_checker/cf/cf.py
```diff
@@ -2729,28 +2730,62 @@ def check_cell_methods(self, ds):
...
+                                # otherwise check that the comment is the last
+                                # item in the parentheses
+                                elif i != len(split_vals) - 1:
+                                    valid_modifier.messages.append('The non-standard "comment:" element must come after any standard elements in cell_methods for variable {}'.format(var.name))
+                                #
+                                else:
+                                    valid_modifier.score += 1
+                            else:
+                                valid_modifier.out_of += 1
+                                valid_modifier.messages.append('Invalid cell_methods keyword "{}" for variable {}. Must be one of [interval, comment]'.format(keyword, var.name))
```
lukecampbell: This is some awfully nested code, strongly recommend breaking the nested branches into functions.

benjwadams: Will do.

compliance_checker/cf/cf.py
```diff
@@ -2719,44 +2719,95 @@ def check_cell_methods(self, ds):
...
+        valid_info = TestCtx(BaseCheck.MEDIUM,
+                             '§7.3.3 {} has valid cell_methods modifiers'.format(var.name))
+        # if there are no colons, this is a simple comment
+        # TODO: are empty comments considered valid?
+        if ':' not in paren_contents:
+            valid_info.out_of += 1
+            valid_info.score += 1
+            return valid_info
+        # otherwise, split into k/v pairs
+        kv_pair_pat = r'(\S+:)\s+(.*(?=\s+\w+:)|[^:]+$)\s*'
```
benjwadams: Not having quotation marks or commas like many key value pair specifications made this rather tricky to parse.  Hopefully this is robust.

compliance_checker/cf/cf.py
```diff
@@ -2726,72 +2726,88 @@ def check_cell_methods(self, ds):
...
+        """
+        Checks that the spacing and/or comment info contained inside the
+        parentheses in cell_methods is well-formed
+        """
+        #valid_info = TestCtx(BaseCheck.MEDIUM,
+        #                            '§7.3.2 {} has valid cell_methods spacing/other info'.format(var.name))
+        valid_info = TestCtx(BaseCheck.MEDIUM,
+                             '§7.3.3 {} has valid cell_methods modifiers'.format(var.name))
+        # if there are no colons, this is a simple comment
+        # TODO: are empty comments considered valid?
```
lukecampbell: Yes, empty comments are valid, annoying, but valid.

PR #526, opened 2017-10-10
Add IOOS netcdf profile 1.1 checker

Bobfrat:


setup.py
```diff
@@ -50,7 +50,9 @@ def readme():
             'cf = compliance_checker.cf.cf:CFBaseCheck',
             'acdd-1.1 = compliance_checker.acdd:ACDD1_1Check',
             'acdd-1.3 = compliance_checker.acdd:ACDD1_3Check',
-            'ioos = compliance_checker.ioos:IOOSBaseCheck',
+            'ioos_sos = compliance_checker.ioos:IOOSBaseSOSCheck',
+            'ioos-0.1 = compliance_checker.ioos:IOOS0_1Check',
+            'ioos-1.1 = compliance_checker.ioos:IOOS1_1Check',
```
Bobfrat: We now have 3 different IOOS tests: sos, netCDF 0.1, and the new netCDF 1.1

compliance_checker/ioos.py
```diff
@@ -25,9 +28,6 @@ def _has_attr(cls, ds, attr, concept_name, priority=BaseCheck.HIGH):
 
         return Result(priority, val, concept_name, msgs)
 
-
-class IOOSNCCheck(BaseNCCheck, IOOSBaseCheck):
-
     @classmethod
     def _has_var_attr(cls, dataset, vname, attr, concept_name, priority=BaseCheck.HIGH):
```
benjwadams: Just a minor comment here: `_has_var_attr` sounds very similar to the Python builtin `hasattr`.  We may want to change this down the road.

compliance_checker/tests/test_ioos_profile.py
```diff
@@ -0,0 +1,293 @@
...
+            self.assert_result_is_good(result)
+
+    def test_bad_platform_variables(self):
+        '''
+        Tests that the platform variable attributes check is working
+        '''
+
+        # Create an empty dataset that writes to /dev/null This acts as a
+        # temporary netCDF file in-memory that never gets written to disk.
+        nc_obj = Dataset(os.devnull, 'w', diskless=True)
```
benjwadams: Keep in mind here that if you're doing some parallel test runs that multiple open files to `os.devnull` will throw an exception in the NetCDF library even though the file isn't actually going to be writing to disk.

Bobfrat: Good point, we use this all over the project. Can you suggest a better solution and create an issue to fix this for all the tests?

compliance_checker/ioos.py
```diff
@@ -157,23 +189,175 @@ def check_variable_units(self, ds):
...
+    https://github.com/ioos/compliance-checker/issues/69
+    https://github.com/ioos/compliance-checker/issues/358
+    '''
+    _cc_spec_version = '1.1'
+    _cc_description = 'IOOS Metadata Profile for NetCDF, Version 1.1'
+    _cc_url = 'https://ioos.github.io/ioos-netcdf/ioos-netcdf-metadata-description-v1-1.html#ioos-netcdf-metadata-profile-attributes'
+    register_checker = True
+
+    def __init__(self):
+        super(IOOS1_1Check, self).__init__()
```
lukecampbell: I would strongly caution against using `super` and instead using 

```python
IOOS1_1Check.__init__(self)
```

I don't remember if this is the right reference as to why but it might link to it.
https://fuhm.net/super-harmful/

All I remember is that `super` bit us really really hard in the past.

compliance_checker/ioos.py
```diff
@@ -211,7 +395,7 @@ def check_suggested(self, ds):
         ]
 
 
-class IOOSSOSDSCheck(BaseSOSDSCheck, IOOSBaseCheck):
+class IOOSSOSDSCheck(BaseSOSDSCheck, IOOSBaseSOSCheck):
```
lukecampbell: `super` is definitely dangerous when you have multiple inheritance.

compliance_checker/ioos.py
Bobfrat: I'll check it out, I'm wondering if I need that line at all

PR #527, opened 2017-10-11
Formula terms fix

benjwadams:


benjwadams:
I'm assuming the test failures here with Py 2.7 are due to the output of sorted() being different than in 3.x.  I'll have a look tomorrow.

benjwadams:
Forgot that printing a list's `repr` in Python 2.7 will prepend `u` to Unicode strings.  This should fix.

compliance_checker/cf/cf.py
```diff
@@ -1796,31 +1796,37 @@ def _check_formula_terms(self, ds, coord):
...
+        # matches.  If it doesn't exactly match the original, the formatting
+        # of the attribute is incorrect
+        reconstructed_formula = ' '.join(m[0] + m[1] for m in matches)
+        valid_formula_terms.assert_true(reconstructed_formula == formula_terms,
+                                        "Attribute formula_terms is not well-formed")
+
         valid_formula_terms.assert_true(standard_name in dimless,
                                         "unknown standard_name for dimensionless vertical coordinate: {}"
                                         "".format(standard_name))
         if standard_name not in dimless:
```
benjwadams: `in dimless_vertical_coordinates` since it is a dict.

PR #540, opened 2017-11-17
Added new supported dataset/file type: GenericFile

agstephens:
The `GenericFile` is a file that is only checked from "outside". The contents are not read but the file must exist. This enables checks of directory/file names and the file size. Potentially other file metadata obtained from the file system (such as creation time) could also be checked.

STFC CEDA is building new checkers that use this file type.


agstephens:
It looks like the 3 tests I added are passing.

Bobfrat:
Thanks @agstephens! Looks good but the Travis CI tests are failing. I'll see if I can track down why

agstephens:
Thanks @Bobfrat, I found that some of the tests were failing when I cloned Master. As far as I can see the three new tests pass. One test in `test_acdd.py` fails for me and all tests under `test_cf_integration.py` fail (in my local clone).

benjwadams:
@agstephens, this is related to #451.  Your code passes the tests when the NetCDF files already exist, but attempting to generate the NetCDF files during the tests causes it to fail.

agstephens:
@benjwadams: thanks for checking it out. I'm not clear about what has caused the failures. Should there be something that generates all the NC files before running any tests?

agstephens:
I have just cloned the current master and ran with `pytest` I found that about 30 tests fail on my system (python2.7 with a virtualenv).  

```
pytest
======= test session starts =========
platform linux2 -- Python 2.7.3, pytest-3.2.5, py-1.5.2, pluggy-0.4.0
rootdir: /vagrant-share/checker-bundle/ioos/compliance-checker, inifile: pytest.ini
collected 140 items

compliance_checker/tests/test_acdd.py .............
compliance_checker/tests/test_base.py ...
compliance_checker/tests/test_cf.py .............................................
compliance_checker/tests/test_cf_integration.py FFFFFFFFFFFFFFF
compliance_checker/tests/test_cli.py FFFFFFF
compliance_checker/tests/test_feature_detection.py ............................
compliance_checker/tests/test_ioos_profile.py ................
compliance_checker/tests/test_ioos_sos.py FF
compliance_checker/tests/test_protocols.py ....
compliance_checker/tests/test_suite.py FF.FFF
compliance_checker/tests/test_util.py .
```

An example failure is:
```
=
____________________ TestCFIntegration.test_ww3 _________________________

self = test_ww3 ( compliance_checker.tests.test_cf_integration:TestCFIntegration.test_ww3 )

    def test_ww3(self):
        dataset = self.load_dataset(STATIC_FILES['ww3'])
>       check_results = self.cs.run(dataset, [], 'cf')

compliance_checker/tests/test_cf_integration.py:245:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
compliance_checker/suite.py:131: in run
    checkers = self._get_valid_checkers(ds, checker_names)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <compliance_checker.suite.CheckSuite object at 0x2836f90>
ds = <[RuntimeError("NetCDF: Not a valid ID") raised in repr()] SafeRepr object at 0x28f4440>, checker_names = ('cf',)

    def _get_valid_checkers(self, ds, checker_names):
        """
            Returns a filtered list of 2-tuples: (name, valid checker) based on the ds object's type and
            the user selected names.
            """
>       assert len(self.checkers) > 0, "No checkers could be found."
E       AssertionError: No checkers could be found.

compliance_checker/suite.py:97: AssertionError

```

It looks like it is able to read the NC file but the checks can't be found.

Bobfrat:
@agstephens I'll try to clarify what's going on here.

Travis CI skips the integration tests that you are seeing errors on. It runs this command:

```
py.test -k "not integration" -s -rxs -v
```

But your PR is failing due to a new version of ncgen as @benjwadams has summarized [here](https://github.com/ioos/compliance-checker/issues/541).  ncgen is called to generate the .nc files in compliance_checker/tests/data. For the sake of project size we have committed .cdl files for testing and when you run the tests for the first time, the netCDF files are generated locally using a subprocess call to ncgen. https://github.com/ioos/compliance-checker/blob/master/compliance_checker/tests/resources.py#L12-L18

@ocefpaf @daf @lukecampbell should we try to fix the cdl's and possibly the tests? Or we could commit the netCDF files? It's about 46 MB. I'm open to other suggestions too.

agstephens:
Hi @Bobfrat, thanks for the clarification. For another test tool (that I'd like to integrate with the compliance-checker) we have taken many real data files and reduced them to a manageable size using the NCO tools (in particular `ncks`) in order to sub-sample the actual data in space and time. 

The test data directory in the repo is here:
 https://github.com/cedadev/time-checks/tree/master/test_data

There are over 50 NetCDF files currently under 8MB in size (which seems reasonable to me). Could a similar approach be taken with the examples used for the compliance-checker?

A typical use of `ncks` to sub-sample in lat/lon space is:

```
ncks -d lat,,,100 -d lon,,,100 -v <var_id> <input_file> <output_file>
```

Bobfrat:
@agstephens Thanks for the tip! That seems like a reasonable approach to me, I'll take a look.

Bobfrat:
@ocefpaf do you know if we can kick off a new appveyor build? Travis CI is passing now so thats a good sign.

ocefpaf:
> @ocefpaf do you know if we can kick off a new appveyor build? Travis CI is passing now so thats a good sign.

I restarted it but sometimes opening and closing the PR can toggle the CIs. However, it would be nice to lay out a plan to create an AppVeyor account for IOOS and move all the tests from my instance to that one :wink:

compliance_checker/tests/test_base.py
```diff
@@ -81,3 +81,18 @@ def verify_dummy(ds):
...
         assert rv3[0] == base.Result(priority, True, 'dummy', [])
+
+
+class TestGenericFile(TestCase):
+    '''
+    Tests the GenericFile class.
+    '''
+
+    def test_create_GenericFile_success(self):
+        path = '/tmp/test.txt'
```
Bobfrat: @agstephens this is just testing the pathname not opening a file, right? Otherwise we'll need to support Windows path names too (os.path.join).

agstephens: Hi @Bobfrat, that is correct. `GenericFile` does not look inside the file. It only looks at it from outside. So it might: 1. Check directory and file name match fixed rules, 2. Get size of the file., etc.

mwengren:
> I restarted it but sometimes opening and closing the PR can toggle the CIs. However, it would be nice to lay out a plan to create an AppVeyor account for IOOS and move all the tests from my instance to that one 😉

@ocefpaf I have no idea if this is possible or not, but if it's necessary we should probably track it elsewhere so we are aware there's a need.  We are first trying to figure out how to comply with NOAA's GitHub guidelines, so this would probably have to come after that, especially if it costs $.

agstephens:
@Bobfrat, thanks for doing the merge. Will start looking at the next integration we discussed a few months ago.

ocefpaf:
> @ocefpaf I have no idea if this is possible or not, but if it's necessary we should probably track it elsewhere so we are aware there's a need

It is possible. We need a dummy/bot e-mail account to open a GitHub account and then it is a matter of enabling AppVeyor and creating a team that has access to it.

> We are first trying to figure out how to comply with NOAA's GitHub guidelines, so this would probably have to come after that, especially if it costs $.

There is no $ cost but I don't know about the guidelines. We do have a but account already btw: `comtbot`, for Coastal and Ocean Modeling Testbed. I am not sure if we should reuse it here though, maybe a more general one would be better.

PR #545, opened 2017-12-07
Support showing output from multiple tests in HTML output

joesingo:
Previously only the last test to run would be shown in the HTML output. In fact, the results for each test would be written to the same output file sequentially, overwriting the results of the previous test.

For example, after running `checker.py -o output.html -f html --test cf --test acdd <dataset>`, `output.html` would only show results for the `acdd` test.

This change splits the HTML template into two files
* one with the header bar, CSS etc
* one to show the results of a single test

The second template is rendered for each test and included in the first.

### DIFF
--- a/compliance_checker/data/templates/ccheck.html.j2
+++ b/compliance_checker/data/templates/ccheck.html.j2
@@ -1,351 +1,137 @@
-<!doctype html>
-<html>
-  <head>
-    <meta charset="UTF-8">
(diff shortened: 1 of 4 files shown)


Bobfrat:
@joesingo This is great, thank you for contributing. @benjwadams care to have a look?

benjwadams:
@joesingo, thanks for the fix!

PR #547, opened 2017-12-18
Fix running checks on multiple datasets

joesingo:
Multiple arguments were accepted for `dataset_location`, but the
checker was only actually run on the first one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -141,7 +141,7 @@ optional arguments:
   -f {text,html,json}, --format {text,html,json}
                         Output format
   -o OUTPUT, --output OUTPUT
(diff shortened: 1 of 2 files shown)


Bobfrat:
👍  Nice catch! Do you have to pass in multiple output files so it doesn't overwrite on each iteration?

joesingo:
@Bobfrat Of course, you're right. How about allowing `--output` to be given multiple times, and checking that the number of output files and datasets is the same. Then each output file could be paired up with a dataset

Bobfrat:
Yeah that should do the trick

PR #559, opened 2018-03-12
Add the check version number in all output

Bobfrat:
@benjwadams @daltonkell 
Compliance checker lets you pass in check names such as ```acdd``` or ```acdd:latest``` and the output passes these name through as well, making it impossible to know which version of ACDD was actually used. This puts the version number into the output so its very clear which version was used.

Also, small updates to the priority header function


benjwadams:
I'll merge this, but I'm also going to cherry-pick #558 into this branch so we can have consistent numbers for "latest"

PR #560, opened 2018-03-12
Move Suite.get_priority_headers into Checker based class attributes

daf:
To prevent staleness and allow 3rd party plugins to override priority names, a new class attribute (`_cc_display_headers`) is introduced mapping priority names to display strings.  If not defined in the derived checker, the base ones are used.

- create _cc_display_headers attr in each Checker
- make checker name a required positional argument where used

Will need to be done in the NCEI suite of checkers too (I can do this if no one else can get to it, but tell me).


Bobfrat:
I can update NCEI plugins, thanks @daf

Issue #103, opened 2015-04-23
Results grouping can effectively increase priority of a fail


mhidas:
For example, in [this file](http://thredds.aodn.org.au/thredds/catalog/IMOS/eMII/checker_test/ANMN/timeSeries/catalog.html?dataset=IMOS/eMII/checker_test/ANMN/timeSeries/IMOS_ANMN-AM_GST_20140407T040000Z_NRSMAI_FV01_NRSMAI-CO2-1404-delayed_END_20140922T220000Z_C-20150311T084035Z.acceptable_units.nc) (with `LATITUDE:units = "degree_north"`) the only failed checks are for latitude and longitude units:

```
latitude                               :3:     3/ 4 :  
    LATITUDE                           :3:     3/ 4 :  
        correct_units                  :2:     2/ 3 : LATITUDE units are
                                                      acceptable, but not
                                                      recommended
```

The check that failed is MEDIUM priority, but because of the way result values are aggregated, it makes it look like a HIGH priority fail. So even if I think "acceptable" units are good enough and specify `--criteria=lenient` on the command line, this file will still fail the CF checks.

mhidas:
I don't know yet how many checks are affected by this issue, but it could potentially be an annoying problem for the way we'd like to use this checker at IMOS (see https://github.com/ioos/compliance-checker/issues/37#issuecomment-93616335).

Suggested solutions:
* Only group together checks of the same priority level; or
* When deciding on the final PASS/FAIL result, only consider the value & priority of each individual check, not the aggregated score & top-level priority of each group.

lukecampbell:
I can take a look at the file if you could provide an updated URL for it. As far as regrouping goes, the grouping and aggregations are based on the language used in the documents. For example when ACDD uses the language "highly recommended" the check counts as a high priority or 3 points. When ACDD uses language like "recommended" it's a medium check. 

I'll take regrouping into consideration.

mhidas:
@lukecampbell Sorry, we've made some  changes to our THREDDS server. [Here](http://thredds.aodn.org.au/thredds/catalog/IMOS/eMII/checker_test/ANMN/timeSeries/catalog.html?dataset=eMII/checker_test/ANMN/timeSeries/IMOS_ANMN-AM_GST_20140407T040000Z_NRSMAI_FV01_NRSMAI-CO2-1404-delayed_END_20140922T220000Z_C-20150311T084035Z.acceptable_units.nc) is an updated link to the file mentioned above.

[...]

PR #561, opened 2018-03-12
Groups

Bobfrat:
Remove the score_groups but use sensible names that will get grouped during the reasoning routine.
Also cleans up the HTML and standard out reports.

Addresses
https://github.com/ioos/compliance-checker/issues/103

### DIFF
--- a/compliance_checker/base.py
+++ b/compliance_checker/base.py
@@ -340,6 +340,12 @@ def ratable_result(value, name, msgs):
 
 
 def score_group(group_name=None):
+    '''
+    Warning this is deprecated as of Compliance Checker v3.2!
+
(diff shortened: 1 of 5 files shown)


compliance_checker/base.py
```diff
@@ -340,6 +340,12 @@ def ratable_result(value, name, msgs):
 
 
 def score_group(group_name=None):
+    '''
+    Warning this is deprecated as of Compliance Checker v3.2!
+
+    Please do not using scoring groups and update your plugins
+    if necessary
+    '''
     def _inner(func):
```
benjwadams: Consider using `warnings.warn` to notify that score grouping is deprecated.

PR #563, opened 2018-03-13
Fix #562 duplicate axis coordinates

### LINKED ISSUE #562: CF duplicate axis check fails incorrectly
For a file with the following structure:
```
dimensions:
        TIME = 26577 ;
        HEIGHT_ABOVE_SENSOR = 25 ;
        DIST_ALONG_BEAMS = 25 ;

variables:
        double TIME(TIME) ;
                TIME:axis = "T" ;
        double LATITUDE ;
                LATITUDE:axis = "Y" ;
        double LONGITUDE ;
                LONGITUDE:axis = "X" ;
        float HEIGHT_ABOVE_SENSOR(HEIGHT_ABOVE_SENSOR) ;
                HEIGHT_ABOVE_SENSOR:axis = "Z" ;
        float NOMINAL_DEPTH ;
                NOMINAL_DEPTH:axis = "Z" ;
        float DIST_ALONG_BEAMS(DIST_ALONG_BEAMS) ;
                // no axis attribute (this is not strictly a depth/height coordinate)

        float ABSIC1(TIME, DIST_ALONG_BEAMS) ;
                ABSIC1:coordinates = "TIME LATITUDE LONGITUDE DIST_ALONG_BEAMS" ;
                ABSIC1:_FillValue = 999999.f ;
                ABSIC1:long_name = "backscatter_intensity_from_acoustic_beam_1" ;
                ABSIC1:units = "count" ;
```
I get a failing CF check with the message:
```
§5.0 Variable ABSIC1 does not contain d:3:     5/ 6 : duplicate axis Z defined
                                                      by NOMINAL_DEPTH
```
This is referring to the section of [CF](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#coordinate-system) (6th paragraph) that says:
> it is not permissible for a data variable to have both a coordinate variable and an auxiliary coordinate variable, or more than one of either type of variable, having an axis attribute with any given value

However in this case the `ABSIC1` variable has _no_ associated coordinate variables of either kind that define an axis="Z" attribute, so I think the checker is wrong.

It's true that there are two variables int he file that have axis="Z" (`HEIGHT_ABOVE_SENSOR` and `NOMINAL_DEPTH`), but I think this is allowed, as long as no other variable names _both_ of them as associated coordinate variables, which is the case in this file.

mhidas:
This proposed solution fixes the problem I reported in #562 for that particular test files. The change makes sense to me because I don't really understand the logic behind the lines that I've removed. However, there may be a good reason to keep them! I'm just putting this up for discussion and review.

This broke one unittest:
compliance_checker/tests/test_feature_detection.py::TestFeatureDetection::test_global_feature_detection
I have updated this test so that it now passes.

### DIFF
--- a/compliance_checker/cfutil.py
+++ b/compliance_checker/cfutil.py
@@ -677,20 +677,6 @@ def get_axis_map(ds, variable):
             if coord_name not in axis_map[axis]:
                 axis_map[axis].append(coord_name)
 
-    # Sometimes dimensionless coordinates are considered coordinate variables.
-    # I know, it's confusing.
-    if 'X' not in axis_map and longitudes:
-        axis_map['X'] = longitudes
-
-    if 'Y' not in axis_map and latitudes:
-        axis_map['Y'] = latitudes
-
-    if 'T' not in axis_map and times:
-        axis_map['T'] = times
-
-    if 'Z' not in axis_map and heights:
-        axis_map['Z'] = heights
-
     return axis_map
 
 
--- a/compliance_checker/tests/test_feature_detection.py
+++ b/compliance_checker/tests/test_feature_detection.py
@@ -283,8 +283,9 @@ class TestFeatureDetection(TestCase):
 
             axis_map = util.get_axis_map(nc, 'temperature')
             assert axis_map['T'] == ['time']
-            assert axis_map['Y'] == ['lat']
-            assert axis_map['X'] == ['lon']
+            assert axis_map['Z'] == []
+            assert axis_map['Y'] == []
+            assert axis_map['X'] == []
 
         with Dataset(resources.STATIC_FILES['index_ragged']) as nc:
             assert util.guess_feature_type(nc, 'temperature') == "single-trajectory"
@@ -300,7 +301,7 @@ class TestFeatureDetection(TestCase):
 
             axis_map = util.get_axis_map(nc, 'sea_surface_height')
             assert axis_map['T'] == ['time']
-            assert axis_map['Z'] == ['z']
+            assert axis_map['Z'] == []
             assert axis_map['Y'] == ['lat']
             assert axis_map['X'] == ['lon']
 


mhidas:
Regarding the failing unittest, the test file looks like this:
```
netcdf climatology.nc {
dimensions:
    time=1;
    nv=2;
variables:
    float temperature(time);
        temperature:long_name = "surface air temperature";
        temperature:cell_methods = "time: mean within days time: mean over days";
        temperature:units = "K";

    float lat;
        lat:standard_name = "latitude";
        lat:units = "degrees_north";
    float lon;
        lon:standard_name = "longitude";
        lon:units = "degrees_east";
    double time(time);
        time:climatology = "climatology_bounds";
        time:units = "hours since 1997-4-1";
        time:standard_name = "time";
    double climatology_bounds(time, nv);
}
```

And the failing assertion:
```python
            axis_map = util.get_axis_map(nc, 'temperature')
            assert axis_map['Y'] == ['lat']
            assert axis_map['X'] == ['lon']
```
Given that the variable `temperature` has only `time` as a coordinate variable, and no `coordinates` attribute, why should the "axis_map" include `lat` and `lon`?
Or to put it the other way around, if those variables _are_ supposed to be the relevant coordinates for `temperature`, they _should_ be listed in a `coordinates` attribute.

benjwadams:
@lukecampbell, do you know the original rationale for the lines removed by this commit?

compliance_checker/cfutil.py
```diff
@@ -677,20 +677,6 @@ def get_axis_map(ds, variable):
             if coord_name not in axis_map[axis]:
                 axis_map[axis].append(coord_name)
 
-    # Sometimes dimensionless coordinates are considered coordinate variables.
```
mhidas: According to [CF](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#scalar-coordinate-variables):
> Since there is no associated dimension these scalar coordinate variables should be attached to a data variable via the coordinates attribute.

Alternatively, such a variable can be given a dimension of the same name (of size 1) and then be attached to other variables via its dimension.

Either way, the code above would find this coordinate variable. If it didn't find it, then it's not meant to be associated with the `variable` under consideration, and should _not_ be included in the `axis_map`.

mhidas:
I have updated the assertions in that unittest so that it now passes. The new assertions are what I would expect, based on what I think the `cfutil.get_axis_map()` function is trying to do, and the files being tested.

PR #580, opened 2018-05-15
pin packages until cf_units 2.0 is released

ocefpaf:
I can fix the conda-forge package without a new release but PyPI users will need a new release to get this change.

We need to revert that once `cf_units 2.0` is released. It should be out next week or so.

PS: there is one test failure and I think it is unrelated to this change. Can you double check that @Bobfrat?

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,12 +1,12 @@
 OWSLib>=0.8.3
 lxml>=3.2.1
-cf_units>=0.1.1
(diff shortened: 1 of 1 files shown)


Bobfrat:
When I run the tests on this branch with a fresh virtualenv, I don't get that error, just the FutureWarning you fixed in your latest PR. I'll keep digging though.

PR #581, opened 2018-05-15
from int to np.signedinteger is deprecated

ocefpaf:
This fixes the `FutureWarning` we are from latest `numpy`

```
compliance_checker/tests/test_cf.py::TestCF::test_check_flag_masks
  /home/travis/build/ocefpaf/compliance-checker/compliance_checker/cf/cf.py:1351: FutureWarning: Conversion of the second argument of issubdtype from `int` to `np.signedinteger` is deprecated. In future, it will be treated as `np.int64 == np.dtype(int).type`.
    type_ok = (np.issubdtype(variable.dtype, int) or
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -44,6 +44,7 @@ nosetests.xml
 coverage.xml
 *,cover
 .hypothesis/
+.pytest_cache/
 
 # Translations
 *.mo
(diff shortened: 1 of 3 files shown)


Bobfrat:
thanks @ocefpaf!

PR #584, opened 2018-05-15
Use versioneer

Bobfrat:


setup.cfg
```diff
@@ -0,0 +1,12 @@
...
+# See the docstring in versioneer.py for instructions. Note that you must
+# re-run 'versioneer.py setup' after changing this section, and commit the
+# resulting files.
+
+[versioneer]
+VCS = git
+style = pep440
+versionfile_source = compliance_checker/_version.py
+versionfile_build = compliance_checker/_version.py
+tag_prefix = v
```
Bobfrat: @ocefpaf all of the previous releases don't have the v prefix. So moving forward I'll need to tag using v.

ocefpaf: You don't have to use the v, I do recommend it though, makes the releases more readable.

Bobfrat:
@ocefpaf can you test this out? I removed the v prefix but still getting the following when asking for the version number with a fresh virtualenv and running python setup.py install

```
Finished processing dependencies for compliance-checker==4.0.0+10.ga2d8c84
(versioneer-test) ➜  compliance-checker git:(versioneer) python cchecker.py -V
IOOS compliance checker version 0.2.0+956.ga2d8c84
```

ocefpaf:
That actually looks about right if you are installing from `master` b/c `4.0.0` is the latest tag, `+` is the separator, and `10.ga2d8c84` is a GitHub hash for the latest commit. You'll get a clean version if you checkout the tag (after merging this).

Bobfrat:
yeah so that looked right but when I run python cchecker.py -V I'd expect the same thing but I get

```
IOOS compliance checker version 0.2.0+956.ga2d8c84
```

benjwadams:
We've been using numeric tags for quite some time now, so I see no reason to switch to using "v*" tags now.  Maybe I'm missing something, but I don't see a way to blacklist tags in versioneer.  There is the prefix argument, but it looks like it doesn't accept regex, which is what I'd want for accepting numeric args.

Bobfrat:
I can try deleting the v0.1 and v0.2.0 tags

Bobfrat:
Ok I thought i'd try renaming the tags without the v but that made it the most recent tag. I think I'll just delete them unless someone objects

Bobfrat:
We were able to delete the offending tags (plus 0.2.0 that I accidentally created). For anyone interested in staying in sync, you'll need to delete the old tags locally and on your forked repo (origin in my case).
```
git tag -d v0.1
git tag -d v0.2.0
git tag -d 0.2.0

git push origin :refs/tags/v0.1
git push origin :refs/tags/v0.2.0
git push origin :refs/tags/0.2.0
```

PR #585, opened 2018-05-18
Update prefix in _version.py

Bobfrat:
@ocefpaf this was the piece I needed to update after switching the prefix in setup.cfg

### DIFF
--- a/compliance_checker/_version.py
+++ b/compliance_checker/_version.py
@@ -41,7 +41,7 @@ def get_config():
     cfg = VersioneerConfig()
     cfg.VCS = "git"
     cfg.style = "pep440"
-    cfg.tag_prefix = "v"
+    cfg.tag_prefix = ""
     cfg.parentdir_prefix = ""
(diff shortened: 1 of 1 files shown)


compliance_checker/_version.py
```diff
@@ -41,7 +41,7 @@ def get_config():
     cfg = VersioneerConfig()
     cfg.VCS = "git"
     cfg.style = "pep440"
-    cfg.tag_prefix = "v"
+    cfg.tag_prefix = ""
```
ocefpaf: My guess is that this is OK but you should make that change in the `setup.cfg` and re-generate this file. Last time I did this I just ran `versioneer install` again, maybe they added an `update` option now.

Bobfrat: setup.cfg was updated in the last PR but I forgot to commit this file after running ```python versioneer.py install```

ocefpaf: This should be OK then.

PR #592, opened 2018-06-21
Counts number of issues in header instead of weighted score

benjwadams:


benjwadams:
Should probably add tests here as well to ensure the desired behavior is taking place.

benjwadams:
Ok, unit test has been added.

Issue #542, opened 2017-11-22
Checker crashes on files non-string standard_name


drf5n:
I had a file with this bad bit of CDL in it and the compliance checker crashed on the mis-configured standard_name with the integer type:

```
        ...
	float hgrid_face_x(n_hgrid_face) ;
		hgrid_face_x:long_name = "x_coordinate of 2D mesh face" ;
		hgrid_face_x:standard_name = 0 ;
		hgrid_face_x:units = "m" ;
		hgrid_face_x:mesh = "hgrid" ;
        ...

```

```
$ compliance-checker -v -t cf test.nc
Running Compliance Checker on the dataset from: test.nc
Traceback (most recent call last):
  File "/Users/drf/anaconda/envs/python3/bin/compliance-checker", line 11, in <module>
    load_entry_point('compliance-checker==3.0.4', 'console_scripts', 'compliance-checker')()
  File "/Users/drf/anaconda/envs/python3/bin/cchecker.py", line 73, in main
    args.format)
  File "/Users/drf/anaconda/envs/python3/lib/python3.5/site-packages/compliance_checker/runner.py", line 53, in run_checker
    *checker_names)
  File "/Users/drf/anaconda/envs/python3/lib/python3.5/site-packages/compliance_checker/suite.py", line 139, in run
    checker.setup(ds)
  File "/Users/drf/anaconda/envs/python3/lib/python3.5/site-packages/compliance_checker/cf/cf.py", line 107, in setup
    self._find_geophysical_vars(ds)
  File "/Users/drf/anaconda/envs/python3/lib/python3.5/site-packages/compliance_checker/cf/cf.py", line 289, in _find_geophysical_vars
    self._geophysical_vars[ds] = cfutil.get_geophysical_variables(ds)
  File "/Users/drf/anaconda/envs/python3/lib/python3.5/site-packages/compliance_checker/cfutil.py", line 274, in get_geophysical_variables
    if is_geophysical(ds, variable):
  File "/Users/drf/anaconda/envs/python3/lib/python3.5/site-packages/compliance_checker/cfutil.py", line 137, in is_geophysical
    if 'status_flag' in standard_name or hasattr(ncvar, 'flag_meanings'):
TypeError: argument of type 'numpy.int32' is not iterable
$
```
I realize the problem is within my file, (now fixed,) but perhaps the compliance checker should check types and report the error in the input rather than hard faulting.

Thanks

benjwadams:
[...]

PR #594, opened 2018-06-21
Valid string units and standard name

benjwadams:
Fixes a number of issues related to non-string units and standard_name attributes in variables.  Addresses #542

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -920,14 +920,11 @@ class CFBaseCheck(BaseCheck):
         :rtype: tuple
         :return: 2-tuple of standard_name and modifier as strings
         '''
-        standard_name_modifier = None
-        if not isinstance(standard_name, basestring):
-            return (None, None)
 
-        if ' ' in standard_name:
-            standard_name, standard_name_modifier = standard_name.split(' ', 1)
-
-        return (standard_name, standard_name_modifier)
+        if isinstance(standard_name, basestring) and ' ' in standard_name:
+            return standard_name.split(' ', 1)
+        else:
+            return standard_name, None
 
     def _check_valid_cf_units(self, ds, variable_name):
         '''
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -899,7 +899,7 @@ class CFBaseCheck(BaseCheck):
                                 "§3.1 Variable {}'s units attribute is a string".format(variable.name))
             # side effects, but better than teasing out the individual result
             if units_attr_is_string.assert_true(isinstance(units, basestring),
-                                                "'units' attribute must be a string"):
+                                                "'units' attribute must be a string compatible with UDUNITS"):
                 valid_udunits = self._check_valid_udunits(ds, name)
                 ret_val.append(valid_udunits)
             ret_val.append(units_attr_is_string.to_result())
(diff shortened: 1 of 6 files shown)


benjwadams:
This also prevents compliance checker from crashing outright if there are variables with non-string standard name.

@kwilcox, feel free to comment on implementation if you'd like.


Aside from unit tests updated here, also tested against the following .cdl:

```

netcdf bad_units_standard_name {
dimensions:
        n_hgrid_face = 2 ;
variables: 
    float hgrid_face_x(n_hgrid_face) ;
        hgrid_face_x:long_name = "x_coordinate of 2D mesh face" ;
        hgrid_face_x:standard_name = 0 ;
        hgrid_face_x:units = "m" ;
        hgrid_face_x:mesh = "hgrid" ;
    float test_var(n_hgrid_face) ;
        test_var:long_name = "Dummy variable for practical salinity" ;
        test_var:standard_name = "sea_water_practical_salinity" ;
        test_var:units = 1 ;
        test_var:mesh = "hgrid" ;
}
```

compliance_checker/cfutil.py
```diff
@@ -52,6 +54,46 @@
     'atmosphere_sleve_coordinate'
 ]
 
+def check_attr_membership(attr_val, value_set, attr_type=basestring,
```
Bobfrat: If this is just a helper function, I'd suggest changing the name away from the ```check_``` convention we've reserved for checks in the check classes.

benjwadams: Good point, will fix.

compliance_checker/cf/cf.py
```diff
@@ -895,9 +895,14 @@ def check_units(self, ds):
             valid_units = self._check_valid_cf_units(ds, name)
             ret_val.append(valid_units)
 
-            if isinstance(units, basestring):
+            units_attr_is_string = TestCtx(BaseCheck.MEDIUM,
+                                "§3.1 Variable {}'s units attribute is a string".format(variable.name))
+            # side effects, but better than teasing out the individual result
+            if units_attr_is_string.assert_true(isinstance(units, basestring),
+                                                "'units' attribute must be a string"):
```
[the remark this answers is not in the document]
benjwadams: Both should be checked, so I'm going to remove the conditional here and run both checks.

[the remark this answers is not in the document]
benjwadams: integer value of `1` is valid udunits according to cfunits:

```python
from cfunits import Unit
# does not throw exception
Unit(1)
```

compliance_checker/cfutil.py
```diff
@@ -52,12 +54,55 @@
...
+        is_in_set = modifier_fn(new_attr_val) in value_set
+    except Exception as e:
+        warnings.warn('Could not apply modifier function {} to value: '
+                      ' {}'.format(modifier_fn, e.msg))
+        return False
+
+    return is_in_set
+
 
 def get_unitless_standard_names(xml_tree, units):
```
Bobfrat: This is confusing, above in cf.py we're passing in standard_name as the second arg.

```
std_name_unitless = cfutil.get_unitless_standard_names(self._std_names._root, standard_name)
```

benjwadams: Given that the above is going to be implemented in https://github.com/ioos/compliance-checker/pull/597 instead, please approve this PR so we can merge and hopefully close out #542 or indicate where you may want further improvements.

PR #597, opened 2018-06-27
Updates names, check for dimensionless standard names

benjwadams:
Renames a number of functions and variables related to checks for units
when standard_name is determined to be dimensionless.  Also expands the
accepted UDUnits dimensionless units names checked.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -947,8 +947,9 @@ class CFBaseCheck(BaseCheck):
                                                                standard_name)
         # Is this even in the database? also, if there is no standard_name,
         # there's no way to know if it is dimensionless.
-        should_be_dimensionless = (variable.ndim == 0 or variable.dtype.char == 'S' or
-                                   std_name_units_dimensionless or standard_name is None)
+        should_be_dimensionless = (variable.dtype.char == 'S' or
+                                   std_name_units_dimensionless or
+                                   standard_name is None)
 
         # 1) Units must exist
         valid_units = TestCtx(BaseCheck.HIGH, '§3.1 Variable {} contains valid CF units'.format(variable_name))
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -942,7 +942,7 @@ class CFBaseCheck(BaseCheck):
         units = getattr(variable, 'units', None)
         standard_name_full = getattr(variable, 'standard_name', None)
         standard_name, standard_name_modifier = self._split_standard_name(standard_name_full)
-        std_name_units_dimensionless = cfutil.get_dimensionless_standard_names(self._std_names._root,
+        std_name_units_dimensionless = cfutil.is_dimensionless_standard_name(self._std_names._root,
                                                                standard_name)
         # Is this even in the database? also, if there is no standard_name,
         # there's no way to know if it is dimensionless.
@@ -979,7 +979,7 @@ class CFBaseCheck(BaseCheck):
         units = getattr(variable, 'units', None)
         standard_name = getattr(variable, 'standard_name', None)
         standard_name, standard_name_modifier = self._split_standard_name(standard_name)
-        std_name_units_dimensionless = cfutil.get_dimensionless_standard_names(self._std_names._root,
+        std_name_units_dimensionless = cfutil.is_dimensionless_standard_name(self._std_names._root,
                                                                standard_name)
 
         # If the variable is supposed to be dimensionless, it automatically passes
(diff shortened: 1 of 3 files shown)


compliance_checker/cfutil.py
```diff
@@ -53,15 +53,25 @@
 }
 
 
-def get_unitless_standard_names(xml_tree, units):
+def get_dimensionless_standard_names(xml_tree, standard_name):
```
Bobfrat: i'm confused as to why you changed all these to dimensionless. I thought it was just the units -> standard_name arg that needed fixing

benjwadams: The terminology "unitless" doesn't appear in the CF manual.  Furthermore what were "unitless" variables could sometimes have units of a constant, i.e. '1e-3' or '1', so it wasn't truly unitless in that sense either, but these checks do refer to "dimensionless" variables, as noted in http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#units

compliance_checker/cfutil.py
```diff
@@ -53,15 +53,25 @@
...
     '''
-    Returns True if the units are unitless. Unitless includes units that have
-    no units and units that are defined as '1'.
+    Returns True if the units for the associated standard name are unitless.
+    Unitless includes units that have no units and units that are defined as
+    constant units in the CF standard name table i.e. '1', or '1e-3'.
```
mhidas: Maybe use the term "dimensionless" here too, for consistency? "units that have no units" is not very informative.

compliance_checker/cfutil.py
```diff
@@ -53,15 +53,25 @@
...
+        return False
+    found_standard_name = xml_tree.find(".//entry[@id='{}']".format(standard_name))
     if found_standard_name is not None:
         canonical_units = found_standard_name.find('canonical_units')
-        return canonical_units is None or canonical_units.text == '1'
+        # so far, standard name XML table includes
+        # 1 and 1e-3 for unitless dims, but expanding to valid udunits
+        # prefixes to be on the safe side
+        # taken from CF Table 3.1 of valid UDUnits prefixes
+        dimless_units = r'1(?:e-?(?:1|2|3|6|9|12|15|18|21|24))$'
```
mhidas: This regex would not match "1". I think you want a zero-or-one quantifier '?' for the exponent, instead of a non-capturing group "(?:"?
```python
dimless_units = r'1(e-?(1|2|3|6|9|12|15|18|21|24))?$'
```
(https://regex101.com/ is always a good place to check)

compliance_checker/cfutil.py
benjwadams: @mhidas, thanks for the feedback, I think your points are pretty spot on here, so I'll be implementing your suggestions.

compliance_checker/cfutil.py
benjwadams: yeah, forgot the `?` here at the end.  I usually use non capturing groups if I'm not backreferencing.

compliance_checker/cf/cf.py
```diff
@@ -979,20 +979,20 @@ def _check_valid_udunits(self, ds, variable_name):
...
         standard_name, standard_name_modifier = self._split_standard_name(standard_name)
-        std_name_unitless = cfutil.get_unitless_standard_names(self._std_names._root,
+        std_name_units_dimensionless = cfutil.get_dimensionless_standard_names(self._std_names._root,
                                                                standard_name)
 
-        # If the variable is supposed to be unitless, it automatically passes
-        should_be_unitless = (
+        # If the variable is supposed to be dimensionless, it automatically passes
+        should_be_dimensionless = (
             variable.ndim == 0 or
```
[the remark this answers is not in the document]
benjwadams: @mhidas, this is the one remaining point to be addressed.  Please comment on suggested variable nomenclature here, along with any suggestions you may have.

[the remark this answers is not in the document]
mhidas: The name `should_be_dimensionless` is fine. I forgot to actually state this (sorry), but what I was trying to say is that you should remove `variable.ndim == 0` here. It is not a condition that exempts `variable` from having a correct units attribute (it refers to the structure of the netCDF variable, not the physical quantity it represents).

(Apologies for delayed response - I don't work on Fridays.)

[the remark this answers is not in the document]
benjwadams: Done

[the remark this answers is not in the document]
mhidas: Looks like this needs to be done in two places. I can see you've removed `variable.ndim == 0` on line [950](https://github.com/ioos/compliance-checker/blob/d9977f8905f6b329aa8e2d1872c97d58a3ea3cf8/compliance_checker/cf/cf.py#L950), but it's still here, on line [989](https://github.com/ioos/compliance-checker/blob/d9977f8905f6b329aa8e2d1872c97d58a3ea3cf8/compliance_checker/cf/cf.py#L989)

Issue #587, opened 2018-05-22
Include link to standards in report


Bobfrat:

benjwadams:
Merged in by #598

PR #598, opened 2018-06-29
Metadata Reference Links

daltonkell:
Added logic to include the metadata reference links for each standard in the output for reports (all formats). References #587


compliance_checker/data/templates/ccheck.html.j2
```diff
@@ -8,6 +8,7 @@
   </div>
   <div class="col-md-12">
     <h3>{{testname}}</h3>
+    <a href={{cc_url}}>Reference Page</a>
```
Bobfrat: Can you make the test name in the line above be the hyperlink instead? And have it open in a new tab?

compliance_checker/suite.py
```diff
@@ -358,13 +367,16 @@ def standard_output(self, ds, limit, check_name, groups):
 
         # Let's add the version number to the check name if it's missing
         check_name = self._get_check_versioned_name(check_name)
+        check_url = self._get_check_url(check_name)
+        _len_ = 120
```
mhidas: Could this be set to `2 * self.col_width` instead?
And could all the other hard-coded numbers below be replaced with a reference to this variable?
(Then the column width could even be an optional command-line argument, but that's for another PR...)

daltonkell: @mhidas I made the change to line 371, but I'm quite certain the string alignment in the following lines has to be specified in each string as dictated in 

https://docs.python.org/3.4/library/string.html#formatstrings

compliance_checker/data/templates/ccheck.html.j2
daltonkell: Done, see the latest commit

compliance_checker/suite.py
mhidas: The documentation doesn't elaborate on this possibility, but there are examples that show you can specify any of the formatting fields from variables (see last section under https://docs.python.org/3.4/library/string.html#format-examples). So you could do e.g.
```python
        print('{:^{width}}'.format("IOOS Compliance Checker Report", width=_len_))
        print('{:^{width}}'.format(check_name + ' check', width=_len_))
        print('{:^{width}}'.format(check_url, width=_len_))
```

daltonkell: :man_facepalming: How did I miss those... I'll get those changes in ASAP.

benjwadams:
Uh oh, build failures on Travis.   FWIW, looks unrelated to the current PR:

    ImportError: libiconv.so.2: cannot open shared object file: No such file or directory

compliance_checker/suite.py
```diff
@@ -425,8 +437,8 @@ def process_table(res, check):
                 # datasets
                 if _top_level:
                     print("\n")
-                    print('{:^80}'.format(level_name))
-                    print("-" * 80)
+                    print('{:^120}'.format(level_name))
```
Bobfrat: Should this be replaced by ```_len_ ``` like the others? Interesting variable name btw, i would've went with ```width```

PR #603, opened 2018-08-22
Allow StandardNameTable to optionally load CF standard name table fro…

lwgordonimos:
…m a local path defined by an environment variable

This PR is mostly to gauge whether you are interested in supporting alternate means of loading the CF standard name table, in addition to the current mechanisms.

The use case for this in particular is that we install compliance_checker via configuration management, and would like to de-couple the management of the XML from the compliance-checker package, so that package updates don't require re-downloading the file when we have already determined what we consider to be our internally supported version.

It would be nice to able to manage the XML outside the package's data directory as part of the deployed environment, hence the environment variable idea.


Bobfrat:
This looks good to me, I'll add a section to the README so others know about this new capability

Issue #602, opened 2018-08-15
Release 4.1.0 Proposal


daltonkell:
After merging PR #600, I think it's appropriate for a new minor release to 4.1.0 with the following description:
# Compliance Checker 4.1.0
### Functionality
Pull Request #600 enhances compliance checker speed with the use of an lru-cache (Least Recently Used) from `functools`. It also allows backwards-compatibility with the lru-cache for python2.7 via the [`functools32`](https://pypi.org/project/functools32/) backport module. 

PR #599 updates the Climate and Forecasting integration tests.

PR #598 puts reference links to metadata schemas into the compliance-checker output.

### Bug Fixes
Fixes the TravisCI build failures in `conda` environments found which initially caused #598 and #599 to fail.

---
@'ing @Bobfrat and @benjwadams , but if anyone has input on this it would be great to hear

benjwadams:
Overall looks good, but should fix the error message which occurs when the memoized subclass of `netCDF4.Dataset` is deleted.

mhidas:
Could you please also include PR #603 in this release?

Bobfrat:
@mhidas yes, everything in master to date will get in this next release.

benjwadams:
Could you create a milestone instead?

Bobfrat:
New release is on PyPi
https://pypi.org/project/compliance-checker/

PR #604, opened 2018-08-28
Dealloc ds error fix

benjwadams:
Fixes error messages when deallocating MemoizedDataset

### DIFF
--- a/compliance_checker/runner.py
+++ b/compliance_checker/runner.py
@@ -66,6 +66,9 @@ class ComplianceChecker(object):
 
             score_groups = cs.run(ds, [] if skip_checks is None else skip_checks,
                                 *checker_names)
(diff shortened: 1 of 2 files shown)


benjwadams:
Related to #602

compliance_checker/runner.py
```diff
@@ -66,6 +66,9 @@ def run_checker(cls, ds_loc, checker_names, verbose, criteria,
 
             score_groups = cs.run(ds, [] if skip_checks is None else skip_checks,
                                 *checker_names)
+            # TODO: consider wrapping in a proper context manager instead
```
Bobfrat: 👍 thats not a bad idea! Can you make it happen? Perhaps open an issue so we don't lose track of it.

PR #605, opened 2018-08-29
remove cython

ocefpaf:
I don't know why `cython` was added but it is not used to build any code in compliance-checker and it is rarely a run time dependency. If we ever use `cython` in the future we must make it a `build` dependency only and never `run`.


ocefpaf:
@Bobfrat this is a show stopper to package `compliance-checker` do you mind taking a look and issuing a new point release if this is OK to you.

ocefpaf:
Done #606

Issue #579, opened 2018-05-15
PyPI install broken due to NetCDF4 1.4.0 package


lwgordonimos:
Something has changed in "NetCDF==1.4.0" which causes a basic "out of the box" install of compliance-checker from PyPI to fail due to the "netcdftime" module. Since the package is not pinned with a maximum version, any new installs of "compliance-checker" will be broken by default.

Not sure which is the most appropriate solution for this package, but I've included two actions which resolve the issue for me.

## Steps to replicate:

```bash
$ virtualenv compliance-checker
$ source compliance-checker/bin/activate
$ pip install compliance-checker

$ pip freeze
certifi==2018.4.16
cf-units==1.2.0
cftime==1.0.0b1
chardet==3.0.4
compliance-checker==4.0.0
Cython==0.28.2
idna==2.6
isodate==0.6.0
Jinja2==2.10
lxml==4.2.1
MarkupSafe==1.0
netCDF4==1.4.0
numpy==1.14.3
OWSLib==0.16.0
pendulum==2.0.1
pygeoif==0.7
pyproj==1.9.5.1
python-dateutil==2.7.3
pytz==2018.4
pytzdata==2018.5
regex==2017.7.28
requests==2.18.4
six==1.11.0
tabulate==0.8.2
typing==3.6.4
urllib3==1.22

$ compliance-checker -l
Traceback (most recent call last):
  File "/home/lwgordon/virtualenvs/compliance-checker/bin/compliance-checker", line 7, in <module>
    from cchecker import main
  File "/home/lwgordon/virtualenvs/compliance-checker/bin/cchecker.py", line 7, in <module>
    from compliance_checker.cf.util import download_cf_standard_name_table
  File "/home/lwgordon/virtualenvs/compliance-checker/local/lib/python2.7/site-packages/compliance_checker/cf/__init__.py", line 1, in <module>
    from compliance_checker.cf.cf import (
  File "/home/lwgordon/virtualenvs/compliance-checker/local/lib/python2.7/site-packages/compliance_checker/cf/cf.py", line 8, in <module>
    from compliance_checker.cf import util
  File "/home/lwgordon/virtualenvs/compliance-checker/local/lib/python2.7/site-packages/compliance_checker/cf/util.py", line 10, in <module>
    from cf_units import Unit
  File "/home/lwgordon/virtualenvs/compliance-checker/local/lib/python2.7/site-packages/cf_units/__init__.py", line 41, in <module>
    import netcdftime
[...]

PR #606, opened 2018-08-30
Unify req files

ocefpaf:
@daf and @Bobfrat depending on `cython` as a run time dependency is problematic and unnecessary here. After digging about this for a while it turns out that `pkg_resources` was responsible for the bogus dependencies requirements. I turned `.load(require=False)` to allow the package managers, `pip`, `conda`, etc do that for us. This also allows for some free testing of tightly pinned packages without jumping too many hoops.

I also cut a lot some of the fat in all those requirement files, that could cause problems in our dependency tracking and it was getting out of hand. Again, we have only two: `requirements.txt` and `test_requirements.txt`. Travis-CI and AppVeyor where updated to use them. (I'll probably need to adjust AppVeyor a little bit but it is late and I need to :zzz:)


daf:
I'm happy with the consolidation of the requirements + tooling around them.  It's been a long time since I wrote the entrypoint stuff so I don't quite remember the ins and outs of `load()`, but at a slightly-longer-than-first glance it seems ok.  I'll have a chance to verify it further in the morning.

daf:
Ok, I've done more digging, and here's what I've found.  It looks like `cf_units` is the real cause of cython being required - I noticed it first when attempting to get cc-plugin-glider into the env I created for testing this PR (doing a `pip install -e .`):

```
Collecting cython (from cf_units>=2->compliance-checker>=1.0.0->cc-plugin-glider==1.0.5)
  Using cached https://files.pythonhosted.org/packages/99/f9/1c93f3ef9189e53c8647b55ea60ad5f6fd4cdcb0217fc97e87e299d8528d/Cython-0.28.5-cp35-cp35m-manylinux1_x86_64.whl
```

Following up on that, I created a completely fresh Python 3.5 conda env, and did `conda install cc-plugin-glider` to see what was currently available from conda-forge:

```
~/dev/compliance-checker -> conda install cc-plugin-glider                                                                                                                                                       [unify_req_files]&cc-ep
Solving environment: done

## Package Plan ##

  environment location: /home/daf/miniconda3/envs/cc-ep

  added / updated specs: 
    - cc-plugin-glider


The following packages will be downloaded:

    package                    |            build
    ---------------------------|-----------------
    compliance-checker-4.0.0   |           py35_0         603 KB  conda-forge
    cc-plugin-glider-1.0.5     |           py35_2          22 KB  conda-forge
    ------------------------------------------------------------
                                           Total:         625 KB

The following NEW packages will be INSTALLED:

    asn1crypto:           0.24.0-py35_3          conda-forge
    cc-plugin-glider:     1.0.5-py35_2           conda-forge
    cf_units:             2.0.1-py35h7eb728f_2   conda-forge
    cffi:                 1.11.5-py35h5e8e0c9_1  conda-forge
    cftime:               1.0.1-py35h7eb728f_0   conda-forge
    chardet:              3.0.4-py35_3           conda-forge
    compliance-checker:   4.0.0-py35_0           conda-forge
    cryptography:         2.3.1-py35hdffb7b8_0   conda-forge
    cryptography-vectors: 2.3.1-py35_0           conda-forge
    curl:                 7.61.0-h93b3f91_1      conda-forge
    expat:                2.2.5-hfc679d8_1       conda-forge
    hdf4:                 4.2.13-0               conda-forge
    hdf5:                 1.10.2-hc401514_1      conda-forge
    icu:                  58.2-hfc679d8_0        conda-forge
    idna:                 2.7-py35_2             conda-forge
    isodate:              0.6.0-py_1             conda-forge
    jinja2:               2.10-py_1              conda-forge
    jpeg:                 9c-h470a237_0          conda-forge
    krb5:                 1.14.6-0               conda-forge
    libgfortran:          3.0.0-1                conda-forge
    libgfortran-ng:       7.2.0-hdf63c60_3       conda-forge
    libiconv:             1.15-h470a237_2        conda-forge
    libnetcdf:            4.6.1-he6cff42_7       conda-forge
    libopenblas:          0.2.20-h9ac9557_7                 
    libssh2:              1.8.0-h5b517e9_2       conda-forge
    libxml2:              2.9.8-h422b904_3       conda-forge
    libxslt:              1.1.32-h88dbc4e_1      conda-forge
    lxml:                 4.2.4-py35hc9114bc_0   conda-forge
    markupsafe:           1.0-py35h470a237_1     conda-forge
    netcdf4:              1.4.1-py35h62672b6_0   conda-forge
    numpy:                1.14.3-py35h28100ab_1             
    numpy-base:           1.14.3-py35h0ea5e3f_1             
    owslib:               0.16.0-py_1            conda-forge
    pendulum:             2.0.3-py35_0           conda-forge
    proj4:                5.1.0-h470a237_1       conda-forge
    pycparser:            2.18-py_1              conda-forge
    pygeoif:              0.7-py_1               conda-forge
    pyopenssl:            18.0.0-py35_0          conda-forge
    pyproj:               1.9.5.1-py35h58270c7_2 conda-forge
    pysocks:              1.6.8-py35_2           conda-forge
    python-dateutil:      2.7.3-py_0             conda-forge
    pytz:                 2018.5-py_0            conda-forge
    pytzdata:             2018.5-py_0            conda-forge
    regex:                2017.07.28-py35_0      conda-forge
    requests:             2.19.1-py35_1          conda-forge
    six:                  1.11.0-py35_1          conda-forge
    tabulate:             0.8.2-py_0             conda-forge
    typing:               3.6.6-py35_0           conda-forge
    tzlocal:              1.5.1-py_0             conda-forge
    udunits2:             2.2.27.6-h3a4f0e9_0    conda-forge
    urllib3:              1.23-py35_1            conda-forge

Proceed ([y]/n)?  


Downloading and Extracting Packages
compliance-checker-4 | 603 KB    | ############################################################################################################################################################################################## | 100% 
cc-plugin-glider-1.0 | 22 KB     | ############################################################################################################################################################################################## | 100% 
Preparing transaction: done
Verifying transaction: done
Executing transaction: done
```

No `cython`, ok, cool.  However, running it:

```
~/dev/compliance-checker -> compliance-checker -l                                                                                                                                                                [unify_req_files]&cc-ep
Could not load acdd-1.3 = compliance_checker.acdd:ACDD1_3Check : The 'cython' distribution was not found and is required by cf-units
Could not load ioos-0.1 = compliance_checker.ioos:IOOS0_1Check : The 'cython' distribution was not found and is required by cf-units
Could not load acdd-1.1 = compliance_checker.acdd:ACDD1_1Check : The 'cython' distribution was not found and is required by cf-units
Could not load cf = compliance_checker.cf.cf:CFBaseCheck : The 'cython' distribution was not found and is required by cf-units
Could not load ioos-1.1 = compliance_checker.ioos:IOOS1_1Check : The 'cython' distribution was not found and is required by cf-units
Could not load ioos_sos = compliance_checker.ioos:IOOSBaseSOSCheck : The 'cython' distribution was not found and is required by cf-units
Could not load gliderdac = cc_plugin_glider.glider_dac:GliderCheck : The 'cython' distribution was not found and is required by cf-units
IOOS compliance checker available checker suites:
```

further lending weight to the idea that it's not `pkg_resources`, but instead `cf_units`.

I still like the requirements cleanup of this PR though.

ocefpaf:
Yes, but `cf_units` does use cython to build! Not to run. `pkg_resources` is overstepping its functionality IMO.

We could jump some hoops and use a `setup_requires` in cf-units but IMO pkg_resources is just broken when trying to enforce dependencies of dependencies like that.

daf:
Ok, I'm starting to understand - I was trying to grok the [conda recipe metadata host](https://conda.io/docs/user-guide/tasks/build-packages/define-metadata.html#host) docs but obviously missed how Host stuff works.

I'll continue to try to get this PR to work locally.

daf:
(there are at least one more `load()` calls that also need to be addressed.)

ocefpaf:
> I was trying to grok the conda recipe metadata host docs but obviously missed how Host stuff works.

Even in a non-conda scenario that issue would still exist. My main issue with that is not the `cython` dependency per-se, but the overly strict policy on running compliance checker with only the pinned packages using by its dependencies. Imagine if I want to try it with the latest `regex`? I  would have to re-build `compliance-checker` for that. With this PR we let the package managers take care of that and, if someone wants to break the dependencies pins, they can with the proper package manager options to over those.

Bobfrat:
@ocefpaf FWIW, I think we started to pin packages to mitigate issues like this from cropping up:
https://github.com/ioos/compliance-checker/issues/579

ocefpaf:
> @ocefpaf FWIW, I think we started to pin packages to mitigate issues like this from cropping up:
#579

Some pinning, when there are known incompatibilities, is OK. Pinning everything strictly is bad practice IMO. The case you refer to is an example where I'd rather not have a pin, and address breakages when they occur, rather than having a strict pin and prevent users from using latest packages in their envs.

(I'm the one that has to explain conda users why they cannot have `netcdf4 1.4` when installing `compliance-checker` :wink:)

PS0: note that strict pinning for deployment is another thing. I would definitely be very strict about the versions that goes in the web version of compliance-checker app.

PS1: the bogus `cython` dependency is an artifact of the strict pinning b/c it is not really need at run time.

PS3: when using `require=False` in `pkg_resources` we are not breaking the pin in the requirement files! We are only letting the package manager handle that and avoiding `pkg_resources` from overstepping its functionalities.

gen_conda_requirements.py
```diff
@@ -0,0 +1,90 @@
...
+REQS_DIR = os.path.dirname(__file__)
+CONDA_PATTERN = '#conda:'
+
+
+def read_conda_reqs(fname, options):
+    lines = []
+    with open(fname, 'r') as fh:
+        for line in fh:
+            line = line.strip()
+            if '#conda:' in line:
```
daf: I know you didn't write this but y "u" no `CONDA_PATTERN`

daf:
Testing locally, this seems correct.  I'm getting a Deprecation notice about `Parameters to load are deprecated.  Call .resolve and .require separately`.

Looking at [EntryPoint docs](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#entrypoint-objects), seems like we could just use `.resolve()` instead of `.load(require=False)`.

gen_conda_requirements.py
ocefpaf: Sorry. I'm not sure I follow your comment.
(And yes, I did not write it, shameless stolen from `iris` :smile:)

ocefpaf:
> Looking at EntryPoint docs, seems like we could just use .resolve() instead of .load(require=False).

Let me try that here. Hold on...

daf:
:+1:  I tested this by adding an uninstalled import in my local copy of cc-plugin-glider (pandas), so when I try to list checkers, it gives me this:

```
/dev/compliance-checker -> python cchecker.py -l -v                                                                                              [unify_req_files*]&cc-unify-reqs
Could not load gliderdac = cc_plugin_glider.glider_dac:GliderCheck : No module named 'pandas'
IOOS compliance checker available checker suites:
 - acdd (v4.1.0+7.g39c3e0c.dirty)
...
```

This seems correct.

gen_conda_requirements.py
daf: The following line gets the index/length of `CONDA_PATTERN`, and up above `CONDA_PATTERN = '#conda:'`, but for some reason it re-uses the bare string on this line.

daf:
Related @ocefpaf, is there a `pip install -e .` equivalent for in-development packages like cc-plugin-glider that won't install Cython?  A conda equivalent I guess?

gen_conda_requirements.py
ocefpaf: Ah OK. I did not see the line below until I click `/files` on GitHub.
I can fix that, this script was probably not designed to last. I would love to get rid of it as soon as conda start supporting the new req file standards.

Bobfrat:
@daf feel free to merge if this is good to go. I can make a new release

daf:
@Bobfrat 

> Only those with write access to this repository can merge pull requests.

:man_shrugging:

Issue #607, opened 2018-08-30
Running compliance-checker with no arguments results in 


daf:

daf:
It should act just like `--help` was specified.  I'll make a PR.

PR #608, opened 2018-08-30
Show help when no operations are requested on the command line, updates #607

daf:
This does not technically "fix" #607, as there are still ways to call that method and get the UnboundLocalError esp when using CC programatically, but this makes it behave like a command line tool is expected to.

### DIFF
--- a/cchecker.py
+++ b/cchecker.py
@@ -94,7 +94,7 @@ def main():
 
     if len(args.dataset_location) == 0:
         parser.print_help()
-        return 0
+        return 1
 
     # Check the number of output files
     if not args.output:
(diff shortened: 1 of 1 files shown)


cchecker.py
```diff
@@ -92,6 +92,10 @@ def main():
     if args.download_standard_names:
         download_cf_standard_name_table(args.download_standard_names)
 
+    if len(args.dataset_location) == 0:
+        parser.print_help()
+        return 0
```
ocefpaf: :+1: I just wonder if the return code should really be `0`. In theory no args is an error and printing the `help` without the user requesting it is a "hint" and not an expected result.

daf:
I set it at 1, [according to LDP](http://tldp.org/LDP/abs/html/exitcodes.html) there's nothing more specific that applies.

PR #611, opened 2018-08-30
Use argparser's choices param for format validation

daf:


Bobfrat:
@daf i think this just needs a rebase on master to get through travis

daf:
@Bobfrat done and CI seems happy.

PR #614, opened 2018-10-22
added class variable as an option to alternative html templates

giumas:


compliance_checker/suite.py
```diff
@@ -40,6 +40,7 @@
 class CheckSuite(object):
 
     checkers = {}       # Base dict of checker names to BaseCheck derived types, override this in your CheckSuite implementation
+    templates_root = 'compliance_checker'  # modify to load alternative Jinja2 templates
```
Bobfrat: @giumas do you have some alternative Jinja2 templates? I'd be interested in improving the HTML output.

giumas: I don't, but I will likely have. 

My intent is to write a checker for a format similar to the [SONAR-netCDF4 format](http://www.ices.dk/sites/pub/Publication%20Reports/Cooperative%20Research%20Report%20(CRR)/CRR341/CRR341.pdf). Thus, having a way to modify the path where to read for the templates would be great!

giumas: This is how I can then use the variable: https://github.com/giumas/xsf-checker/blob/master/xsf_checker/cli/main.py#L11

Bobfrat: very cool. Not sure why appveyor failed. @ocefpaf are you able to kick over a new build?

PR #622, opened 2019-01-23
Goal: reformat ACDD checks so they are collapsed into compact groups,…

daltonkell:
… similar to the CF reformatting.

Through the changes described below, we go from output like this:
```
                                  Recommended                                   
acknowledgment/acknowledgement
* Attr acknowledgement not present

comment
* Attr comment not present

creator_email
* Attr creator_email not present

creator_name
* Attr creator_name not present

creator_url
* Attr creator_url not present
```

to output like this:
```
                                  Recommended                                   
Global Attributes
* acknowledgment/acknowledgement not present
* id not present
* naming_authority not present
* history not present
* comment not present
* date_created not present
* creator_name not present
* creator_url not present

```

acdd.py:
  - now using a common variable header, "self._var_header" to be the Result
    name for individual variables being checked
  - Global Attributes:
    - Now, the global attribute Results are
      each created without a name argument, allowing grouping once the
      reasoning routine is performed

  - check_acknowledgment:
    - name argument is removed from the Result so it can be grouped
      with the Global Attributes.

  - verify_convention_version:
    - Wraps some of the code in a try-except clause to properly access the
      variable, and if not there, yield the appropriate message

base.py:
  - class Result:
    - added a short block to address the to take length-0 tuple and make a
      length-2 tuple of (0, 0).

  - attr_check:
    - removed redundant occurances of "Attr" in the messages assigned
    - removing the names from Result creation for better grouping,

suite.py
  - class CheckSuite
    - inserted a block which separates the global attribute Results by
    priority weight, and then appends the Results to the main list as is
    convention in the code.

test_acdd.py
  - class testACDD1_1
    - because the Global Attribute Results now don't have name arguments,
      all of the ncei_exceptions must now be tested by the message strings
      rather than the names. This is applied throughout this unit test file.
    - test_acknowledgement_check
      Removed because the "acknowledgement" is part of the Global Attributes.
      - this took place in the code twice

test_base.py
  - changed string comparisons to reflect the newly updated messages
  - removed names from any Global Attribute Result creation so the tests
    would be accurate
  - updated test_attr_fn to better reflect the current attr_check function


compliance_checker/tests/test_acdd.py
```diff
@@ -395,20 +373,6 @@ def test_variables(self):
         for result in results:
             self.assert_result_is_bad(result)
 
-    def test_acknowledgement(self):
```
daltonkell: I removed this method since the "acknowledgement" attribute is a global attribute. While it does have its own method implemented in `acdd.py`, the method only serves to check if either/both the two possible spelling variations exist in the dataset ("acknowledgment" vs "acknowledg __e__ ment" with the added "e"). What say you, @Bobfrat @benjwadams?

compliance_checker/acdd.py
```diff
@@ -567,7 +574,11 @@ def __init__(self):
             ('Conventions', self.verify_convention_version)
         ])
 
-        self.rec_atts.extend(['geospatial_vertical_positive',
+        self.rec_atts.extend([
+                              #'acknowledgement',
```
daltonkell: Spotted these commented entries... I'll get rid of these

compliance_checker/acdd.py
```diff
@@ -516,19 +521,21 @@ def verify_convention_version(self, ds):
...
-        ]
-        return ratable_result((1, 2),
-                              'Conventions',
-                              messages)
+        try:
+            for convention in ds.getncattr("Conventions").replace(' ', '').split(','):
+                if convention == 'ACDD-' + self._cc_spec_version:
+                    return ratable_result((2, 2), None, []) # name=None so grouped with Globals
+
+                else:
```
daltonkell: This `else... return` statement should actually be outside of the `for` loop, as datasets can have multiple conventions.

compliance_checker/tests/test_acdd.py
Bobfrat: gotcha, my bad, we're in the tests not the checks. I was looking at line 46 and got confused:
```
"acknowledgment/acknowledgement",
```

compliance_checker/acdd.py
```diff
@@ -526,15 +526,12 @@ def verify_convention_version(self, ds):
...
 
-                else:
-                    # Conventions attribute is present, but does not include
-                    # proper ACDD version
-                    messages = [
-                    "\"Conventions\" does not contain 'ACDD-{}'".format(self._cc_spec_version)
-                    ]
-                    return ratable_result((1, 2), None, messages)
+            # if no/wrong ACDD convention, return appropriate result
+            return ratable_result((0, 2), None, messages)
```
daltonkell: Since it would have had a "Conventions" attribute, just not the right one, it should get a `(1, 2)` score

compliance_checker/base.py
```diff
@@ -325,12 +339,14 @@ def fix_return_value(v, method_name, method=None, checker=None):
...
     """
     # remove common check prefix
-    method_name = (method_name or method.__func__.__name__).replace("check_",
-                                                                    "")
+    method_name = (method_name or method.__func__.__name__).replace("check_","")
     if v is None or not isinstance(v, Result):
         v = Result(value=v, name=method_name)
 
     v.name         = v.name or method_name
+    if v.name in ["high", "recommended", "suggested"]:
```
daltonkell: Need to find a way to move this out of `base.py`. Can we somehow leverage the capability of calling a check using the 
`("name", bound_method)` model?

daltonkell:
@ocefpaf This AppVeyor build failed with a strange code:
```
Executing transaction: ...working... done
#
# To activate this environment, use
#
#     $ conda activate TEST
#
# To deactivate an active environment, use
#
#     $ conda deactivate
conda activate TEST
conda.exe info --all
'conda.exe' is not recognized as an internal or external command,
operable program or batch file.
Command exited with code 1
```
Any thoughts on this? Is it possible to fix by simply re-running the build?

ocefpaf:
> @ocefpaf This AppVeyor build failed with a strange code:

@daltonkell updating conda inside AppVeyor is broken and latest conda has a number of issues, see https://github.com/conda/conda/issues/8155

I'll try to pin to a working version. PR soon.

PR #624, opened 2019-02-01
Goal: collapse ACDD checks into compact groups

daltonkell:
**NOTE: This is a rebased version of the commits in the "acdd_reformat" branch;
rebasing allows for a cleaner, more continuous commmit history. It's also rebased
on master to incorporate the temporary conda bug fix for the AppVeyor tests.**

acdd.py:
  - class ACDDBaseCheck

    - def check_high()
      - changed @check_has decorator to include the kwarg
        "gname=Global Attributes"

    - def check_recommended()
      - changed @check_has decorator to include the kwarg
        "gname=Global Attributes"

    - def check_suggested()
      - changed @check_has decorator to include the kwarg
        "gname=Global Attributes"

    Anywhere Result objects used for Global Attributes now use the name
    "Global Attributes" so they are grouped together (e.g. the Conventions
    attr, the acknowledgment/acknowledgement attr).

  - verify_convention_version:
    - Wraps some of the code in a try-except clause to properly access the
      variable, and if not there, yield the appropriate message

base.py:
  - class Result:
    - added a short block to address the to take length-0 tuple and make a
      length-2 tuple of (0, 0).

  - attr_check:
    - removed redundant occurances of "Attr" in the messages assigned

suite.py
  - class CheckSuite
    - leveraged the _group_raw() function's call to itertools.groupby to
      sort and group the Result objects both by name and then priority to
      enable grouping Global Attributes (and any other Results with the same name)
      together; this produces much cleaner and more compact outputs

test_acdd.py
  - updated string comparisons as needed;

test_base.py
  - changed string comparisons to reflect the newly updated messages
  - updated test_attr_fn to better reflect the current attr_check function

********************************************************
********** below are squashed commit messages **********
********************************************************

Grouping Cleanup

While retaining the same functionality of the previous commit, replaced
loops with far more efficient sorting methods.

acdd.py
  - the previous commit had a logic error in the conventions check where
    the function would return in the middle of a loop. This did not cause
    any failures, but would have yielded potentially incorrect results.
  - removed commented code that slipped through the cracks

runner.py
  - added a couple comments for clarity

suite.py
  - standard_output_generation()
     - removed unused arguments being passed into reasoning_routine
  - reasoning_routine()
     - docstring
     - removed unused parameters from function definition
     - added comments for clarity

  - _group_raw()
     - removed unused args from reasoning_routing recursive call
     - added comments
     - removed unused internal function build_group
     - added return value to group_func internal function to improve the
       sorting and grouping
     - removed the functional but inefficient (and hardcoded) looping
       routine used to group the Global Attributes

test_acdd.py
  - TestACDD1_1
      - test_highly_recommended
          - added a "missing" field
          - test now loops through messages in the Results to check
  - TestACDD1_3
      - test_highly_recommended
           - this dataset did not have a missing Conventions attr, so this
             was removed

test_base.py
  - removed commented "else"

Further ACDD Reformatting

Removes "hardcoding" the Global Attributes messages by leveraging the
existing @check_has decorator and passing in another keyword argument.

acdd.py
  - class ACDDBaseCheck

    - def check_high()
      - changed @check_has decorator to include the kwarg
        "gname=Global Attributes"

    - def check_recommended()
      - changed @check_has decorator to include the kwarg
        "gname=Global Attributes"

    - def check_suggested()
      - changed @check_has decorator to include the kwarg
        "gname=Global Attributes"

    Anywhere I had previously eliminated the Result name so that the
    Results would group together, I replaced "None" with "Global Attributes".

    - def check_conventions()
      - if some type of Conventions attr is present, the score is (1, 2) not
        (0, 2)
      - Result name now uses "Global Attributes"

base.py
  - class BaseCheck

    - def attr_check()
      - Defined a new keyword parameter, "gname"; if supplied,
        will name the Result gname instead of the function name.
        This allows Global Attributes to be grouped together
        without having to later change their names in a hardcoded fashion.

    - def check_has()
      - includes new parameter "gname" which is passed to attr_check

    - def fix_return_value()
      - removed hardcoding of changing names from "high" "recommended" or
        "suggested" to "Global Attributes"

    Added some minor commenting and doc-strings.

test_base.py
  - any tests which had previously removed the names from Result objects
    now use the original name
  - updated tests that are comparing strings to use the correct strings

test_cf.py
  - test_check_packed_data
    - was comparing the wrong bool value for a particular assertion

test_cf_integration.py
  - updated string comparisons

It should also be noted that I re-generated `bad_data_type.nc` specifying
the netCDF4 file format so the tests could accurately check the data type. The
previous dataset did not have a failing data type (int64) for the variable
`temp`, even though the .cdl file specified this.

re-add acknowledgement test

Re-added the test to check acknowledgement attribute

Additionally, even though the acknowledgment/acknowledgement attribute is a Global Attribute, it deserves its own check because of the dual-spelling; because of this, it is kept *out* of the lists of Global Attributes but will still group with them.


compliance_checker/base.py
```diff
@@ -123,7 +123,11 @@ def __init__(self,
         if value is None:
             self.value = None
         elif isinstance(value, tuple):
-            assert len(value) == 2, 'Result value must be 2-tuple or boolean!'
+            try:
+                assert len(value) == 2, 'Result value must be 2-tuple or boolean!'
+            except AssertionError:
+                if value == (): # just make it (0, 0)
+                    value = (0, 0)
```
Bobfrat: When does an empty tuple get passed in? Was this something new that was introduced in this PR or is this fixing something that had always been an issue?

compliance_checker/tests/test_cf_integration.py
```diff
@@ -323,9 +323,9 @@ def test_pr_inundation(self):
             u'§2.6.2 depth:comment should be a non-empty string', 
             u'§2.6.2 institution global attribute should be a non-empty string', 
             u'§2.6.2 comment global attribute should be a non-empty string', 
-            u"'units' attribute of 'LayerInterf' must be a string compatible with UDUNITS", 
-            u"'units' attribute of 'time_bounds' must be a string compatible with UDUNITS", 
-            u"'units' attribute of 'Layer' must be a string compatible with UDUNITS",
```
Bobfrat: let's keep this text, "compatible with UDUNITS". It tells the user a little extra so hopefully they can fix the units in one shot.

PR #625, opened 2019-02-05
Cf1.7 updates

daltonkell:
First draft of the CF 1.7 updates. **Not to be merged before** #624! Tests will 99.999% likely all blow up.

@benjwadams, @Bobfrat Looking forward to the review. The latest commit message has all the relevant detail.


daltonkell:
This is hilarious that the tests passed.

compliance_checker/tests/test_cf.py
```diff
@@ -1010,3 +1014,197 @@ def test_units_temporal(self):
...
+        # if variable values are equal, actual_range should not exist
+        dataset = MockTimeSeries()
+        dataset.createVariable('a', 'd', ('time',)) # dtype=double, dims=time
+        dataset.variables["a"][0:500] = 0 # set all 500 vals to 0
+        dataset.variables["a"].setncattr("actual_range", [1, 1]) # shouldn't exist
+        result = self.cf.check_actual_range(dataset)
+        score, out_of, messages = self.get_results(result)
+        assert score < out_of
+        assert len(messages) == 1
+        assert messages[0] == u"\"a\"'s values are all equal; actual_range shouldn't exist"
```
benjwadams: Python 3 uses Unicode strings by default, which are in UTF-8.  UTF-8 as an encoding has the property that the ASCII character set is defined in one byte, and is so is equivalent to the first 128 characters in ASCII.  Thus I don't think the `u` isn't strictly necessary as the bytes str (Python 2) and unicode str (Python 3) should have the same representation in terms of the byte content.

compliance_checker/tests/test_cf.py
```diff
@@ -1010,3 +1014,197 @@ def test_units_temporal(self):
...
+
+        with MockTimeSeries() as dataset:
+            dataset.createVariable('PS', 'd', ('time',)) # dtype=double, dims=time
+            dataset.variables["PS"].setncattr("cell_measures", 'area: cell_area')
+            dataset.setncattr("external_variables", ["cell_area"])
+            
+            # run the check
+            results = self.cf.check_cell_measures(dataset)
+            score, out_of, messages = self.get_results(results)
+            assert score > 0
```
benjwadams: Could combine these asserts with `and`

compliance_checker/tests/test_cf.py
```diff
@@ -94,8 +93,13 @@ def test_check_data_types(self):
 
         dataset = self.load_dataset(STATIC_FILES['bad_data_type'])
         result = self.cf.check_data_types(dataset)
-        assert result.msgs[0] == u'The variable temp failed because the datatype is int64'
-        assert result.value == (6, 7)
+
+        # TODO
+        # the acdd_reformat_rebase branch has a new .nc file that I made
```
benjwadams: Try to avoid use of the first person in technical writing.

compliance_checker/suite.py
```diff
@@ -186,6 +186,7 @@ def _get_valid_checkers(self, ds, checker_names):
         the user selected names.
         """
 
+        import pdb; pdb.set_trace()
```
benjwadams: Please remove.  I also sometimes use a pre-commit Git hook to ensure debugger statements aren't present.

compliance_checker/cf/appendix_f.py
```diff
@@ -235,3 +235,33 @@
         )
     ]
 }
+
+grid_mapping_dict17 = grid_mapping_dict16.copy() # need shallow copy; update() returns None
+grid_mapping_dict17.update({
+    'geos': [
```
benjwadams: Shouldn't this be `geostationary` per http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_geostationary_projection?

compliance_checker/tests/test_cf.py
```diff
@@ -1010,3 +1014,197 @@ def test_units_temporal(self):
...
+        dataset.createVariable('a', 'd', ('time', ))
+        dataset.variables['a'][0] = -299 # set some arbitrary val to not all equal
+        dataset.variables['a'][1] = 10E36 # set some arbitrary max > _FillValue default
+        dataset.variables['a'].setncattr("valid_range", [1, 3])  # should conflict
+        dataset.variables['a'].setncattr("actual_range", [-299, 10E36])
+        result = self.cf.check_actual_range(dataset)
+        score, out_of, messages = self.get_results(result)
+        assert score < out_of
+        assert len(messages) == 1
+        assert messages[0] == "\"a\"'s actual_range must be == valid_range"
```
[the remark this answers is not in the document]
daltonkell: Yes, that is actually the specification. That will be corrected

compliance_checker/cf/appendix_f.py
daltonkell: Ah... yeah. Looks like I typed too quickly and overlooked that

compliance_checker/suite.py
daltonkell: I don't see this in my code -- is this from the latest commit?

compliance_checker/tests/test_cf.py
daltonkell: Good point -- this will be addressed for the next commit

compliance_checker/suite.py
benjwadams: Guess this is outdated, so ignore.

benjwadams:
`compliance_checker/tests/data/bad_data_type.nc`

Was this file being tracked?  I thought we only tracked `.cdl`s

benjwadams:
> `compliance_checker/tests/data/bad_data_type.nc`
> 
> Was this file being tracked? I thought we only tracked `.cdl`s

`*.nc` is in .gitignore and the `.nc` files for testing should be generated from `.cdl` if they aren't present.   You can try issuing `git rm --cached <files>` on the NetCDF files in the test directory to remove them from future tracking.

daltonkell:
> > `compliance_checker/tests/data/bad_data_type.nc`
> > Was this file being tracked? I thought we only tracked `.cdl`s
> 
> `*.nc` is in .gitignore and the `.nc` files for testing should be generated from `.cdl` if they aren't present. You can try issuing `git rm --cached <files>` on the NetCDF files in the test directory to remove them from future tracking.

I just checked and you're right, but I also noticed that there are plenty of `.nc` files already in this repo. Would you like me to remove them as well?

benjwadams:
Closing in favor of #627

Issue #626, opened 2019-02-07
check_climatological_statistics applied to non-climatology variables


mhidas:
I'm just  testing the latest release (4.2.0) and getting warnings about climatological variables for a dataset that doesn't have any!

For example, [this file](http://thredds.aodn.org.au/thredds/catalog/IMOS/ABOS/SOTS/Pulse/real-time/catalog.html?dataset=IMOS/ABOS/SOTS/Pulse/real-time/IMOS_ABOS-SOTS_W_20150325T110000Z_Pulse_FV00_Pulse-11-2015-MRU-Surface-wave-height-realtime.nc) has (subset of CDL):
```
dimensions:
        TIME = 8526 ;
variables:
        double TIME(TIME) ;
                TIME:standard_name = "time" ;
                TIME:units = "days since 1950-01-01 00:00:00 UTC" ;
         float ACCEL_STDDEV(TIME) ;
                ACCEL_STDDEV:cell_methods = "TIME: standard_deviation (interval: 10 min)" ;
                ACCEL_STDDEV:coordinates = "TIME LATITUDE LONGITUDE" ;
                ACCEL_STDDEV:units = "metres second-2" ;
        float ACCEL_AVG(TIME) ;
                ACCEL_AVG:cell_methods = "TIME: mean (interval: 10 min)" ;
                ACCEL_AVG:coordinates = "TIME LATITUDE LONGITUDE" ;
                ACCEL_AVG:units = "metres second-2" ;
```
As far as I can tell this is perfectly valid use of the `cell_methods` attribute, and they are **not** climatological variables. However, when I run the checker v4.2.0 on it, I get:
```
                                    Warnings                                    
--------------------------------------------------------------------------------
§7.4 Climatological Statistics
* The "time: method within years/days over years/days" format is not correct in variable ACCEL_STDDEV.
* The "time: method within years/days over years/days" format is not correct in variable ACCEL_AVG.
* The "time: method within years/days over years/days" format is not correct in variable ACCEL_STDDEV.
* The "time: method within years/days over years/days" format is not correct in variable ACCEL_AVG.
```

mhidas:
It looks like the problem is [here](https://github.com/ioos/compliance-checker/blob/master/compliance_checker/cf/cf.py#L2970), applying the climatology-specific regex to _all_ `cell_methods` attributes, not just those of identified climatology data variables.

[...]

PR #628, opened 2019-02-07
Fixes issue of checking for climatological cell_methods when climatol…

daltonkell:
…ogy attr not present

PR for #626 @mhidas

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -2907,9 +2907,8 @@ class CFBaseCheck(BaseCheck):
         # find the variables with climatology attributes
         clim_containing_vars = ds.get_variables_by_attributes(
             climatology=lambda s: s is not None)
-        clim_var = clim_containing_vars[0] if clim_containing_vars else None  # NOTE why only get the first element? There could be multiple, no?
 
-        if clim_var:
+        for clim_var in clim_containing_vars:
             if hasattr(clim_var, 'bounds'):
                 reasoning.append('Variable {} has a climatology attribute and cannot also have a bounds attribute.'.format(clim_var.name))
                 result = Result(BaseCheck.MEDIUM,
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -2888,7 +2888,7 @@ class CFBaseCheck(BaseCheck):
         total_climate_count = 0
         valid_climate_count = 0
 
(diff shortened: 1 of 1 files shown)


compliance_checker/cf/cf.py
```diff
@@ -2903,10 +2903,12 @@ def check_climatological_statistics(self, ds):
         # variable, we need to make sure it has the attribute "climatology",
         # but not the attribute "bounds"
         meth_regex = "(?:{})".format("|".join(methods))
+
+        # find the variables with climatology attributes
         clim_containing_vars = ds.get_variables_by_attributes(
             climatology=lambda s: s is not None)
```
mhidas: I think the code here is doing the right thing, but the naming is confusing. This is identifying _climatological time axis_ variable(s) (see [CF 7.4](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#climatological-statistics), second paragraph), _not_ variables containing the actual climatological statistics (those would have this variable as a coordinate).

compliance_checker/cf/cf.py
```diff
@@ -2952,34 +2954,33 @@ def check_climatological_statistics(self, ds):
...
+            # info, e.g.
+            # "time: method1 within years time: method2 over years (sidereal years)"
+
+            meth_regex = "(?:{})".format("|".join(methods))
+            re_string = (r'^time: {0} within (years|days)'
+                         r' time: {0} over \1(?<=days)(?: time: {0} over years)?'
+                         r'(?: \([^)]+\))?$'.format(meth_regex))
+
+            # find any variables with a valid climatological cell_methods
+            for cell_method_var in ds.get_variables_by_attributes(cell_methods=lambda s: s is not None):
```
mhidas: Strictly speaking this check should only be applied to variables that have `clim_var`  as a coordinate. (CF doesn's say you can't have non-climatological variables in the same file -- though it would probably be unwise to do so.)

compliance_checker/cf/cf.py
mhidas: This also implies that in the loop below, `clim_var` is a time _coordinate variable_, in the CF sense, i.e. it has only one dimension, with the same name as itself. So I don't understand the reference to "2d coordinate bounds" in line [2929](https://github.com/ioos/compliance-checker/blob/master/compliance_checker/cf/cf.py#L2929) below??

PR #630, opened 2019-02-13
Re-tool the climatology check for clarity and efficiency

daltonkell:
Additionally, makes some regex strings with invalid escape sequences valid by making the string a raw string

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -794,7 +794,7 @@ class CFBaseCheck(BaseCheck):
         valid = False
         reasoning = []
         if hasattr(ds, 'Conventions'):
-            conventions = regex.split(',|\s+', getattr(ds, 'Conventions', ''))
+            conventions = regex.split(r',|\s+', getattr(ds, 'Conventions', ''))
             for convention in conventions:
                 if convention == 'CF-1.6':
                     valid = True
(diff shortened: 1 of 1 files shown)


daltonkell:
@mhidas This implements your comments on #628. If you think it's good to go, I'll merge and make a quick patch release

compliance_checker/cf/cf.py
```diff
@@ -2899,77 +2900,84 @@ def check_climatological_statistics(self, ds):
...
+            elif clim_coord_var.climatology not in ds.variables:
                 reasoning.append("Variable {} referenced in time's climatology attribute does not exist".format(ds.variables['time'].climatology))
                 result = Result(BaseCheck.MEDIUM,
                                 False,
                                 (self.section_titles['7.4']),
                                 reasoning)
                 ret_val.append(result)
                 return ret_val
+
             # handle 1-d and 2d coordinate bounds
```
mhidas: I can't imagine any case where the time coordinate would be 2-dimensional??
As far as I can tell `clim_coord_var.ndim` should always be 1, which would allow the code and reasoning message below to be much simpler...
(I realise this code is from a couple of years ago ... https://github.com/ioos/compliance-checker/commit/1f7c70b413598a078863b9f11141c5a8d6f6bdad)

benjwadams: http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#climatological-statistics

> Intervals of climatological time are conceptually different from ordinary time intervals; a given interval of climatological time represents a set of subintervals which are not necessarily contiguous. To indicate this difference, a climatological time coordinate variable does not have a bounds attribute. Instead, it has a climatology attribute, which names a variable with dimensions (n,2), n being the dimension of the climatological time axis.

So yeah, this probably could be simplified.

PR #631, opened 2019-02-18
support url fillmismatch workaround

ocefpaf:
Latest `netcdf-c` broke a lot of our datasets due to https://github.com/Unidata/netcdf-c/issues/1299. For now we have a workaround by adding `#fillmismatch` to the end of the URL endpoint, but that breaks `compliance-checker`'s `is_opendap`.

This PR adds a support for the URL with the workaround.

### DIFF
--- a/compliance_checker/protocols/opendap.py
+++ b/compliance_checker/protocols/opendap.py
@@ -14,7 +14,10 @@ def is_opendap(url):
     :param str url: URL for a remote OPeNDAP endpoint
     '''
     # If the server replies to a Data Attribute Structure request
-    das_url = url + '.das'
+    if url.endswith('#fillmismatch'):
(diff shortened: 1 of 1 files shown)


ocefpaf:
Thanks @Bobfrat! It would be nice to get a new release with this and the other changes so far.

Bobfrat:
👍 
I think @daltonkell is tidying up the HTML output option after all the updates the the group scoring, then we should be good for another release

ocefpaf:
> I think @daltonkell is tidying up the HTML output option after all the updates the the group scoring, then we should be good for another release

Awesome! Thanks!

daltonkell:
Good to go @Bobfrat @ocefpaf

Issue #519, opened 2017-09-20
ACDD 1.3 date_issued and date_modified question


Bobfrat:
These attributes are suggested and hence their presence is checked as BaseCheck.LOW. However if it does exist, the [check](https://github.com/ioos/compliance-checker/blob/master/compliance_checker/acdd.py#L607-L627) that tests if the string is ISO compliant is a BaseCheck.MEDIUM. 

So in this case you'd get a better compliance score if you omitted the attribute rather than provided a non ISO date string. Is this desired? Or should the ISO check be a BaseCheck.LOW as well?

wckoeppen:
Are any dates actually being checked for ISO compliance?

I had some dates as "March 2014" and no errors were reported by compliance_checker.__version__ = 3.1.1 We figured out there was an issue when they got stuck going into the IOOS catalog.

benjwadams:
I don't see a BaseCheck.MEDIUM in the current result.
Could you confirm this, or point out what I'm missing?

Bobfrat:
This is where their existence is defined as LOW: https://github.com/ioos/compliance-checker/blob/6fcb4babc5ef6ab7ef2d03e77a5adccb76cda231/compliance_checker/acdd.py#L67-L78

This is where the ISO compliant check is MEDIUM: https://github.com/ioos/compliance-checker/blob/6fcb4babc5ef6ab7ef2d03e77a5adccb76cda231/compliance_checker/acdd.py#L652-L672

benjwadams:
@Bobfrat, OK to close after PR?

Bobfrat:
i'll give it a test drive

PR #642, opened 2019-04-15
Alter date_issued and date_modified ISO date validation as LOW

benjwadams:
Related to #519


benjwadams:
Going to try to patch this at the global attribute level.

Issue #534, opened 2017-10-17
Incorrect return value when running multiple check suites 


mhidas:
When a single call to the checker requests multiple check suites, the final True/False return value is determined by just one of the suites, while the others are effectively ignored.

E.g. this if a file passes the CF checks:
```python
In [4]: from compliance_checker.runner import ComplianceChecker, CheckSuite

In [5]: CheckSuite.load_all_available_checkers()

In [6]: infile = 'cf_yes_acdd_no.nc'

In [7]: ComplianceChecker.run_checker(infile, ['cf'], 0, 'normal')
--------------------------------------------------------------------------------
                    The dataset scored 291 out of 291 points                    
                              during the cf check                               
--------------------------------------------------------------------------------
All tests passed!
Out[7]: (True, False)

```
... but fails ACDD:
```python
In [8]: ComplianceChecker.run_checker(infile, ['acdd'], 0, 'normal')


--------------------------------------------------------------------------------
                     The dataset scored 42 out of 60 points                     
                             during the acdd check                              
--------------------------------------------------------------------------------
                               Scoring Breakdown:                               
...
Out[8]: (False, False)
```

However, if I run both check suites at once,
```python
In [9]: ComplianceChecker.run_checker(infile, ['acdd', 'cf'], 0, 'normal')


--------------------------------------------------------------------------------
                     The dataset scored 42 out of 60 points                     
                             during the acdd check                              
--------------------------------------------------------------------------------

...

--------------------------------------------------------------------------------
                    The dataset scored 291 out of 291 points                    
                              during the cf check                               
--------------------------------------------------------------------------------
All tests passed!
Out[9]: (True, False)
```
I get the text correct output for both suites, but the final result is `True`, i.e. compliant.

Surely the correct return value in this case would be `False`?

I think the problem is in the [ComplianceChecker.stdout_output](https://github.com/ioos/compliance-checker/blob/3a696d14aab08638b3b4ebb47fc4f8cfcb2444e6/compliance_checker/runner.py#L104) method, which only returns the result `groups` for the last check suite it iterates over.

benjwadams:
@mhidas, I've just gotten back around to revisiting some of the compliance checker results.  I'm pretty sure the above commit will fix the issues described, but I don't have any files like the one you described where all values are passing for a particular checker.  Do you still have that input file and could I use it or something similar to create a test to ensure that it works as desired?

benjwadams:
[...]

PR #644, opened 2019-04-16
Handle return value from multiple check suites

benjwadams:
Handles the return status of multiple check suites, returning non-zero
if any of the check suites did not pass and/or if unhandled
compliance-checker errors.  Possible fix for #534.

### DIFF
--- a/compliance_checker/runner.py
+++ b/compliance_checker/runner.py
@@ -48,6 +48,7 @@ class ComplianceChecker(object):
 
         @returns                If the tests failed (based on the criteria)
         """
+        all_groups = []
         cs = CheckSuite()
         # using OrderedDict is important here to preserve the order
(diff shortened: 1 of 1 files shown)


daf:
:+1: from me too.

I've seen similar issues with conda-forge stuff today, I think that channel might just have some core package related issues, let's hope it's cleared up tomorrow.

benjwadams:
Ideally I'd like to have some metadata which passes all checks for one suite but fails for one or issues in the other for the purposes of unit testing.  I currently can't find any metadata which meets this criteria, even with some of the IOOS "gold standard" metadata I've looked at.

benjwadams:
I'll merge anyways, but note that we should add a test case when we receive metadata like the above.

PR #645, opened 2019-04-17
Fix Travis-CI

ocefpaf:
@benjwadams and @daf this should fix the CI failures you are seeing.

For more info on the problem see: https://conda-forge.org/docs/user/tipsandtricks.html#how-to-fix-it

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,16 +40,19 @@ matrix:
       env: TEST_TARGET=cc-checker-ugrid PY=3.7
 
 before_install:
-  - wget http://bit.ly/miniconda -O miniconda.sh
-  - bash miniconda.sh -b -p $HOME/miniconda
-  - export PATH="$HOME/miniconda/bin:$PATH"
-  - conda config --set always_yes yes --set changeps1 no --set show_channel_urls true
-  - conda update conda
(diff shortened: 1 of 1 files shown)


daf:
@ocefpaf ok thanks - i thought I read you had recommended removing defaults and going conda-forge only, but now I can't find a reference to that anywhere.  Is [your gist](https://gist.github.com/ocefpaf/863fc5df6ed8444378fbb1211ad8feb1) still modern?

ocefpaf:
> i thought I read you had recommended removing defaults and going conda-forge only, 

We had that as a workaround from before we had `strict`, a feature I've been fight for since day 1 with conda, but now that we have we can drop all the workarounds. One of those was the bad habit of vendorizing the compilers (the error we had here).

> but now I can't find a reference to that anywhere. Is [your gist](https://gist.github.com/ocefpaf/863fc5df6ed8444378fbb1211ad8feb1) still modern?


Yep. I'm trying to keep that one updated. Note that I have both channels and `strict` enabled. BTW, `strict` will be default in conda 4.7, so we will be able to drop it.

daf:
@ocefpaf perfect, thanks.  I've updated my local .condarc to reflect.

Issue #468, opened 2017-01-25
Create check to verify valid_min valid_max data type


lukecampbell:
For some datasets the `valid_min` and `valid_max` variable attribute has a mismatched data type from the parent variable.

benjwadams:
I'm assuming `valid_range` also falls under this same category, as well as the `actual_min`, `actual_max`, and `actual_range` attributes added in CF-1.7.

lukecampbell:
`valid_range` is an array attribute but the data type of the array should match the parent variable. 

Example:
```
float temperature(time);
    temperature:standard_name = "air_temperature";
    temperature:units = "deg_C";
    temperature:valid_range = -20.0f, 50.0f
```

benjwadams:
Appears to be already handled by https://github.com/ioos/compliance-checker/blob/master/compliance_checker/cf/cf.py#L3036-L3056

I need to check the tests as well as with some files that have this issue.

mhidas:
>  Appears to be already handled by https://github.com/ioos/compliance-checker/blob/master/compliance_checker/cf/cf.py#L3036-L3056

Yes, but that's only within `check_packed_data`. This check  should apply to all variables.

And it should also apply to the `_FillValue` attribute (http://cfconventions.org/Data/cf-conventions/cf-conventions-1.6/build/cf-conventions.html#missing-data).

benjwadams:
Whoops, my omission.

PR #646, opened 2019-04-17
Verifying Special-Attribute Data Types

daltonkell:
Per issue #468, some special attributes defined in the NetCDF User Guide (NUG)
had data types which did not match their parent variable. E.g.
    float temperature(time);
        temperature:standard_name = "air_temperature";
        temperature:units = "deg_C";
        temperature:valid_range = -20, 50; # --> integers, not floats

Currently, type-mismatches for these special attributes are checked in
"check_packed_data()" for only variables which are packed. This commit
adds a new check which examines each variable. If the variable has one
of these special attributes and a type-mismatch exists, it is scored.
This commit also adds unit tests for the new check.

It is worth noting that, while "actual_range", "actual_min", and
"actual_max" are not defined in the CF-1.6 spec, this check includes
them for easier integration into the CF-1.7.

@mhidas I couldn't request your review, so tagging you here as you commented on the issue

### DIFF
--- a/compliance_checker/tests/test_cf.py
+++ b/compliance_checker/tests/test_cf.py
@@ -13,6 +13,7 @@ from compliance_checker.tests.helpers import MockTimeSeries, MockVariable
 from compliance_checker.cf.appendix_d import no_missing_terms
 from itertools import chain
 
+import numpy as np
 import os
 import re
 import sys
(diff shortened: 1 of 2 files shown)


compliance_checker/tests/test_cf.py
```diff
@@ -97,6 +98,41 @@ def test_check_data_types(self):
...
+        ds.variables['temp'].setncattr("valid_min", 45)
+        result = self.cf.check_child_attr_data_types(ds)
+        self.assert_result_is_bad(result)
+        
+        # now give invalid string for valid_max
+        ds.variables['temp'].setncattr("valid_max", "eighty")
+        result = self.cf.check_child_attr_data_types(ds)
+        self.assert_result_is_bad(result)
+
+        # TODO for CF-1.7: actual_range, actual_min/max
```
benjwadams: CF 1.7 will likely be tested in a different unit test class.

PR #649, opened 2019-04-18
Handle vertical min/max, including zero length arrays

benjwadams:
Adds unit tests for geospatial vertical min/max.  Also handles case
where all Z axis array is empty or all values are masked by raising a
failure case, as the min/max cannot be established from an empty input
array.

### DIFF
--- a/compliance_checker/acdd.py
+++ b/compliance_checker/acdd.py
@@ -400,30 +400,48 @@ class ACDDBaseCheck(BaseCheck):
         :param netCDF4.Dataset ds: An open netCDF dataset
         :param str z_variable: Name of the variable representing the Z-Axis
         '''
-        vert_min = float(ds.geospatial_vertical_min)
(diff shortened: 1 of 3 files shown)


compliance_checker/acdd.py
```diff
@@ -400,30 +400,48 @@ def _check_total_z_extents(self, ds, z_variable):
         :param netCDF4.Dataset ds: An open netCDF dataset
         :param str z_variable: Name of the variable representing the Z-Axis
         '''
-        vert_min = float(ds.geospatial_vertical_min)
-        vert_max = float(ds.geospatial_vertical_max)
         msgs = []
         total = 2
+        try:
```
daltonkell: Definitely an omission by me to not wrap these in a try-except like I had done for the lat and lon extents, which is essentially the same check. Good catch.

PR #652, opened 2019-04-23
Fixes multi checker return status

benjwadams:
Fixes an issue when multiple test suites were selected where only the
last result was returned.  Adds unit test to verify checks are working
as intended.


benjwadams:
Should fix #534, see unit tests

PR #658, opened 2019-06-27
Handle bad CF standard name table versions

benjwadams:
Handles bad and unparseable CF standard name table versions by raising a
warning and returning `False` inside `_find_cf_standard_name_table`


benjwadams:
@Bobfrat, this fixes the earlier issues we encountered with the standard name version string.  See also the unit test included, which uses the same bad attribute string for `standard_name_vocabulary`

Issue #610, opened 2018-08-30
External checker plugins without either cc_spec and cc_spec_version attributes hidden from checker listing


daf:
e.g. cc-plugin-glider, which does not yet implement `_cc_spec` attributes (but does have `_cc_spec_version`, see ioos/cc-plugin-glider#26)

Logic intended to keep `...:latest` version checkers from showing in the non-verbose list also squelches plugins that haven't switched to the latest name/version attr spec.  Verbose listing will show them.

daltonkell:
Is this good to close as [ioos/cc-plugin-glider#26] (https://github.com/ioos/cc-plugin-glider/pull/26) has been merged?

benjwadams:
^ We should list those checkers anyways, but going forward prefer that a version number is supplied.

benjwadams:
Implemented by #659, closing.

Issue #557, opened 2018-03-09
Moving to Python3


benjwadams:
Back when compliance checker was first written, scientific library support was rather mixed for Python 3.  Nowadays, almost all (or maybe all?) the libraries used by compliance checker have Python 3 support.  We currently support Python 2.x series, but it is a major nuisance in some categories, i.e. shoehorning unicode support and all the differences between how 2 and 3 handle unicode.  There are also other features that are supported in Python 3 which would probably be desirable for compliance checker, such as type annotations and type checking support.   This issue is intended to discuss moving away from Python 2 to solely 3, advantages, potential pitfalls, etc.

Cc: @Bobfrat, @kwilcox, @daf

MBARIMike:
I'm running compliance-checker in Python 3.6 and haven't see any issues. So maybe all is O.K. ?

One of the other features is asyncio which could greatly improve the speed of walking large OPeNDAP directories. Is there any comment on the possibility of asyncio working with compliance-checker?

kwilcox:
Hey Mike - have you tried using xarray/dask to crawl a large directory of datasets? I usually crawl though a nested directory structure for netCDF files like so but would be interested in some comparisons if you have an `asyncio` or `concurrent.futures` (https://docs.python.org/3.6/library/concurrent.futures.html) solution implemented.

```
import xarray
from pathlib import Path
from dask.distributed import Client
from dask.diagnostics import ProgressBar

client = Client()
ncpath = Path('/mnt/store/data/packrat/cordc/east_gulf_coast_1km/processed/')
globfiles = sorted(list(ncpath.glob('**/*.nc')))
with ProgressBar():
    ds = xarray.open_mfdataset(
        globfiles,
        parallel=True,
        compat='broadcast_equals',
        data_vars='all',
        coords='minimal',
        chunks={}
    )
```

benjwadams:
I should open a milestone instead.

benjwadams:
@giumas: this is less about compatibility with Python 2 and 3 and more about moving to just Python 3 in some new compliance checker version.  In particular, handling Unicode in Python 2 has been fairly painful.

I think it'd be good to revisit this sometime after the CF 1.7 milestone is completed.

giumas:
@benjwadams got it! Given the coming [Py2 EOF](https://pythonclock.org/), it makes total sense.

benjwadams:
[...]

PR #659, opened 2019-07-02
Allows checkers using .name attr to be listed

benjwadams:
Allows checkers using the `.name` attribute instead of the `._cc_spec` and
`cc_spec_version` to be listed in the checker output.  Checkers using
the former will raise a `DeprecationWarning` indicating the checker is
not preferred.  Also refactors some of the check suite logic for
printing listings and adds unit test for aforementioned behavior.
Addresses #610.

### DIFF
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -59,7 +59,7 @@ class CheckSuite(object):
 
         return cls.suite_generators
 
-    def _print_suites(self, verbose=False):
+    def _print_suites(self, verbose=0):
         """
         Prints out available check suites.  If the verbose argument is True,
         includes the internal module version number of the check and also displays
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -37,6 +37,7 @@ if sys.stdout.encoding is None:
 if sys.stderr.encoding is None:
     sys.stderr = codecs.getwriter('utf8')(sys.stderr)
 
+
 class CheckSuite(object):
 
     checkers = {}       # Base dict of checker names to BaseCheck derived types, override this in your CheckSuite implementation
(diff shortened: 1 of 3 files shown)


compliance_checker/tests/test_cli.py
```diff
@@ -92,6 +95,41 @@ def test_single_json_output(self):
...
+
+    def test_list_checks(self):
+        """
+        Tests listing of both old-style, deprecated checkers using .name
+        attributes, and newer ones which use ._cc_spec and _cc_spec_version
+        attributes
+        """
+
+        # hack: use argparse.Namespace to mock checker object with attributes
+        # since SimpleNamespace is Python 3.3+ only
```
daf: are we still supporting <3.3?!?

compliance_checker/suite.py
```diff
@@ -83,10 +101,21 @@ def load_all_available_checkers(cls):
...
+        for c in checkers:
             try:
-                xl = x.resolve()
-                cls.checkers[':'.join((xl._cc_spec, xl._cc_spec_version))] = xl
+                check_obj = c.resolve()
+                cls.checkers[':'.join((check_obj._cc_spec,
+                                       check_obj._cc_spec_version))] = check_obj
             # TODO: remove this once all checkers move over to the new
             #       _cc_spec, _cc_spec_version
             except AttributeError:
```
daf: this is probably fine, but maybe a future cleanup to explicitly check for `_cc_spec`/`_cc_spec_version` with `hasattr` instead of assuming the `AttributeError` has to do with these two lookups.

compliance_checker/suite.py
```diff
@@ -95,7 +124,16 @@ def load_all_available_checkers(cls):
                 # the checker as the main class
                 # TODO: nix name attribute in plugins.  Keeping in for now
```
daf: feels like this wording can be slightly updated too

compliance_checker/tests/test_cli.py
benjwadams: As Python 2.7 isn't deprecated yet and much of the code supports Python 2.7, yes.  Handling of various Python 2.7 quirks/differences, especially Unicode, has been a pain point.  Thus, I'd like to do away with backwards compatibility with Python 2.7 in the next major version, especially with the upcoming Python 2.7 EOL.  See https://github.com/ioos/compliance-checker/issues/557 , which I closed mainly to clean up some issues.

daf: sure that's fine.  maybe add this as a list item of pending things to clear out when 2.7 is yanked.

PR #661, opened 2019-07-03
Draft --describe-checks feature

benjwadams:
Draft PR for `-D`/`--describe-checks` feature.  User can specify a
particular checker for which to list all defined checks, or '*' to list
all defined checks for all defined checkers.  Prints abridged docstrings
from each `check_*` method in checker classes.  Useful in combination
with `-l` output, or to describe the behavior of certain checks should a
user wish to disable some with `-s`/`--skip-checks`.  PR is still draft
quality and will likely need improvements on formatting and display of
results, but for the time being, I just want to solicit feedback on the general functionality.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,9 +119,9 @@ parse it's contents.
 
 ```
 usage: cchecker.py [-h] [--test TEST] [--criteria [{lenient,normal,strict}]]
-                   [--verbose] [--skip-checks SKIP_CHECKS]
+                   [--verbose] [--describe-checks] [--skip-checks SKIP_CHECKS]
                    [-f {text,html,json,json_new}] [-o OUTPUT] [-V] [-l]
-                   [-d DOWNLOAD_STANDARD_NAMES]
+                   [-d [DOWNLOAD_STANDARD_NAMES]]
                    [dataset_location [dataset_location ...]]
 
 positional arguments:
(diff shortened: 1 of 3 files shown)


benjwadams:
@mhidas, I can't add you as a reviewer, but please feel free to weigh in on this draft functionality.

mhidas:
@benjwadams Thanks. It looks like a helpful feature. Unfortunately I've run out of time today, and I'm on leave for the next two weeks, so won't be able to look at it in detail.

benjwadams:
@daf 
>     * the text formatting is very hard to read - descriptions and the following test names all run together. maybe use list ticks and indenting to help the eye differentiate?  Example (from acdd): ...

I've updated the formatting to be more in line with what you've suggested above.

> specfiying an unknown checker gives an error message with a python object repr:

Fixed.  I had forgotten to use the `file` kwarg in the `print` function to specify stderr.

> specifying * as indicated in help text doesn't work:

Not able to replicate this.  The command as written `python cchecker.py -D *` will probably not work as intended in many shells due to glob expansion.  `python cchecker.py -D '*'` did work for me.  But perhaps it'd be better just to implement it as your next point:

> this is a maybe, but should the -t parameter be used to specify tests like normal, and just have -D take no arguments and do the describe feature? If no -t specified, warn/error that nothing specified, and default to printing list of available checkers?

I hadn't considered this, but this is probably a better design since we already have logic for the `-t` flag.

daf:
> I've updated the formatting to be more in line with what you've suggested above.

:+1: looking much clearer, thanks

> The command as written python cchecker.py -D * will probably not work as intended in many shells due to glob expansion.

Ah good point, wasn't thinking about that.

benjwadams:
@daf, rewrote to use `-t` in tandem with `-D`.  `-D` is now evaluated for simple presence with `action='store_true`' and prints any checker's check functions defined in `-t`, or all defined checkers when `-t` is not present.

daf:
Looks good, thanks @benjwadams !

benjwadams:
Going to squash down some of these commits, rebase against master, change the branch name and open a new, non-draft commit.  Thanks for the feedback.

PR #665, opened 2019-08-02
Adds a new method `get_test_ctx` to reuse `TestCtx`

benjwadams:
Adds a new method to `compliance_checker.base.BaseCheck` called
`get_test_ctx`, which attempts to fetch a `TestCtx` from a defaultdict
specific to the checker instance.  If the check does not exist, it is
created in the dict, otherwise, the corresponding dict entry is returned
for further modifications.  Allows predefined `TestCtx` objects to be
reused across functions so that the same `TestCtx` doesn't have to be
passed around one monolithic function.  Potentially also useful for
extending checking behavior defined in newer checker versions, i.e.
additions to check behavior between CF versions.

### DIFF
--- a/compliance_checker/base.py
+++ b/compliance_checker/base.py
@@ -13,6 +13,7 @@ from owslib.swe.sensor.sml import SensorML
 from owslib.namespaces import Namespaces
 from compliance_checker import __version__, MemoizedDataset
 from compliance_checker.util import kvp_convert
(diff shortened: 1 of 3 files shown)


benjwadams:
Many checks create a `TestCtx` per variable, so that will need to be accounted for.  Hold off on approving until then.

Issue #664, opened 2019-07-30
Checklist for CF 1.7 implementation


benjwadams:
# Checklist of tasks to implement for CF 1.7 milestone
*Feel free to update or make suggestions as needed*

- [x] [#31](https://cf-trac.llnl.gov/trac/ticket/31) Adding an attribute for `actual_range` , see [Section 2.5.1](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#missing-data)
- [x] [#118](https://cf-trac.llnl.gov/trac/ticket/118), Add `geoid_name` and `geopotential_datum_name` to the list of Grid Mapping Attributes. see [Table F1](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#table-grid-mapping-attributes)
- [x] [#148](https://cf-trac.llnl.gov/trac/ticket/148) Added `maximum_absolute_value`, Added maximum_absolute_value, minimum_absolute_value and mean_absolute_value to cell methods in [Appendix E ](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#appendix-cell-methods)
- [x] [#77](https://cf-trac.llnl.gov/trac/ticket/77), Add sinusoidal projection [here](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_sinusoidal)
- [x] [#92](https://cf-trac.llnl.gov/trac/ticket/92) Add oblique mercator projection, see [here](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_oblique_mercator)
- [x]  [#72](https://cf-trac.llnl.gov/trac/ticket/72), Adding the [geostationary projection](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_geostationary_projection)
- [x] [#102](https://cf-trac.llnl.gov/trac/ticket/102), additional `cell_methods` (`root_mean_square` and `sum_of_squares`) , changes in Appendix E and section 7.3; see [here](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#appendix-cell-methods)

- [x]  [#61](https://cf-trac.llnl.gov/trac/ticket/61) Two new cell methods Appendix E
  `root_mean_square` and `mean_of_upper_decile`
[...]

Issue #295, opened 2016-09-05
Why are subclasses of the user-selected checker classes also run?


mhidas:
The `_get_valid_checkers` method in ([suite.py](https://github.com/ioos/compliance-checker/blob/master/compliance_checker/suite.py#L88)) is supposed to filter out, from a list of checkers selected by the user, those checker classes that are appropriate for the dataset being checked. However, it also sneakily adds any subclasses of those checkers, leading to unexpected results.

On [line 109](https://github.com/ioos/compliance-checker/blob/master/compliance_checker/suite.py#L109-L113):

``` python
            # add any subclasses of the checker class
            for subc in a.__subclasses__():
                if subc not in all_checked:
                    all_checked.add(subc)
                    checker_queue.add((name, subc))
```

I am in the process of creating version 1.4 of the IMOS checker, which I was hoping to be simply a subclass of the v1.3 checker class. However, with the code above, if I do this I can no longer run just the v1.3 checks, because v1.4 is then automatically selected.

I could avoid this by following the example in acdd.py (with the different versions being subclasses of a common base class). I just wonder whether there's a use case that makes the above behaviour desirable?

kwilcox:
Valid point @mhidas. I have actually consider this a "feature" since we can subclass something like "CF1.6" and implement very specific checks in addition to those in "CF1.6".  Maybe it's time to consider the ability to run more than one checker at once via all interfaces?

lukecampbell:
@daf do you recall why we took this approach? It's been a while since I dug into the checker selection mechanics.

daf:
It's rooted in being able to support multiple types of incoming data sources and apply the same checks to both: netcdf files/opendap urls AND sos gc/ds xml files/urls.  The idea was to define your checks in a Base class, and have a stub class inheriting from both your specific checker's Base and the BaseNCCheck/BaseSOSGCCheck/BaseSOSDSCheck classes to pair them up with sources.  The IOOS metadata checker is the only one still operating in that fashion, CF and ACDD appear to have moved on (CF wouldn't really make sense there anyway but ACDD has a case, it used to work).

There was also the `register_checker` concept that got added during plugin work by @fxmzb123 but now I don't see that being used in the discovery logic.  I don't remember what happened there.

@mhidas, I'm not sure I understand the issue - how are you selecting the checker to run on the command line?  The versioning capability (`--check your_checker:1.3`) should be handling this, or I'm not understanding what's going on..

mhidas:
[...]

PR #667, opened 2019-08-13
Flexible conventions version check

daltonkell:
First commit addresses #295, which bubbled up and affected the implementation of making the version check more flexible. Separate commits were used in the event the changes should be assimilated independently. The second commit addresses no. 76 in #664.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -135,6 +135,33 @@ class CFBaseCheck(BaseCheck):
         self._find_cf_standard_name_table(ds)
         self._find_geophysical_vars(ds)
 
+    def _check_conventions_version(self, ds):
+        '''
+        CF §2.6.1 the NUG defined global attribute Conventions to the string
+        value "CF-<version_number>"; check the Conventions attribute contains
+        the appropriate string.
+
+        :param netCDF4.Dataset ds: An open netCDF dataset
+        :rtype: compliance_checker.base.Result
+        '''
+
+        valid = False
+        reasoning = []
+        correct_version_string = "{}-{}".format(self._cc_spec, self._cc_spec_version).upper()
+        if hasattr(ds, 'Conventions'):
+            conventions = regex.split(r',|\s+', getattr(ds, 'Conventions', ''))
+            for convention in conventions:
+                if convention == correct_version_string:
+                    valid = True
+                    break
+            else:
+                reasoning = ['§2.6.1 Conventions global attribute does not contain '
+                             '"{}"'.format(correct_version_string)]
+        else:
+            valid = False
+            reasoning = ['§2.6.1 Conventions field is not present']
+        return Result(BaseCheck.MEDIUM, valid, self.section_titles['2.6'], msgs=reasoning)
+
     def _dims_in_order(self, dimension_order):
         '''
         :param list dimension_order: A list of axes
@@ -882,7 +909,7 @@ class CF1_6Check(CFNCCheck):
 
     def check_conventions_are_cf_16(self, ds):
         '''
-        Check the global attribute conventions to contain CF-1.6
+        Check the global attribute conventions to contain CF-1.6.
 
         CF §2.6.1 the NUG defined global attribute Conventions to the string
         value "CF-1.6"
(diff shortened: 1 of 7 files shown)


compliance_checker/cf/cf.py
```diff
@@ -3664,6 +3678,20 @@ def check_cell_measures(self, ds):
 
         return ret_val
 
+    def check_conventions_are_cf_17(self, ds):
```
benjwadams: `1_7` would be preferable to `17` here.

compliance_checker/tests/test_cf.py
```diff
@@ -1296,3 +1296,30 @@ def test_cell_measures(self):
         score, out_of, messages = self.get_results(results)
         message = u'Cell measure variable box_area referred to by PS is not present in dataset variables'
         assert message in messages
+
+    def test_check_conventions_are_cf_16(self):
```
benjwadams: I'm probably going to remove the inheritance from `TestCF1_6` as it largely ends up being a bunch of repeated tests.

compliance_checker/ioos.py
```diff
@@ -358,6 +358,11 @@ class IOOSBaseSOSCheck(BaseCheck):
 
 
 class IOOSSOSGCCheck(BaseSOSGCCheck, IOOSBaseSOSCheck):
+    _cc_spec = 'ioos_sosGC'
+    _cc_spec_version = '0.1'
```
benjwadams: Are you sure on the IOOS SOS spec version here?

setup.py
```diff
@@ -61,7 +61,8 @@ def pip_requirements(fname='requirements.txt'):
             'cf-1.7 = compliance_checker.cf.cf:CF1_7Check',
             'acdd-1.1 = compliance_checker.acdd:ACDD1_1Check',
             'acdd-1.3 = compliance_checker.acdd:ACDD1_3Check',
-            'ioos_sos = compliance_checker.ioos:IOOSBaseSOSCheck',
+            'ioos_sosGC = compliance_checker.ioos:IOOSSOSGCCheck',
```
benjwadams: We may want to keep a special case for the IOOSBaseSOSCheck since the original behavior was to generate the GetCapabilities or DescribeSensor when pointed at an arbitrary endpoint.

compliance_checker/ioos.py
daltonkell: As far as I am aware, the versions have not changed. As these were both sub-classes of the `BaseSOSGCCheck` and `IOOSBaseSOSCheck`, which had version 0.1, I used the existing version.

compliance_checker/tests/test_cf.py
daltonkell: Without the inheritance, though,  wont't we lose most of the existing checks that `CF-1.7` *does* use?

benjwadams: Yes, but they're largely redundant because they will have the same behavior.  For those that don't we can just call the extended method and test for behavior specific to that checker subclass.

benjwadams:
@mhidas, we had to revisit #295 in order to stop the auto calling of the subclasses for the in progress CF 1.7 checker, which extends the original CF checker class.   The compromise we have arrived at was to remove the current behavior, except for the SOS checker, which relied on being able to dispatch to different classes depending on whether a `GetCapabilities` or `DescribeSensor` request needed to be issued.

mhidas:
@benjwadams Thanks for the heads up. I don't think that will be a problem for IMOS/AODN use of the checker.

benjwadams:
@daf, @agstephens , please also let us know if this behavior is OK.  I tried to check other existing plugins in GitHub's "Used By" feature and the ones I looked at don't seem to rely upon the subclass behavior, nor can I find this behavior documented.  If this looks like a reasonable solution keeping backwards compatibility with the SOS checker invocations while eliminating behavior which might surprise plugin authors.  If this seems good, I'll merge this and finally close off #295.

benjwadams:
Going forward with merging this.  Worst case scenario, we can revert/change the behavior down the line prior to merging with master if we decide it will cause problems.

agstephens:
@benjwadams I am also happy this won't cause us problems

PR #668, opened 2019-08-15
Verify OGC Vert datum name

benjwadams:
Adds functionality to check grid mapping attributes for `geoid_name` and `geopotential_datum_name`.

Also overhauls return type for `check_grid_mapping` in both CF 1.6 and 1.7 to use `collections.OrderedDict` instead of `list` so that the 1.7 check can add to the results of the common behavior of the check between 1.6 and 1.7.   This mechanism works with stdout output, but if the raw results are being used by any other programs, the behavior may need to be modified, as iterator over `OrderedDict` will return keys instead of values.  PR is marked as draft as of the time of this writing to allow for discussion about pros/cons and possible other mechanisms for extending checks between different versions of the same checker parent class.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -38,6 +38,7 @@ def print_exceptions(f):
             print_exc()
     return wrapper
 
+
 # helper to see if we should do DSG tests
 def is_likely_dsg(func):
     @wraps(func)
@@ -114,6 +115,7 @@ class CFBaseCheck(BaseCheck):
             "9.5": "§9.5 Coordinates and metadata",
             "9.6": "§9.6 Missing Data"
         }
+
 
     ################################################################################
     # Helper Methods - var classifications, etc
@@ -606,7 +608,6 @@ class CF1_6Check(CFNCCheck):
     ###############################################################################
     # Chapter 2: NetCDF Files and Components
     ###############################################################################
-
 
     def check_data_types(self, ds):
         '''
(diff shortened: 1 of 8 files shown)


benjwadams:
Some of the OrderedDict and check caching behavior probably could go in its own PR if desired, but I felt it would make more sense to display an example of how it could be used when extending behavior of a check.  In this case it was the `check_grid_mapping` check.

compliance_checker/cf/cf.py
```diff
@@ -3664,6 +3656,69 @@ def check_cell_measures(self, ds):
...
+                    except sqlite3.Error as e:
+                        # if we hit an error, skip the check
+                        warn("Error occurred while trying to query "
+                                        "Proj4 SQLite database at {}: {}"
+                                        .format(proj_db_loc, str(e)))
+            prev_return[var_name] = test_ctx.to_result()
+
+        return prev_return
+
+    def _process_v_datum_str(self, v_datum_str, conn):
```
benjwadams: I realize this is fairly ugly to to hit a SQLite3 database included with PyProj to determine whether or not there is a valid vertical datum name.   I'm certainly open to more elegant suggestions, but despite trying, I was unable to find a list of valid vertical datum names maintained by OGC, and I didn't see any API bindings in either Proj4 or PyProj to determine if a given string was a valid OGC WKT `VERT_DATUM_NAME` as requested by CF 1.7.

benjwadams:
Looks like the CI builds with Conda are failing because the Proj4 database appears to be in a different location than when installing with Pip.  Let me see if there's something inside the PyProj library that returns the database path.

PR #670, opened 2019-08-19
`crs_wkt` implementation

benjwadams:
Implements checking for `crs_wkt` grid mapping attribute under CF 1.7 checker.


benjwadams:
> If the `crs_wkt` provides only part of a grid mapping attribute, is that considered invalid?

Well, the WKT has to be a valid CRS, so if it's incomplete, proj should raise an error.
 
> Additionally, for future discussion, I believe that if the `grid_mapping` attribute overrides what is provided in the `crs_wkt`, the check should fail. The two should be congruent, no?

This is not what the CF spec mentions.  In the case where both a crs_wkt and an individual grid mapping attribute that isn't crs_wkt are present and both differ, the latter takes precedence according to the spec.  See https://github.com/pyproj4/pyproj/issues/401 for further discussion.

Issue #675, opened 2019-10-07
Handling of OPeNDAP attribute containers DODS_EXTRA and EXTRA_DIMENSION when data is accessed from TDS


neumannd:
# Software Version

IOOS Compliance Checker currently available via https://compliance.ioos.us/index.html (access date: 2019-10-07)

# Issue

I wanted to check a file, which is provided by a THREDDS via OPeNDAP.

When I access a NetCDF file from a THREDDS via OPeNDAP, the THREDDS might add two OPeNDAP attribute containers `DODS_EXTRA` and `EXTRA_DIMENSION` according to this text: https://www.unidata.ucar.edu/software/netcdf-java/current/CDM/Opendap.html 

These attribute containers and the contained attributes are considered as global attributes by the NetCDF4-C library. The name of the resulting global attributes seems to be `CONTAINER_NAME.ATTRIBUTE_NAME`. In most situations it does not seem to matter whether these global attributes are present or not. Unfortunately, the IOOS Checker does not like `EXTRA_DIMENSION.xxx` and prints a warning.

The log of the checker is attached:

[compliance_12e59c1cd05a905d3b26053d97ecd37266830c5b.txt](https://github.com/ioos/compliance-checker/files/3698385/compliance_12e59c1cd05a905d3b26053d97ecd37266830c5b.txt)

# Solution

~~Please consider ignoring all global attributes starting with DODS_EXTRA and EXTRA_DIMENSION.~~

Global attributes starting with `DODS` are already ignored in the current version of the IOOS Checker in function `check_naming_conventions` of `cf.py`.

I would like to add ignoring global attributes starting with `EXTRA_DIMENSION`.

I will create a pull request for this.

# Reproduction of the error

try the file:

http://thredds-iow.io-warnemuende.de/thredds/dodsC/meramo/IOW-THREDDS-meramo_hbm-cgt-ergom_cmaq16_2012_nosilrestart_phyrestart_iteration1_coarse_2018-12-07-18.nc

# Additional note

_removed by edit_

~~I also get Unidentifiable feature for variable errors for each variable. It did not have time to look into it but the attribute _ChunkSizes might cause this. _ChunkSizes actually is a virtual attribute of the original NetCDF file. The TDS seems to interprete it as real attribute and puts it into the file header when send via OPeNDAP. This issue only arises when a deflated NetCDF4 file is accessed via OPeNDAP.~~

[...]

PR #678, opened 2019-10-09
ignore global attr starting with EXTRA_DIMENSION

neumannd:
A THREDDS Server might add global attributes starting with EXTRA_DIMENSION as documented by unidata:
https://www.unidata.ucar.edu/software/netcdf-java/current/CDM/Opendap.html

These global attributes should be ignored by the compliance checker.


neumannd:
should fix issue #675

PR #76, opened 2015-02-12
Fix typo/line order in conda commands in README.md

daf:
Updates #65 to address an issue discovered by @abirger.

PR #680, opened 2019-10-21
Multi string conventions

daltonkell:
TracTicket #76: More Than One Name in "Conventions"
    
    Implement backwards-compatible check for CF-1.6 where the Conventions
    attribute is only permitted to contain one name. In CF-1.7, multiple names
    are allowed. Remove the inherited method from the BaseCheck, as that
    implementation does not belong there.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -139,33 +139,6 @@ class CFBaseCheck(BaseCheck):
         self._find_cf_standard_name_table(ds)
         self._find_geophysical_vars(ds)
 
-    def _check_conventions_version(self, ds):
-        '''
-        CF §2.6.1 the NUG defined global attribute Conventions to the string
-        value "CF-<version_number>"; check the Conventions attribute contains
-        the appropriate string.
-
-        :param netCDF4.Dataset ds: An open netCDF dataset
-        :rtype: compliance_checker.base.Result
-        '''
-
-        valid = False
-        reasoning = []
-        correct_version_string = "{}-{}".format(self._cc_spec, self._cc_spec_version).upper()
-        if hasattr(ds, 'Conventions'):
-            conventions = regex.split(r',|\s+', getattr(ds, 'Conventions', ''))
-            for convention in conventions:
-                if convention == correct_version_string:
-                    valid = True
-                    break
-            else:
-                reasoning = ['§2.6.1 Conventions global attribute does not contain '
-                             '"{}"'.format(correct_version_string)]
-        else:
-            valid = False
-            reasoning = ['§2.6.1 Conventions field is not present']
-        return Result(BaseCheck.MEDIUM, valid, self.section_titles['2.6'], msgs=reasoning)
-
     def _dims_in_order(self, dimension_order):
         '''
         :param list dimension_order: A list of axes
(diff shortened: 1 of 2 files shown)


mhidas:
Is there any chance we could reconsider this? I can see that the [paragraph](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#identification-of-conventions) in CF-1.6 describing the `Conventions` attribute _can_ be interpreted to mean "CF-1.6" should be the _entire_ value of this attribute (and the discussion on [ticket #76](https://cf-trac.llnl.gov/trac/ticket/76) suggests this is the intended meaning). However the same CF paragraph also refers to the [netCDF Conventions web page](http://www.unidata.ucar.edu/netcdf/conventions.html), which clearly states the possibility of listing multiple conventions as a blank-/comma-separated string. 

Why would the CF-1.6 convention need to insist on itself being the _only_ convention a file can follow?

I understand that strictly speaking you're just implementing what the convention document says. Is there any particular use case that requires this behaviour for the CF-1.6 checks?

The problem for me is that we (IMOS/AODN) have tens of thousands of files with
`Conventions = "CF-1.6,IMOS-1.4"`
So when a change like this makes it into the latest compliance-checker release, we'll either have to stop upgrading the checker we use in our data ingestion software, or explicitly skip this check.

daltonkell:
We wrestled with this a bit, and it seems most appropriate to make the checks adhere to the standards as best they can. While we agree that it's inconvenient and a bit pedantic to implement a retrospective fix for the CF 1.6 standards, the [Trac Ticket mentioned](https://cf-trac.llnl.gov/trac/ticket/76) was accepted, and we felt that the existing implementation for 1.6 was actually incorrect.

We are in the process of migrating our own data to CF-1.7, so we'll be taking a few doses of this as well.

Like you noted, there are ways to skip checks or alter the severity when using the command line tool. Providing too many workarounds will undermine the credibility of the checker as well as the standard.

benjwadams:
@mhidas, technically CF 1.6 does seem to only allow one value.  However, I have seen many files out in the wild using multiple values.  Skipping the Conventions check outright might miss some malformed conventions. A good middle ground approach might be to add a pragma/passable flag or option to the Compliance Checker to disable strict checking of the conventions string under CF 1.6, e.g. `cchecker.py -t cf:1.6 -p NON_STRICT_CF_1_6_CONVENTION <files>`.  This would be a new command line option to the checker.

mhidas:
@benjwadams Thanks for the suggestion. That would be slightly preferable to skipping the entire check, but I really don't think it's worth the effort of developing another command line option, and then _still having to add that option every time we use the checker_.

If you want to go down this path, how about making _strict_ checking optional, and allowing a list of conventions by default? I'm willing to bet strict checking will be by far the less frequently required behaviour!

@daltonkell 
> We wrestled with this a bit, and it seems most appropriate to make the checks adhere to the standards as best they can.

I agree with the general sentiment, but I don't think it needs to be adhered to above all else. In this particular case, I think nearly _everyone_ interprets the `Conventions` attribute as being a blank-/comma-separated list of conventions, as per the [netCDF Conventions](http://www.unidata.ucar.edu/netcdf/conventions.html).

For example, the [Argo Users Manual](http://dx.doi.org/10.13155/29825) (section 2.2.1, p19) specifies this global attribute as:
`:Conventions = "Argo-3.2 CF-1.6";`

[OceanSITES Data Format Reference Manual](http://www.oceansites.org/docs/oceansites_data_format_reference_manual.pdf) (section 2.1, p9):
`Conventions=”CF-1.6,OceanSITES-1.3, ACDD1.2”`

mwengren:
I've been reading this and I think you guys made the right decision to continue to allow multiple convention strings.  This is definitely quite common in practice.  

I wonder if this conflicting guidance in the CF docs has been discussed on the mailing list previously?  

Thanks for the feedback @mhidas.  Can you email the CF list to point this out as well?

PR #687, opened 2019-10-29
Revert conventions multistring

daltonkell:
Original interpretation of the CF-1.6 "Conventions" definition implemented in #680 was incongruent with user expectation and interpretation, as @mhidas points out. This PR reverts it, and adds an additional commit which renames `_check_conventions_version()` to `check_conventions_version()`, as the desired behavior is encapsulated within the inherited method.


mhidas:
:+1: Thank you!

PR #690, opened 2019-11-06
Appendix A attribute check

benjwadams:
Adds checks for attributes, their locations, and types against Appendix A for CF 1.6 and 1.7.

Implements https://cf-trac.llnl.gov/trac/ticket/86 and https://cf-trac.llnl.gov/trac/ticket/87 mentioned in #664

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -138,6 +138,14 @@ class CFBaseCheck(BaseCheck):
         self._find_metadata_vars(ds)
         self._find_cf_standard_name_table(ds)
         self._find_geophysical_vars(ds)
+        coord_containing_vars = ds.get_variables_by_attributes(coordinates=
+                                       lambda val: isinstance(val, basestring))
+        self.coord_data_vars = {coord_var_name for var in
+                                coord_containing_vars for
+                                coord_var_name in
+                                var.coordinates.strip().split(' ') if
+                                coord_var_name in ds.variables}
+        self.appendix_a_results = self.appendix_a_validate(ds)
 
     def check_conventions_version(self, ds):
         '''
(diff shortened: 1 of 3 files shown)


compliance_checker/cf/cf.py
```diff
@@ -130,14 +130,19 @@ def setup(self, ds):
 
         :param netCDF4.Dataset ds: An open netCDF dataset
         """
-        self._find_coord_vars(ds)
+        self.coord_vars = self._find_coord_vars(ds)
```
benjwadams: This is unused due to the distinction between "coordinate variables" (variables with a single dimension that has the same name as the variable) and "coordinate data variables", which in my interpretation appear to be variables referenced by the `coordinates` attribute in other variables.

benjwadams:
[Section 2.6.2 of CF 1.6](http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#description-of-file-contents) says:

> The NUG defines title and history to be global attributes. We wish to allow the newly defined attributes, i.e., institution, source, references, and comment, to be either global or assigned to individual variables. When an attribute appears both globally and as a variable attribute, the variable’s version has precedence.`

The "individual variables" seems to conflict with Appendix A of CF 1.6, which says that the `comment` attribute can appear in only global attributes or attributes associated with a non-coordinate variable, as mentioned in https://cf-trac.llnl.gov/trac/ticket/87 .   In the event that there seem to be contradictions within the spec, I am usually in favor of supporting the more lenient interpretation of the spec which would cause fewer reported issues.

The changes introduced for checking CF Appendix A caused the NCEI gold standard point file to fail under the CF 1.6 checker, as there were `comment` attributes in coordinate-related variables.  Changing to the `Conventions` attribute to `CF-1.7` in  66d9f94d012f33a34b445d8b10ef33175e9c1fbf fixed these issues, but appears to have broken CircleCI tests. (possible testing cache issue?)  I may want to remove the Appendix A check for `comment` under CF 1.6 entirely and revert the aforementioned commit.

@mhidas, any thoughts on this?

P.S. Reading and knowing what I know now, I'd move this PR to draft status if I could in GitHub, but I'd have to recreate the PR entirely to do so.   I will probably squash+rebase if we decide to remove some of the behavior for checking comments, etc.

compliance_checker/cf/cf.py
```diff
@@ -627,6 +749,48 @@ class CF1_6Check(CFNCCheck):
...
+                  'add_offset': {'Type': 'N', 'attr_loc': {'D'}, 'cf_section': '8.1'},
+                  'ancillary_variables': {'Type': 'S', 'attr_loc': {'D'}, 'cf_section': '3.4'},
+                  'axis': {'Type': 'S', 'attr_loc': {'C'}, 'cf_section': '4'},
+                  'bounds': {'Type': 'S', 'attr_loc': {'C'}, 'cf_section': '7.1'},
+                  'calendar': {'Type': 'S', 'attr_loc': {'C'}, 'cf_section': '4.4.1'},
+                  'cell_measures': {'Type': 'S', 'attr_loc': {'D'}, 'cf_section': '7.2'},
+                  'cell_methods': {'Type': 'S', 'attr_loc': {'D'}, 'cf_section': '7.3'},
+                  'cf_role': {'Type': 'C', 'attr_loc': {'C'}, 'cf_section': '9.5'},
+                  'climatology': {'Type': 'S', 'attr_loc': {'C'}, 'cf_section': '7.4'},
+                  'comment': {'Type': 'S', 'attr_loc': {'G', 'D'}, 'cf_section': '2.6.2'},
```
benjwadams: TODO: possibly comment out?

benjwadams:
cc: @mwengren, @ocefpaf

mhidas:
> In the event that there seem to be contradictions within the spec, I am usually in favor of supporting the more lenient interpretation of the spec which would cause fewer reported issues.

Me too! :+1: 

> I may want to remove the Appendix A check for comment under CF 1.6 entirely

:+1: That sounds good to me. I can't imagine any situation where a coordinate variable having a `comment` attribute would be a problem. The only thing to check on a `comment` attribute (wherever it is) is that it's a string.

I don't really know how CF changes are managed, but https://cf-trac.llnl.gov/trac/ticket/87 was labelled as a "defect" and marked as "acceted" and "fixed" 3 years ago. I would have liked that to mean the change was applid retrospectively to CF-1.6 too, but it seems it wasn't.

benjwadams:
Going to try recreating this PR due to AppVeyor failures I'm unable to explain (possibly due to squash + force push)?

PR #694, opened 2019-11-12
Check Appendix A attributes for CF 1.6 and 1.7

benjwadams:
Adds a check for CF 1.6 and 1.7 attributes, as well as a general
framework to execute these checks.

Practically the same as #690, recreating to see if AppVeyor failures can  be prevented.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -145,7 +145,6 @@ class CFBaseCheck(BaseCheck):
                                 coord_var_name in
                                 var.coordinates.strip().split(' ') if
                                 coord_var_name in ds.variables}
-        self.appendix_a_results = self.appendix_a_validate(ds)
 
     def check_conventions_version(self, ds):
         '''
@@ -685,7 +684,7 @@ class CFBaseCheck(BaseCheck):
         else:
             return standard_name, None
 
-    def appendix_a_validate(self, ds):
+    def check_appendix_a(self, ds):
         """
         Validates a CF dataset against the contents of its Appendix A table for
         attribute types and locations. Returns a list of results with the
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -138,6 +138,14 @@ class CFBaseCheck(BaseCheck):
         self._find_metadata_vars(ds)
         self._find_cf_standard_name_table(ds)
         self._find_geophysical_vars(ds)
+        coord_containing_vars = ds.get_variables_by_attributes(coordinates=
+                                       lambda val: isinstance(val, basestring))
+        self.coord_data_vars = {coord_var_name for var in
+                                coord_containing_vars for
+                                coord_var_name in
+                                var.coordinates.strip().split(' ') if
+                                coord_var_name in ds.variables}
+        self.appendix_a_results = self.appendix_a_validate(ds)
 
     def check_conventions_version(self, ds):
         '''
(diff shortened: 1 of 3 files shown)


compliance_checker/suite.py
```diff
@@ -359,6 +360,9 @@ def run(self, ds, skip_checks, *checker_names):
                     vals.extend(self._run_check(c, ds, max_level))
                 except Exception as e:
                     errs[c.__func__.__name__] = (e, sys.exc_info()[2])
+            if isinstance(checker, CFBaseCheck):
+                for res in checker.appendix_a_results:
+                    vals.append(res)
```
mhidas: I don't understand why this needs to be here? Why couldn't `CFBaseCheck` (or it subclass) include the appendix A checks in its output? That way `suite.py` could remain independent of any particular check suite, which I thought was the intention.

compliance_checker/cf/cf.py
```diff
@@ -3432,6 +3621,53 @@ class CF1_7Check(CF1_6Check):
...
+                  'source': {'Type': 'S', 'attr_loc': {'G', 'D'}, 'cf_section': '2.6.2'},
+                  'standard_error_multiplier': {'Type': 'N',
+                                                'attr_loc': {'D'},
+                                                'cf_section': None},
+                  'standard_name': {'Type': 'S', 'attr_loc': {'D', 'C'}, 'cf_section': '3.3'},
+                  'title': {'Type': 'S', 'attr_loc': {'G'}, 'cf_section': None},
+                  'units': {'Type': 'S', 'attr_loc': {'D', 'C'}, 'cf_section': '3.1'},
+                  'valid_max': {'Type': 'N', 'attr_loc': {'D', 'C'}, 'cf_section': None},
+                  'valid_min': {'Type': 'N', 'attr_loc': {'D', 'C'}, 'cf_section': None},
+                  'valid_range': {'Type': 'N', 'attr_loc': {'D', 'C'}, 'cf_section': None}}
```
daltonkell: Instead of copying a bunch of the 1.6 Appendix A, could you use

```
self.appendix_a.update({...})
```
with only the needed information? Or is that not feasible with this design?

compliance_checker/cf/cf.py
```diff
@@ -677,6 +685,141 @@ def _split_standard_name(self, standard_name):
         else:
             return standard_name, None
 
+    def appendix_a_validate(self, ds):
```
daltonkell: What's the reason for this now being a `check_` method? Won't this make it a little more difficult to adhere to the high-level checker interface?

benjwadams: Will move to check instead.

compliance_checker/suite.py
benjwadams: Will move to a `check_` method instead.

compliance_checker/cf/cf.py
```diff
@@ -677,6 +685,141 @@ def _split_standard_name(self, standard_name):
...
+        """
+        Validates a CF dataset against the contents of its Appendix A table for
+        attribute types and locations. Returns a list of results with the
+        outcomes of the
```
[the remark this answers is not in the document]
benjwadams: Yes

compliance_checker/cf/cf.py
benjwadams: It's possible, yes.  Will DRY things up a bit.

benjwadams:
@ocefpaf, do you know why the AppVeyor might be failing?  All tests pass when run under Python 2.7 and 3 locally.  The only thing I can think of is that AppVeyor is somehow caching some of the test netCDF files.

ocefpaf:
The failing test is:

```
================================== FAILURES ===================================
___________________ TestCLI.test_multi_checker_return_value ___________________
self = test_multi_checker_return_value (compliance_checker.tests.test_cli.TestCLI)
    def test_multi_checker_return_value(self):
        '''
        Tests that any failure for multiple checkers results in a failure return
        status
        '''
        # CF should pass here
        return_value, errors = ComplianceChecker.run_checker(
            ds_loc=STATIC_FILES['ncei_gold_point_1'],
            verbose=0,
            criteria='strict',
            checker_names=['cf:1.6'],
            output_filename=self.path,
            output_format='text'
        )
>       self.assertTrue(return_value)
E       AssertionError: False is not true
compliance_checker\tests\test_cli.py:227: AssertionError
```

We would need to check what return value we are getting there. Windows is not like *nix where we get 0 or 1. There are multiple passing and failing values.

benjwadams:
The failing test should be returning a Python boolean 
via `(all(cs.passtree(groups, limit) ...)` so should not be affected by different return status conventions, AFAICT.

benjwadams:
@ocefpaf, @Bobfrat tried out installing this branch on Windows under Conda.  We weren't able to reproduce test failures encountered in CircleCI when installing compliance-checker via Conda.  I'm going to merge this into the `cf_1_7_dev_rebase` branch and we can pursue the root cause of the test failures further prior to merging into `master`.

PR #698, opened 2019-11-15
Trac ticket 80 ogc wkt

daltonkell:


compliance_checker/cf/cf.py
```diff
@@ -3698,6 +4035,286 @@ def check_cell_measures(self, ds):
...
+        :param netCDF4.Variable var
+        :rtype tuple
+        :return two-tuple (bool, str)
+        """
+
+        msg = ("If any of reference_ellipsoid_name, prime_meridian_name, "
+               "or horizontal_datum_name are defined, all must be defined.")
+
+        if (not all([
+            x in var.ncattrs() for x in [
```
benjwadams: Store `var.ncattrs()` and do a set comparison of `{'reference_ellipsoid_name', 'prime_meridian_name', 'horizontal_datum_name'}` to avoid calling ncattrs() each time.

compliance_checker/cf/cf.py
```diff
@@ -3698,6 +4035,286 @@ def check_cell_measures(self, ds):
...
+        """
+        Evalute the condition for the geogpotential_datum_name attribute.
+
+        :param val: value to be tested
+        :rtype tuple
+        :return two-tuple of (bool, str)
+        """
+
+        conn = self._get_projdb_conn()
+        query_str = ('SELECT 1 FROM vertical_datum WHERE name = ? '
```
benjwadams: A lot of the format of these database methods is roughly the same.  You could make a helper function that takes a query string and an error message/test context and reduce the amount of repetition somewhat.

compliance_checker/cf/cf.py
```diff
@@ -3698,6 +4035,286 @@ def check_cell_measures(self, ds):
...
+
+        # must be double-precision array
+        elif val.dtype != np.float64:
+            return (False, msg)
+
+        # must be of length 3, 6, or 7
+        elif not val.shape: # single value
+            return (False, msg)
+
+        elif not any([(val.shape[0] == 3) or (val.shape[0] == 6) or (val.shape[0] == 7)]):
```
benjwadams: Using `val.size` may be somewhat easier here -- array should probably only have one dimension/be a vector.   Also use something like `val.size in (3, 6, 7)` to abridge the chained or statements.

compliance_checker/cf/cf.py
```diff
@@ -677,6 +707,90 @@ def _split_standard_name(self, standard_name):
...
+            if attr_type == 'N':
+                if not is_numeric:
+                    return [False,
+                            "{} must be numeric".format('{}')]
+            elif attr_type == 'D':
+                # TODO: handle edge case where variable is unset here
+                var_dtype = getattr(variable, 'dtype', None)
+                if (underlying_dtype != var_dtype):
+                    return [False,
+                            "{} must be numeric and must match {} dtype".format('{}', var_dtype)]
```
benjwadams: If you are really sure you want to format returning a format string, you should use `{{}}` within the string and skip the `'{}'` format argument.

compliance_checker/cf/cf.py
```diff
@@ -677,6 +707,90 @@ def _split_standard_name(self, standard_name):
...
+        if variable is None and 'G' not in attr_dict['attr_loc']:
+            raise ValueError('Non-global attributes must be associated with a '
+                             ' variable')
+        attr_str = ('Global attribute {}'.format(attr_name)
+                    if 'G' in attr_dict['attr_loc'] and variable is None
+                    else "Attribute {} in variable {}".format(attr_name,
+                                                              variable.name))
+
+        # check the type
+        return_value = self._check_attr_type(attribute, attr_type, variable)
```
benjwadams: I would just pass in `attr_str` to this function rather than reassigning the formatted value later.

compliance_checker/cf/cf.py
```diff
@@ -2412,6 +2527,189 @@ def check_reduced_horizontal_grid(self, ds):
...
+        Evaluate a condition (or series of conditions) for a particular
+        attribute. Implementation for CF-1.6.
+
+        :param attr: attribute to teset condition for
+        :param str attr_name: name of the attribute
+        :rtype tuple
+        :return two-tuple of (bool, str)
+        """
+
+        if attr_name == 'latitude_of_projection_origin':
```
benjwadams: Using a `defaultdict` or similar with functions as values can be faster than large chained if/elif structure.

compliance_checker/cf/cf.py
```diff
@@ -3698,6 +4035,286 @@ def check_cell_measures(self, ds):
...
+        :param attr: attribute to teset condition for
+        :param str attr_name: name of the attribute
+        :rtype tuple
+        :return two-tuple of (bool, str)
+        """
+
+        if attr_name == 'geographic_crs_name':
+            return self._evaluate_geographic_crs_name(attr)
+
+        elif attr_name == 'geoid_name': # TODO
```
benjwadams: What's TODO here?  If the corresponding TODO is finished, please remove it.  If it is not, indicate what needs to be done here or implement it and remove.

compliance_checker/cf/cf.py
daltonkell: I understand where you're coming from here, but I wanted to make the helper function as atomic as possible so it could be used in other functions requiring attribute type checking. I think it's a really clear way to compose the function.

compliance_checker/cf/cf.py
daltonkell: See newest commit for changes

compliance_checker/cf/cf.py
daltonkell: See newest commit for changes

compliance_checker/cf/cf.py
daltonkell: See newest commit for changes

compliance_checker/cf/cf.py
daltonkell: See newest commit for changes

benjwadams:
Closing in favor of #704

PR #699, opened 2019-11-18
Use OWSLib < 0.19.0 for Python 2 support

daltonkell:
OWSLib was recently updated. In order to not break the build, constrain the version.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,4 +1,4 @@
-OWSLib>=0.8.3
+OWSLib>=0.8.3,<0.19.0
 lxml>=3.2.1
 cf_units>=2
 requests>=2.2.1


requirements.txt
```diff
@@ -1,4 +1,4 @@
-OWSLib>=0.8.3
+OWSLib>=0.8.3,<0.19.0
```
ocefpaf: If the problem is only `Python 2` how about:

OWSLib>=0.8.3,<0.19.0; python_version <= '2'
OWSLib>=0.8.3; python_version > '2'

instead so we don't penalize Python 3 users?

benjwadams: Argh, I hadn't considered this in my haste to unbreak the build from the OWSLib 0.19.0 changes.  We can create another PR so that Py3 users can use the latest version.

daltonkell: Good idea, I'll create the PR  so we can get that in.

PR #700, opened 2019-11-18
Only restrict OWSLib version upper bound for Python 2

daltonkell:


.travis.yml
```diff
@@ -42,7 +42,7 @@ matrix:
 before_install:
   # Install miniconda and create TEST env.
   - |
-    wget http://bit.ly/miniconda -O miniconda.sh
+    wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Linux-x86_64.sh -O miniconda.sh
```
ocefpaf: That `bit.ly` URL is stale. Up-to-date examples on how to set CIs up can be found in https://github.com/ioos/ioos-python-package-skeleton/blob/master/.travis.yml#L26

benjwadams:
@ocefpaf, thanks for the work on fixing the tests.  I saw some of the commit messages and am definitely in favor of removing backwards compatibility with Python 2.x after the CF 1.7 release has been made.

requirements.txt
```diff
@@ -1,4 +1,5 @@
-OWSLib>=0.8.3,<0.19.0
+OWSLib>=0.8.3,<0.19.0; python_version <= '2' #conda: OWSLib>=0.8.3,<0.19.0 (only python=2)
+OWSLib>=0.8.3; python_version > '2' #conda: OWSLib>=0.19.0 (only python=3)
```
ocefpaf: This case was not covered to I had to code it in `gen_conda_requirements.py`, but I cannot wait for `compliance-checker` to drop Python 2.7!

benjwadams: Same.  For me, wrangling with Unicode issues while supporting both Python 2 and 3 was quite the headache back in the day.

benjwadams:
@ocefpaf, good to merge this into master?  I'm going to also merge into the CF 1.7 branch once we do that.

ocefpaf:
> @ocefpaf, good to merge this into master? I'm going to also merge into the CF 1.7 branch once we do that.

Yep. No more commits on my part.

PR #704, opened 2019-11-19
Grid mapping attr check rebase

daltonkell:
These three commits abstract the attribute type checking, add checks for grid_mapping attributes, and provide a small fixup prior to rebasing.


daltonkell:
This addresses [Trac Ticket 80](https://cf-trac.llnl.gov/trac/ticket/80) and should be used instead of #698.

PR #710, opened 2019-12-02
Removes old PDF presentation on compliance-checker

benjwadams:


benjwadams:
Sorry @daf.  :frowning_face:   Some of the mechanisms in Compliance Checker have now changed since when this presentation was written.  There's also support for more checkers, both core and through plugin checkers, so in total the information included in this presentation might be a bit misleading compared with the `README`.  If this presentation has sentimental value, perhaps it'd be best to archive it somewhere?

PR #713, opened 2020-01-03
Correct coordinate data variable finding code

benjwadams:
Corrects code which finds coordinate data variables to include both NUG
coordinate variables (variables with one dimension where the name of the
dimension and variable are the same) and existing variables in the
dataset which are referred to by other variables' "coordinates"
attribute.

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -132,7 +132,7 @@ class CFBaseCheck(BaseCheck):
 
         :param netCDF4.Dataset ds: An open netCDF dataset
         """
-        self._find_coord_vars(ds)
+        self.coord_vars = self._find_coord_vars(ds)
         self._find_aux_coord_vars(ds)
         self._find_ancillary_vars(ds)
         self._find_clim_vars(ds)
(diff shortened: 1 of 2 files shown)


compliance_checker/cf/cf.py
```diff
@@ -142,11 +142,28 @@ def setup(self, ds):
...
                                        lambda val: isinstance(val, basestring))
-        self.coord_data_vars = {coord_var_name for var in
-                                coord_containing_vars for
-                                coord_var_name in
-                                var.coordinates.strip().split(' ') if
-                                coord_var_name in ds.variables}
+
+        # coordinate data variables
+
+        # Excerpt from "§1.3 Overview" on coordinate data
```
benjwadams: See below for rationale

compliance_checker/cf/cf.py
```diff
@@ -132,7 +132,7 @@ def setup(self, ds):
 
         :param netCDF4.Dataset ds: An open netCDF dataset
         """
-        self._find_coord_vars(ds)
+        self.coord_vars = self._find_coord_vars(ds)
```
daltonkell: Was the previous call `self._find_coord_vars` without the assignment assigning the coordinate variables internally?

benjwadams: Looks like the function has a side effect and sets the `_coord_vars` variable in addition to returning the list of variables from the function.  The function is called several times elsewhere -- I'd prefer to avoid this side effect, and keep the object state changes in `setup` and then use `self.coord` vars instead of the function in aforementioned places.

PR #724, opened 2020-02-04
grid mapping false east/north optional; fixed #716

neumannd:
Attributes false_easting and false_northing are now optional for grid
mapping variables. Previously, they were mandatory. The `optional` is
in agreement with CF-1.8. In CF version prior to 1.8 it was ambiguous
whether these attributes were mandatory or not.

Updated cf grid mapping test according to changes.

PR #730, opened 2020-02-08
grid mapping false east/north optional; fixes #716

### LINKED ISSUE #716: CF Grid Mapping Required Attributes
The CF-1.7 checker incorrectly requires all attributes listed when only some are applicable. 

[Appendix F](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#appendix-grid-mappings) describes the grid mapping specifications:

> Each recognized grid mapping is described in one of the sections below. Each section contains: the valid name that is used with the grid_mapping_name attribute; a list of the specific attributes that may be used to assign values to the mapping’s parameters; the standard names used to identify the coordinate variables that contain the mapping’s independent variables; and references to the mapping’s definition or other information that may help in using the mapping. Since the attributes used to set a mapping’s parameters may be shared among several mappings, their definitions are contained in a table in the final section. The attributes which describe the ellipsoid and prime meridian may be included, when applicable, with any grid mapping.

The operative phrase is
> a list of the specific attributes that __may__ be used to assign values to the mapping’s parameters

The checker's incorrect behavior is illustrated in the following example, testing the `lambert_conformal_conic` grid mapping (output shown):

```
§5.6 Horizontal Coorindate Reference Systems, Grid Mappings, Projections
* false_easting is a required attribute for grid mapping lambert_conformal_conic
* false_northing is a required attribute for grid mapping lambert_conformal_conic
```
The spec indicates that the attributes `false_easting` and `false_northing` are valid attributes but not *required* attributes (hence the "...__may__ be used to assign values to the mapping's parameters...").

The correct behavior should only test the conditions for the attributes (found in Table F.1) if the attributes exist.

neumannd:
* grid mapping false east/north optional; fixes #716: Attributes false_easting and false_northing are now optional for grid mapping variables. Previously, they were mandatory. The `optional` is in agreement with CF-1.8. In CF version prior to 1.8 it was ambiguous whether these attributes were mandatory or not.
* added projections_*_coordinate to new projections
* updated grid mapping test according to changes

### DIFF
--- a/compliance_checker/cf/appendix_f.py
+++ b/compliance_checker/cf/appendix_f.py
@@ -280,6 +280,10 @@ grid_mapping_dict17.update({
         (
             'false_easting',
             'false_northing'
+        ),
+        (
+            'projection_x_coordinate',
+            'projection_y_coordinate'
         )
     ],
     'oblique_mercator': [
@@ -292,6 +296,10 @@ grid_mapping_dict17.update({
         (
             'false_easting',
             'false_northing'
+        ),
+        (
+            'projection_x_coordinate',
+            'projection_y_coordinate'
         )
     ],
     'sinusoidal': [
@@ -301,6 +309,10 @@ grid_mapping_dict17.update({
         (
             'false_easting',
             'false_northing'
+        ),
+        (
+            'projection_x_coordinate',
+            'projection_y_coordinate'
         )
     ]
 })
--- a/compliance_checker/cf/appendix_f.py
+++ b/compliance_checker/cf/appendix_f.py
@@ -81,11 +81,12 @@ grid_mapping_dict16 = {
     'albers_conical_equal_area': [
         (
             'longitude_of_central_meridian',
-            'latitude_of_projection_origin',
+            'latitude_of_projection_origin'
+        ),
+        (
             'false_easting',
             'false_northing'
         ),
-        (),
         (
             'projection_x_coordinate',
             'projection_y_coordinate'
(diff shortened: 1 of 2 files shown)


neumannd:
The definitions of the three new projections of CF-1.7 seemed to be incomplete because the definition of `projections_x_coordinate` and `projections_y_coordinate` where missing (third entry of the lists defining the projections). I didn't want to create an extra pull request for this minor issue.

neumannd:
replaces #724

benjwadams:
@neumannd, I'm just getting around to reviewing PRs/issues after the IOOS 1.2 checker concept was released.  Are the changes to Appendix F in 1.8 retroactive to CF <= 1.7?

neumannd:
@benjwadams Depends. In my opinion it was implicitely assumed that `false_easting` and `false_northing` were optional in CF <= 1.7 although it was not explicitely stated. The [example 5.7](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#lambert-conformal-projection) in the CF Conventions never used `false_*`. Therefore, I would consider the changes in Appendix F to be retroactive. Similarly, @daltonkell described it in issue #716 .

benjwadams:
I haven't forgot about this, but I'm probably aiming to merge this in once there is a CF-1.8 checker in progress.

daltonkell:
@benjwadams I think it would be okay to move along with this even though we aren't yet working explicitly on a 1.8 release

benjwadams (approved):

PR #729, opened 2020-02-06
generate netCDF4 files from cdl by default

neumannd:
fixes #725:

* generate netCDF4 file from cdl with ncgen
* first try gen netCDF4; fallback Classic
* new cdl test file with netCDF4 features
* two new test for netCDF4 file generation

Background: When a CDL file is provided, currently, a netCDF classic file is created by ncgen. However, a CDL file might contain netCDF4 features such as attributes/variables of type int64 (indicated by `LL` after numbers). Therefore, the `ncgen`-call that creates a netCDF file was modified to create a netCDF4 instead of a netCDF-classic file.

PR #731, opened 2020-02-09
generate netCDF4 file from cdl file by default and corresponding tests

neumannd:
* generate netCDF4 file from cdl with ncgen: When a CDL file is provided, currently, a netCDF classic file is created by ncgen. However, a CDL file might contain netCDF4 features such as attributes/variables of type int64 (indicated by `LL` after numbers). Therefore, the `ncgen`-call that creates a netCDF file was modified to create a netCDF4 instead of a netCDF-classic file. If this does not work, we try to create a netCDF Classic file. If that does not work, we exit the program.
* added cdl file with netcdf4 features for testing
* two new test for netCDF4 file generation

### DIFF
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -695,9 +695,22 @@ class CheckSuite(object):
             ds_str = cdl_path.replace('.cdl', '.nc')
         else:
             ds_str = cdl_path + '.nc'
-        returncode = subprocess.call(['ncgen', '-o', ds_str, cdl_path])
-        if returncode != 0:
-            print("Error generating dataset from .cdl")
+
+        # generate netCDF-4 file
+        iostat = subprocess.run(['ncgen', '-k', 'nc4', '-o', ds_str, cdl_path], capture_output=True)
+        if iostat.returncode != 0:
+          # if not successfull, create netCDF classic file
+          print('netCDF-4 file could not be generated from cdl file with ' +
+                'message:')
+          print(iostat.stderr)
+          print('Trying to create netCDF Classic file instead.')
+          iostat = subprocess.run(['ncgen', '-k', 'nc3', '-o', ds_str, cdl_path], capture_output=True)
+          if iostat.returncode != 0:
+            # Exit program if neither a netCDF Classic nor a netCDF-4 file
+            # could be created.
+            print('netCDF Classic file could not be generated from cdl file' +
+                  'with message:')
+            print(iostat.stderr)
             sys.exit(1)
         return ds_str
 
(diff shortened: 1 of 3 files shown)


neumannd:
This pull request superseds #729

neumannd:
The Travis-CI Python 2.7 test failed. Is that critical because the Python 2.7 support should be dropped in future?

compliance_checker/suite.py
```diff
@@ -695,9 +695,22 @@ def generate_dataset(self, cdl_path):
...
+                'message:')
+          print(iostat.stderr)
+          print('Trying to create netCDF Classic file instead.')
+          iostat = subprocess.run(['ncgen', '-k', 'nc3', '-o', ds_str, cdl_path], capture_output=True)
+          if iostat.returncode != 0:
+            # Exit program if neither a netCDF Classic nor a netCDF-4 file
+            # could be created.
+            print('netCDF Classic file could not be generated from cdl file' +
+                  'with message:')
+            print(iostat.stderr)
```
daltonkell: The `capture_output` keyword is only available in Python 3.7+, but this Compliance Checker needs to support Python 3.5+. If you supply `subprocess.PIPE` as the `stderr` keywrod, you'll be able to keep almost everything the same but maintain compatibility:

```
iostat = subprocess.run(['ncgen', '-k', 'nc4', '-o', ds_str, cdl_path], stderr=subprocess.PIPE)
...
print(iostat.stderr.decode()) # stored as byte string
...
```

If the output is not captured, an exception is thrown upon failure. At first I wanted to use this functionality and wrap in try-except statements, but I think this is a better idea. Pinging @benjwadams to see if he has any input here.

compliance_checker/tests/data/test_cdl_nc4_file.cdl
```diff
@@ -0,0 +1,40 @@
...
+		lat:standard_name = "latitude" ;
+		lat:units = "degrees_north" ;
+		lat:axis = "Y" ;
+
+// global attributes:
+		:title = "A test file" ;
+		:history = "2020-02-04T18:00 manually create by Daniel Neumann, DKRZ" ;
+		:global_att_of_type_int = 2 ;
+		:global_att_of_type_int64 = 2LL ;
+		:Conventions = "CF-1.7" ;
```
daltonkell: Awesome! Thanks for your help and input on the CF 1.7 additions as well.

compliance_checker/suite.py
neumannd: Thanks for the info. I updated the code accordingly.

daltonkell:
@benjwadams I think this is a valuable PR that addresses an issue not thought about much. Any comments?

PR #739, opened 2020-02-14
Splitter validate

benjwadams:
Allows splitting of components.

Adds CSV check for contributor_url.

Changes interface for `attr_check` to only allow

### DIFF
--- a/compliance_checker/base.py
+++ b/compliance_checker/base.py
@@ -16,8 +16,11 @@ from collections import defaultdict
 from lxml import etree
 import sys
 import re
+import csv
+from io import StringIO
 from email.utils import parseaddr
 import validators
+import itertools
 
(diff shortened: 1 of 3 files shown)


compliance_checker/base.py
```diff
@@ -32,30 +35,63 @@ def get_namespaces():
...
+
+    Returns:
+    --------
+    list of str
+        A flattened list from the CSV processing contents
+    """
+    csv_contents = csv.reader(StringIO(input_string))
+    return list(itertools.chain.from_iterable(csv_contents))
+
 class ValidationObject(object):
```
daltonkell: If this is intended to be an abstract base class, it might be beneficial to inherit from `abc.ABC ` or similar. Not critical.

benjwadams: noted

PR #749, opened 2020-02-26
Fix variable print-out bug

daltonkell:
The `gts_ingest` check accidentally prints out the `netCDF4` string representation of the variables it's looking at instead of the name:

```
                               Highly Recommended                               
gts_ingest variable
* The attribute "gts_ingest" of variable "<class 'netCDF4._netCDF4.Variable'>
float64 s.air_temperature(s)
    _ChunkSizes: [240760      1]
    _FillValue: -9999.0
    actual_range: [ 2.16 34.8 ]
    ancillary_variables: air_temperature_qc_agg air_temperature_qc_tests
    gts_ingest: true
    id: 1000735
    ioos_category: Other
    long_name: Air Temperature
    missing_value: -9999.0
    platform: station
    standard_name: air_temperature
    standard_name_uri: http://mmisw.org/ont/cf/parameter/air_temperature
```

This commit uses the name instead:

```
gts_ingest variable
* The attribute "gts_ingest" of variable "s.air_temperature"  must fulfill the following:
  - must be a string with value "true" or "false";
  - have a valid CF standard_name attribute (already checked);
  - have an ancillary variable reqpresenting QARTOD aggregate flags;
  - have a valid udunits units attribute
```

### DIFF
--- a/compliance_checker/ioos.py
+++ b/compliance_checker/ioos.py
@@ -962,7 +962,7 @@ class IOOS1_2Check(IOOSNCCheck):
                    " must fulfill the following:\n"
                    "  - must be a string with value \"true\" or \"false\";\n"
                    "  - have a valid CF standard_name attribute (already checked);\n"
-                   "  - have an ancillary variable reqpresenting QARTOD aggregate flags;\n"
+                   "  - have an ancillary variable representing QARTOD aggregate flags;\n"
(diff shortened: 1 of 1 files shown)


daltonkell:
Will push an additional commit to fix the misspelling of "representing" above.

Issue #747, opened 2020-02-26
IOOS:1.2 redundant messages


daltonkell:
When calling the `ioos:1.2` checker on this [SECOORA dataset](http://erddap.secoora.org/erddap/info/edu_usf_marine_comps_1407d550/index.html), the following messages  appear:

```
platform
* Gridded model datasets are not required to declare a platform
* The global attribute "platform" must be a single string containing no blank characters; it is None

...

contributor_role
* contributor_role should be from NERC or GEOIDE
* contributor_role not present

contributor_role_vocabulary
* contributor_role_vocabulary should be one of NERC or GEOIDE
* contributor_role_vocabulary not present

...

ioos_ingest
* Global attribute "ioos_ingest" must be a string with value "false"
* ioos_ingest not present
```

If a message describes what an attribute *should* be, it is redundant to also display that the attribute is not present in a second message

mwengren:
Resolved by #750, closing.

PR #750, opened 2020-02-26
Ioos  1 2  redundant msgs

daltonkell:
Remove redundant messages for `contributor_role`, `contributor_role_vocabulary`, and `ioos_ingest`. Each has an explicit check, so they are not needed in the recommended attributes.


daltonkell:
References #747

PR #752, opened 2020-03-02
Fix lack of messages where ValidatorObject is used

daltonkell:
Some metadata fields like "contributor_email" were not being
populated with messages; fix this to the desired behavior.

Example

Current behavior:

```
                                  Recommended                                   
--------------------------------------------------------------------------------
contributor_email

contributor_name
* contributor_name not present

contributor_role
* contributor_role should be from NERC or GEOIDE
* contributor_role not present

contributor_role_vocabulary
* contributor_role_vocabulary should be one of NERC or GEOIDE
* contributor_role_vocabulary not present

contributor_url

```

Improved behavior:

```
                                  Recommended                                   
--------------------------------------------------------------------------------
contributor_email
* contributor_email not present

contributor_name
* contributor_name not present

contributor_role
* contributor_role should be from NERC or GEOIDE
* contributor_role not present

contributor_role_vocabulary
* contributor_role_vocabulary should be one of NERC or GEOIDE
* contributor_role_vocabulary not present

contributor_url
* contributor_url not present
```

PR #753, opened 2020-03-03
Change GEOIDE to NOAA-NCEI

daltonkell:
Addresses https://github.com/ioos/compliance-checker/pull/752#issuecomment-593609604

### DIFF
--- a/compliance_checker/ioos.py
+++ b/compliance_checker/ioos.py
@@ -594,27 +594,27 @@ class IOOS1_2Check(IOOSNCCheck):
         vocb_msg = "contributor_role_vocabulary should be one of NERC or NOAA-NCEI"
 
         if role:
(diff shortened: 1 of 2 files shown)


compliance_checker/ioos.py
```diff
@@ -590,38 +590,38 @@ def check_contributor_role_and_vocabulary(self, ds):
...
-                "pointOfContact",
-                "principalInvestigator",
-                "processor",
-                "publisher",
-                "resourceProvider",
-                "rightsHolder",
-                "sponsor",
-                "stakeholder",
-                "user"
+                "author",                # both
```
benjwadams: Is it necessary to have `# both` after each line here?  I think this could be eliminated without losing meaning.

PR #761, opened 2020-03-11
Update CF standard name table to latest version 72

Bobfrat:
Haven't synced this up in a while and this particular version has some IOOS contributions, notably the QARTOD quality flag names.


mwengren:
I was wondering why my CC reports kept on showing Standard Names table v49.  Thanks!

Issue #762, opened 2020-03-11
Add syntax check for 'platform_id'


mwengren:
I updated guidance for the `platform_id` attribute as follows:

- `platform_id` should be a single alphanumeric string with no blank spaces

Can we add a check to verify the attribute contents meet the restriction, similar to `wmo_platform_code` and `platform` attributes?

PR #765, opened 2020-03-13
Ensure platform_id is an alphanumeric string

daltonkell:
This addresses #762 quite succinctly, but I'm worried that the error message at 

https://github.com/ioos/compliance-checker/blob/54b5300663b70e586ca429e2ecdad12f0e155314/compliance_checker/base.py#L488

is going to be quite unclear to some users. I'm going to ping for @benjwadams for some insight. The machinery is already here, which makes a pull request this simple so tempting.

### DIFF
--- a/compliance_checker/ioos.py
+++ b/compliance_checker/ioos.py
@@ -466,7 +466,7 @@ class IOOS1_2Check(IOOSNCCheck):
             'instrument',
             'ioos_ingest',
             'keywords',
-            'platform_id',
+            ('platform_id', re.compile(r"^\w+$")), # alphanumeric only
             'publisher_address',
             'publisher_city',
             'publisher_name',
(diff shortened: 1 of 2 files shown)


benjwadams:
I might opt to get rid of regex object processing entirely.  If you want to go the regex route for this particular checker you may want to create a class that subclasses `base.RegexValidator` so that you can make a human-readable error message.

https://github.com/ioos/compliance-checker/blob/master/compliance_checker/ioos.py#L357-L359

Alternatively, you subclass `ValidationObject` and use some of the built-in string methods.

daltonkell:
@benjwadams I implemented your first suggestion.

compliance_checker/ioos.py
```diff
@@ -358,6 +358,10 @@ class IOOS1_2_ConventionsValidator(base.RegexValidator):
     validator_regex = r"\bIOOS-1.2\b"
     validator_fail_msg = "{} must contain the string \"IOOS 1.2\""
 
+class IOOS1_2_PlatformIDValidator(base.RegexValidator):
+    validator_regex = r"^\w+$"
```
benjwadams: This is almost right if taking the regex approach, but `\w` also includes underscores.

From Python docs:

>  \w
>
>    Matches any alphanumeric character; this is equivalent to the class [a-zA-Z0-9_].

So use something like `[a-zA-Z0-9]` in place of `\w`

Issue #760, opened 2020-03-09
Check of http://testing.erddap.axds.co/erddap/tabledap/org_cormp_sun2 produces incorrect results


mwengren:
Running the following:
```
$ compliance-checker --version
IOOS compliance checker version 4.3.3rc1+11.g3a7c3cd
$ compliance-checker -t ioos -f html http://testing.erddap.axds.co/erddap/tabledap/org_cormp_sun2 > org_cormp_sun2-ioos.html
```

Produces unexpected results.   

For details of the test output, see: 
https://gist.github.com/mwengren/d0a921b5b83661b0098437bfefc05701 

and dataset attribution can be found here:
http://testing.erddap.axds.co/erddap/tabledap/org_cormp_sun2.das

Test| Message 
--------|----------
gts_ingest | gts_ingest 
platform | The global attribute "platform" must be a single string containing no blank characters; it is buoy
wmo_platform_code | The wmo_platform_code must be a numeric string of 5 or 7 characters

Explanation: 

- `platform` and `wmo_platform_code` appear to be correct in the dataset
- `gts_ingest` message is unlcear

@jessicaaustin

mwengren:
Also, I believe `wmo_platform_code` can be alphanumeric.  @kbailey-noaa can you confirm this?

fgayanilo:
correct me if otherwise; wmo is purely numeric as per their syntax (https://www.wmo.int/pages/prog/amp/mmop/wmo-number-rules.html). I have seen a couple of urn implementations that use some local labels and tag them as wmo. I think this should be avoided when possible.

mwengren:
ok, thanks @fgayanilo, I was hoping someone with better knowledge would interject.   I'm assuming Kevin O'Brien has some insight, but I lost his GitHub handle...

kbailey-noaa:
@mwengren @fgayanilo What we're calling "WMO ID" (i.e. wmo_platform_code) for the IOOS Metadata Profile includes both the WMO ID for buoys (numeric, 5 digits), WMO ID for gliders (numeric, 7 digits), and the NWS ID (alphanumeric, 5 digits).  These IDs are all used by NDBC for delivering data to the GTS.

mwengren:
@kbailey-noaa thanks!  So therefore the Compliance Checker test should accept alphanumeric IDs of 5 digits, and numeric IDs of 7 digits only?  Just making sure we're testing with the proper criteria. 

cc @daltonkell

[...]

PR #766, opened 2020-03-16
Ensure wmo_platform_code conforms to specification

daltonkell:
Addresses #760 (in part, see bottom of discussion; other parts are being addressed separately). Ensure the `wmo_platform_code` conforms to the specification of

  - numeric string of 5 or 7 characters
  - alphanumeric strings of 5 characters

This is corresponds with the [specification](https://ioos.github.io/ioos-metadata/ioos-metadata-profile-v1-2.html#platform) (see "wmo_platform_code" in the table).

### DIFF
--- a/compliance_checker/ioos.py
+++ b/compliance_checker/ioos.py
@@ -1063,7 +1063,7 @@ class IOOS1_2Check(IOOSNCCheck):
                isinstance(code, str) and
                (
                    (code.isnumeric() and (len(code)==5 or len(code)==7)) or
-                   re.search(r"^\w{5}$", code)
+                   re.search(r"^(?:[a-zA-Z0-9]{5}|[0-9]{7})$", code)
                )
            ):
                valid = False
--- a/compliance_checker/ioos.py
+++ b/compliance_checker/ioos.py
@@ -1062,7 +1062,6 @@ class IOOS1_2Check(IOOSNCCheck):
            if not (
                isinstance(code, str) and
                (
-                   (code.isnumeric() and (len(code)==5 or len(code)==7)) or
                    re.search(r"^(?:[a-zA-Z0-9]{5}|[0-9]{7})$", code)
                )
            ):
--- a/compliance_checker/ioos.py
+++ b/compliance_checker/ioos.py
@@ -1034,9 +1034,16 @@ class IOOS1_2Check(IOOSNCCheck):
 
     def check_wmo_platform_code(self, ds):
         """
-        If a WMO Platform Code is given as a global variable, check that it is
-        well-formed. According to the WMO, valid codes are a numeric string
-        comprised of 5 or 7 characters.
+        Per the spec:
+
+        "The WMO identifier for the platform used to measure the data. This
+        identifier can be any of the following types:
+          - WMO ID for buoys (numeric, 5 digits)
+          - WMO ID for gliders (numeric, 7 digits)
+          - NWS ID (alphanumeric, 5 digits)"
+
+        This attribute is "required, if applicable" -- a warning message will
+        only show up if the attribute is present and does not conform.
 
         Args:
             ds (netCDF4.Dataset): open Dataset
@@ -1047,14 +1054,17 @@ class IOOS1_2Check(IOOSNCCheck):
 
         valid = True
         ctxt = "wmo_platform_code"
-        msg = "The wmo_platform_code must be a numeric string of 5 or 7 characters"
+        msg = ("The wmo_platform_code must be an alphanumeric string of 5 "
+               "characters or a numeric string of 7 characters")
 
         code = getattr(ds, "wmo_platform_code", None)
         if code:
            if not (
                isinstance(code, str) and
-               code.isnumeric() and
-               (5 <= len(code) <= 7)
+               (
+                   (code.isnumeric() and (len(code)==5 or len(code)==7)) or
+                   re.search(r"^\w{5}$", code)
+               )
            ):
                valid = False
 
--- a/compliance_checker/tests/test_ioos_profile.py
+++ b/compliance_checker/tests/test_ioos_profile.py
@@ -507,13 +507,18 @@ class TestIOOS1_2(BaseTestCase):
         result = self.ioos.check_wmo_platform_code(ds)
         self.assertTrue(result.value)
 
-        # non-numeric, fail
+        # alphanumeric, valid
         ds.setncattr("wmo_platform_code", "abcd1")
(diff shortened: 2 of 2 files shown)


daltonkell:
@mwengren New results from the checker with these changes:

```
Running Compliance Checker on the datasets from: ['org_cormp_sun2.ncCF']


                                    ioos:1.2                                    
      https://ioos.github.io/ioos-metadata/ioos-metadata-profile-v1-2.html      
                               Corrective Actions                               
org_cormp_sun2.ncCF has 21 potential issues


                               Highly Recommended                               
geospatial lat min/max
* Attr 'geospatial_lat_min' is missing
* Attr 'geospatial_lat_max' is missing

geospatial lon min/max
* Attr 'geospatial_lon_min' is missing
* Attr 'geospatial_lon_max' is missing

platform_name
* platform_name not present

§3.3 Standard Name
* standard_name peak_wave_period is not defined in Standard Name Table v72


                                  Recommended                                   
contributor_email
* contributor_email must be a valid email address

contributor_role
* contributor_role should be from NERC or NOAA-NCEI

contributor_role_vocabulary
* contributor_role_vocabulary should be one of NERC or NOAA-NCEI

creator_address
* creator_address not present

creator_city
* creator_city not present

creator_phone
* creator_phone not present

creator_postalcode
* creator_postalcode not present

creator_state
* creator_state not present

instrument
* instrument not present

ioos_ingest
* Global attribute "ioos_ingest" must be a string with value "false"
* ioos_ingest not present

keywords
* keywords not present

platform_id
* platform_id not present

publisher_address
* publisher_address not present

publisher_city
* publisher_city not present

publisher_phone
* publisher_phone not present

publisher_postalcode
* publisher_postalcode not present

publisher_state
* publisher_state not present
```

mwengren:
@daltonkell Those results look a lot better.  Great!


I didn't mention this as part of #760, but since your test output is above, I do have a question:

What happened to the Platform Variable check output for this dataset?  Was that modified as well in this PR?  I think we had an email discussion (not on GH or I'd refer to it), but there's an issue with what ERDDAP output format the CC requests in order to check the Platform Variable dimensionality.  When I run this same dataset from the master branch as of last week (version 4.3.3rc1+11.g3a7c3cd) for this dataset I get:


Test| Message 
---|---
platform variables | With a single platform provided, the  dimension of the cf_role variable s.station (cf_role==timeseries_id)  should also be equal to 1 (it is 7393)

Are you requesting the `.ncCF` or `.ncCFMA` formats from ERDDAP now?  This dataset should pass those checks, I think, just wondering.

daltonkell:
Nothing regarding `platform` was changed in this PR. Is this https://github.com/ioos/compliance-checker/pull/751 the conversation you're thinking about? If I recall correctly, it was conducted over both email and GH.

We're not requesting the `.ncCF` or `.ncCFMA` formats explicitly -- I've been doing local testing with the `.nc` file on my machine. I've used ERDDAP directly before though, I'll see if I can dig up how I did it.

daltonkell:
Additionally -- aren't the `.ncCF` and `.ncCFMA` extensions essentially making different datasets? If the original `.nc` file has multiple platforms, and then the `.ncCF` dataset only has one platform, isn't that altering the core of the data? I'm not really familiar with the transformations from one "format" to another.

mwengren:
In the case of  timeSeries DSG datasets, for example, ERDDAP reduces the dimensionality of variables listed in the `cdm_timeseries_variables` config setting for both the `.ncCF` and `.ncCFMA` output formats, and slices the dataset along the dimension of whatever variables has the `cf_role=timeseries_id` attribute for each of the `cdm_timeseries_variables` variables.  

It's a little easier to see by example.  For WQB-04 look at the .das endpoint:

https://pae-paha.pacioos.hawaii.edu/erddap/tabledap/WQB-04.das

then compare the 'header' versions of each .nc* output type: 
* ncHeader: https://pae-paha.pacioos.hawaii.edu/erddap/tabledap/WQB-04.ncHeader?&time%3E=2020-01-06T16:00:00Z
* ncCFHeader: https://pae-paha.pacioos.hawaii.edu/erddap/tabledap/WQB-04.ncCFHeader?&time%3E=2020-01-06T16:00:00Z
* ncCFMAHeader: https://pae-paha.pacioos.hawaii.edu/erddap/tabledap/WQB-04.ncCFMAHeader?&time%3E=2020-01-06T16:00:00Z

If you look at the `timeseries` dimension (used by all the platform-related variables like latitude, longitude, station_name etc), in the last two, you'll see it's 1, whereas for .ncHeader it's 3876 (the `row` dimension).  

Essentially .nc output in ERDDAP is a flattened table-like version of the CF DSG versions of the same dataset (.ncCF, .ncCFMA).  

I know many RAs will be using the 1.2 checker with ERDDAP datasets, so if we want to implement the Platform Variable dimension check, we'll have to use one of the .ncCF or .ncCFMA outputs to get correct results.  In this dataset, is in fact 1 not 3876.  If you leave off the `time` filter, the row dimension is actually much higher.  

But all that said, we may end up relaxing this requirement anyway.  I was just wondering why I saw different output in the check I ran.   I reserve the right to copy all this to a new issue for posterity...

daltonkell:
@mwengren We digressed a bit here -- are the above results pertaining to `wmo_platform_code` satisfactory? If so, I'll get a code review from @Bobfrat

mwengren:
Sure, looked fine to me and results are good for this test dataset.

compliance_checker/ioos.py
```diff
@@ -1047,14 +1054,17 @@ def check_wmo_platform_code(self, ds):
...
 
         code = getattr(ds, "wmo_platform_code", None)
         if code:
            if not (
                isinstance(code, str) and
-               code.isnumeric() and
-               (5 <= len(code) <= 7)
+               (
+                   (code.isnumeric() and (len(code)==5 or len(code)==7)) or
+                   re.search(r"^\w{5}$", code)
```
benjwadams: `\w` includes underscores, which for the NWS identifier do not form a valid identifier.

Note you can also use `str.isalnum`, but it includes a number of Unicode characters which probably wouldn't be desired.

`isnumeric` and related methods apparently treat Unicode fractions as truthy!  I imagine this is not really what we want, but perhaps we should revisit this later.
https://docs.python.org/3/library/stdtypes.html#str.isnumeric

e.g.  `"ü¾".isalnum()` returns True.
So you might just want to use regex character classes here as so:

`re.search(r"^(?:[a-zA-Z0-9]{5}|[0-9]{7})$", code)`

Issue #754, opened 2020-03-07
Change 'standard_name_uri' to 'standard_name_url'


mwengren:
Compliance Checker 4.3.3rc1 checks for `standard_name_uri`.  We've recently changed this in the profile to `standard_name_url`.  Can we update the code accordingly before next release or rc?

Issue #137, opened 2015-07-02
CF valid units not in UDUNITS


mhidas:
The [CF FAQ](http://cfconventions.org/faq.html#udunits_missing)) page says:

> There are two units acceptable to CF that are not in the UDUNITS library: **sverdrup**, and **decibel** or **dB**. These have been requested for inclusion in future versions of the UDUNITS library.

However, the checker only uses UDUNITS (via the cf_units package) to validate units. So currently it will fail for these units.

Is it worth considering a work-around (until these units are added to UDUNITS)?

ocefpaf:
@mhidas Can you open an issue at https://github.com/SciTools/cf_units too?

lukecampbell:
It may not be necessary to change the core library. udunits uses an XML system to define the units, we should be able to include an auxiliary XML definition for these units.

lukecampbell:
https://github.com/ooici/ion-definitions/tree/master/res/config/units

ocefpaf:
@lukecampbell even though it is easy for us to just nudge units in I think we should try to send that upstream (to UDUNITS).  That way all the projects derived from it will have those units. Once I get back home I will try to send a PR [there](https://github.com/Unidata/UDUNITS-2).

mhidas:
@ocefpaf Logged: https://github.com/SciTools/cf_units/issues/30

kknee:
this is a UDUNITS issue, closing

Issue #134, opened 2015-06-25
Finalize and document the many to many relationship between data sets and data services


dpsnowden:
The relationship between data sets and services is unclear.  It is clearly many to many, which is fine, but the way it is implemented from different configurations of THREDDS, ERDDAP, 52North etc is unclear and uneven.  Let's clarify this once and for all and determine necessary changes for the next round of data base modifications.

dpsnowden:
Sorry, my mistake.  This was supposed to be in the ioos/catalog repo.....

PR #767, opened 2020-03-16
Change standard_name_uri to standard_name_url

daltonkell:


daltonkell:
Error message:

```
=================================== FAILURES ===================================
________________________ TestACDD1_3.test_time_extents _________________________
self = test_time_extents ( compliance_checker.tests.test_acdd:TestACDD1_3.test_time_extents )
    def test_time_extents(self):
        '''
        Test that the time extents are being checked
        '''
        result = self.acdd.check_time_extents(self.ds)
>       self.assert_result_is_good(result)
compliance_checker/tests/test_acdd.py:434: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = test_time_extents ( compliance_checker.tests.test_acdd:TestACDD1_3.test_time_extents )
result = time_coverage_extents_match (*2): False (Failed to retrieve and convert times for variables time.)
    def assert_result_is_good(self, result):
        if isinstance(result.value, bool):
>           assert result.value is True
E           AssertionError
compliance_checker/tests/__init__.py:32: AssertionError
```

This error message is also present in #766. I suspect it's due to a dependency issue, most likely `pendulum`:

https://github.com/ioos/compliance-checker/blob/54b5300663b70e586ca429e2ecdad12f0e155314/compliance_checker/acdd.py#L532-L547

The `pendulum` version was recently upgraded to `2.1.0` https://github.com/sdispater/pendulum/releases

and that's the only instance where that error message can be returned. Digging in a little deeper.

daltonkell:
## UPDATE

The `pendulum` dependency isn't the problem -- it's `cftime`.

__TL;DR__

`netCDF4.num2date()` no longer returns `datetime` instances, but instead returns instances of `cftime.Datetime`

__Details__

Testing with the same dataset provided for the failed unit test, `ncei_gold_point_2.nc`, I encounter the following:

(`cftime==1.0.3.41`)
```
In [17]: nc4.num2date(ds.variables["time"][0], ds.variables["time"].units)                                             
Out[17]: datetime.datetime(2015, 3, 25, 22, 20, 17)
```

With the upgraded version of `cftime==1.1.1`:

```
In [6]: nc4.num2date(ds.variables["time"][0], ds.variables["time"].units)                                                 
Out[6]: cftime.DatetimeGregorian(2015-03-25 22:20:17)
```

`pendulum.instance()` only takes an argument of type `datetime`:

```
(Pdb) l
535                 # num2date returns as naive date, but with time adjusted to UTC
536                 # we need to attach timezone information here, or the date
537                 # subtraction from t_min/t_max will assume that a naive timestamp is
538                 # in the same time zone and cause erroneous results.
539                 # Pendulum uses UTC by default, but we are being explicit here
540  ->             time0 = pendulum.instance(num2date(ds.variables[timevar][0],
541                                           ds.variables[timevar].units), 'UTC')
542                 time1 = pendulum.instance(num2date(ds.variables[timevar][-1],
543                                           ds.variables[timevar].units), 'UTC')
544             except:
545                 return Result(BaseCheck.MEDIUM,
(Pdb) n
> /home/dalton/compliance-checker/compliance_checker/acdd.py(541)check_time_extents()
-> ds.variables[timevar].units), 'UTC')
(Pdb) n
ValueError: instance() only accepts datetime objects.
```

daltonkell:
This behavior was adopted in version `1.1.0`:

> 2/12/2020: version 1.1.0 released. cftime.datetime instances are returned by default from num2date (instead of returning python datetime instances where possible (issue #136)). num2pydate convenience function added (always returns python datetime instances, issue #134). Fix for fraction seconds in reference date string (issue #140). Added daysinmonth attribute (issue #137).

I'll move this to a separate issue.

daltonkell:
New issue in #768

daltonkell:
Addresses #754

PR #769, opened 2020-03-16
Change num2date to num2pydate

daltonkell:
Since cftime release 1.1.0, num2date returns instances of cftime.Datetime.
In order to return datetime objects, use the convenience function
cftime.num2pydate.

### DIFF
--- a/compliance_checker/acdd.py
+++ b/compliance_checker/acdd.py
@@ -6,7 +6,7 @@ project for the verification and scoring of attributes for datasets.
 '''
 
 import numpy as np
-from netCDF4 import num2date
+from cftime import num2pydate
 from datetime import timedelta
 from compliance_checker.base import (BaseCheck, BaseNCCheck, check_has,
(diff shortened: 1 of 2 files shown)


compliance_checker/acdd.py
```diff
@@ -6,7 +6,7 @@
 '''
 
 import numpy as np
-from netCDF4 import num2date
+from cftime import num2pydate
```
ocefpaf: We don't list an explicit dependency on cftime. If we are going to rely on its API we should probably list  `cftime >=1.1.0` in the dependencies.

requirements.txt
```diff
@@ -1,3 +1,4 @@
+cftime>=1.1.0
```
ocefpaf: You were faster than me!

daltonkell: I saw I accidentally pushed before committing that, so I tried to squeeze it in before you saw it. Looks like we did it almost at the exact same time!

PR #772, opened 2020-03-17
Improve ioos_ingest check

daltonkell:
Any value supplied other than "true" or "false" should fail;
add improved error message for clarity.

Now, no more "ioos_ingest" message when testing on https://pae-paha.pacioos.hawaii.edu/erddap/tabledap/WQB-04.html:

```
$ dalton@ubuntu:~/compliance-checker$ python cchecker.py -t ioos:1.2 WQB-04.ncCF | grep "ioos_ingest"
Running Compliance Checker on the datasets from: ['WQB-04.ncCF']
```

```

### DIFF
--- a/compliance_checker/ioos.py
+++ b/compliance_checker/ioos.py
@@ -550,7 +550,7 @@ class IOOS1_2Check(IOOSNCCheck):
         If a dataset contains the global attribute ioos_ingest,
         its value must be "false". All datasets are assumed to be
         ingested except those with this flag. If the dataset should
-        be ingested, no flag should be present.
+        be ingested, no flag (or "true") should be present.
 
         Parameters
         ----------
@@ -565,12 +565,9 @@ class IOOS1_2Check(IOOSNCCheck):
         m = ("To disallow harvest of this dataset to IOOS national products, "
              "global attribute \"ioos_ingest\" must be a string with value \"false\"")
         igst = getattr(ds, "ioos_ingest", None)
-        if igst is not None: # if not supplied, return default pass
-            if isinstance(igst, str):
-                if igst.lower() not in ("true", "false"):
-                    r = False
-            else:
-                r = False
+        if (isinstance(igst, str) and igst.lower() not in ("true", "false")) or (
+            not isinstance(igst, str) and igst is not None):
+            r = False
 
         return Result(BaseCheck.MEDIUM, r, "ioos_ingest", [m])
 
--- a/compliance_checker/ioos.py
+++ b/compliance_checker/ioos.py
@@ -565,8 +565,11 @@ class IOOS1_2Check(IOOSNCCheck):
         m = ("To disallow harvest of this dataset to IOOS national products, "
              "global attribute \"ioos_ingest\" must be a string with value \"false\"")
         igst = getattr(ds, "ioos_ingest", None)
-        if igst is not None:
-            if igst not in ("true", "false"):
+        if igst is not None: # if not supplied, return default pass
+            if isinstance(igst, str):
+                if igst.lower() not in ("true", "false"):
+                    r = False
+            else:
                 r = False
 
         return Result(BaseCheck.MEDIUM, r, "ioos_ingest", [m])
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: compliance_checker/tests/test_ioos_profile.py [hunk 1]
```diff
@@ -707,10 +707,14 @@ def test_check_ioos_ingest(self):
...
-        # value anything but false
+        # value true
         ds.setncattr("ioos_ingest", "true")
-        self.assertFalse(self.ioos.check_ioos_ingest(ds).value)
+        self.assertTrue(self.ioos.check_ioos_ingest(ds).value)
+
+        # anything else should fail
         ds.setncattr("ioos_ingest", 0)
         self.assertFalse(self.ioos.check_ioos_ingest(ds).value)
+        ds.setncattr("ioos_ingest", "True")
```
Bobfrat: So to be sure, this attribute is case sensitive?

daltonkell: The spec does say "false":

https://ioos.github.io/ioos-metadata/ioos-metadata-profile-v1-2.html#ioos-ingest

I thought about making it case-insensitive - @mwengren, any input on this?

mwengren:
I don't know, my feeling is it should be case insensitive, but it should match rules for `gts_ingest`.  I don't think we specify in the guidance one way or the other.

Is there any other precedent?

Bobfrat:
I tend to agree but this attribute is specifically for NDBC right? Do we know if they are expecting it to be case sensitive?

daltonkell:
There are plenty of scenarios where we look for case-insensitive values, e.g.

https://github.com/ioos/compliance-checker/blob/5ff9f248f38f17d17f08d9787fc1fa4b240bea34/compliance_checker/cf/cf.py#L3804-L3805

These are camel-case in the spec

mwengren:
Let's use the same case-sensitivity rules for both `ioos_ingest` and `gts_ingest` whichever way we go.  

@kbailey-noaa @jessicaaustin: do either of you happen to recall if it matters to NDBC case of `gts_ingest` flag?

Unless they've told us otherwise, I think we should have the checker allow either.   

In terms of precedent, I was more wondering what's out there in the wild in terms of `True|true` and `False|false` string attribution customs in netCDF files.

benjwadams:
I can't merge this until I know the desired behavior WRT case-sensitivity.

Bobfrat:
that and the merge conflict...

jessicaaustin:
I would follow whatever the convention is for the rest of the attributes. If typically things are case-insensitive, then these can be too. 

We have no idea if NDBC checks are case-insensitive or not. But [ERDDAP search definitely doesn't care](http://erddap.cencoos.org/erddap/search/index.html?page=1&itemsPerPage=1000&searchFor=GTS_INGEST%3Dtrue).

daltonkell:
Following with other conventions, I'll make this case-insensitive.

mwengren:
I just emailed NDBC to make sure they were aware.   Not expecting any issue with that so case-insensitive is the way to go, thanks.

compliance_checker/ioos.py
```diff
@@ -565,8 +565,11 @@ def check_ioos_ingest(self, ds):
         m = ("To disallow harvest of this dataset to IOOS national products, "
              "global attribute \"ioos_ingest\" must be a string with value \"false\"")
         igst = getattr(ds, "ioos_ingest", None)
-        if igst is not None:
-            if igst not in ("true", "false"):
+        if igst is not None: # if not supplied, return default pass
```
benjwadams: Please combine all these nested `if`statements using `and` statements -- short circuiting ensures that anything following a false condition won't fire and cause exceptions.  Also consider expanding your variable names `igst` -> `ingest`,  `r` -> `return_status`, etc.

benjwadams: In fact you could just do `if isinstance(igst, str) and igst.lower() not in ("true", "false"):` since the not None implicitly is also done by the `isinstance` check.

daltonkell:
@benjwadams Feedback implemented.

> In fact you could just do if isinstance(igst, str) and igst.lower() not in ("true", "false"): since the not None implicitly is also done by the isinstance check.

It still must check if the value is anything *other* than None, though, because the value can *not* be supplied and still pass.

Issue #727, opened 2020-02-06
StandardNameTable enhancement


daltonkell:
The CF StandardNameTable uses list comprehensions in a variety of places, e.g.

https://github.com/ioos/compliance-checker/blob/380e16a4f84135605632d75e398564d6885d0de1/compliance_checker/cf/util.py#L276-L279

More efficient data structures, such as a `set`, could be used to improve lookup performance.

benjwadams:
Will revisit later, currently not a bottleneck on performance so closing.

PR #779, opened 2020-03-24
Adds support for forecast reference metadata in CF checks.

lukecampbell:
This commit updates cfutil's feature detection to ignore variables used
to declare forecast metadata. Specifically, variables with standard
names `forecast_period` and `forecast_reference_time` are used in
conjunction with `time` to indicate to dataset consumers the appropriate
model initialization time, forecast hour, and valid time. This is
useful for datasets that aggregate multiple forecast runs, such as FMRC.

### DIFF
--- a/compliance_checker/cfutil.py
+++ b/compliance_checker/cfutil.py
@@ -194,6 +194,9 @@ def is_geophysical(ds, variable):
     if variable in get_auxiliary_coordinate_variables(ds):
         return False
 
+    if variable in get_forecast_metadata_variables(ds):
+        return False
+
     # Is it dimensionless and unitless?
     if len(ncvar.shape) == 0 and unitless:
         return False
@@ -294,6 +297,26 @@ def get_auxiliary_coordinate_variables(ds):
         ret_val.append(aux_var)
 
     return ret_val
+
+
+def get_forecast_metadata_variables(ds):
+    '''
+    Returns a list of variables that represent forecast reference time
+    metadata.
+
+    :param netCDF4.Dataset ds: An open netCDF4 Dataset.
+    :rtype: list
+    '''
+    forecast_metadata_standard_names = [
+        'forecast_period',
+        'forecast_reference_time',
+    ]
+    forecast_metadata_variables = []
+    for varname in ds.variables:
+        standard_name = getattr(ds.variables[varname], 'standard_name', None)
+        if standard_name in forecast_metadata_standard_names:
+            forecast_metadata_variables.append(varname)
+    return forecast_metadata_variables
 
 
 def get_cell_boundary_map(ds):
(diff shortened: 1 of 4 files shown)


compliance_checker/tests/test_feature_detection.py
```diff
@@ -226,6 +226,20 @@ def test_auxiliary_coordinates(self):
             aux_coord_vards = util.get_auxiliary_coordinate_variables(nc)
             assert set(['lat', 'lon']) == set(aux_coord_vards)
 
+    def test_forecast_reference_metadata(self):
+        '''
+        Tests variables used for forecast reference metadata to ensure they are
+        not misclassified as geophysical variables.
+        '''
+        with Dataset(resources.STATIC_FILES['forecast_reference']) as nc:
+            assert util.is_geophysical(nc, 'forecast_reference_time') is False
```
benjwadams: Use `self.assertTrue`, `self.assertFalse`, and `self.assertEqual` in the below methods.  In the event the tests fail, the error messages are more useful.  `assert` can also be ignored completely in Python using the `-O` flag.  While I haven't ever seen the aforementioned flag used in practice, I still prefer using `unittest` methods for most unit test assertion.

compliance_checker/cfutil.py
```diff
@@ -296,6 +299,26 @@ def get_auxiliary_coordinate_variables(ds):
...
+    metadata.
+
+    :param netCDF4.Dataset ds: An open netCDF4 Dataset.
+    :rtype: list
+    '''
+    forecast_metadata_standard_names = [
+        'forecast_period',
+        'forecast_reference_time',
+    ]
+    forecast_metadata_variables = []
```
benjwadams: If you're only doing a simple membership check, please use a set instead of a list due to `O(1)` vs `O(n)` performance when using `in`.

compliance_checker/tests/test_feature_detection.py
lukecampbell: :heavy_check_mark:

compliance_checker/cfutil.py
lukecampbell: :heavy_check_mark:

compliance_checker/cfutil.py
```diff
@@ -307,10 +307,10 @@ def get_forecast_metadata_variables(ds):
     :param netCDF4.Dataset ds: An open netCDF4 Dataset.
     :rtype: list
     '''
-    forecast_metadata_standard_names = [
+    forecast_metadata_standard_names = {
         'forecast_period',
         'forecast_reference_time',
-    ]
+    }
     forecast_metadata_variables = []
```
benjwadams: I actually meant `forecast_metadata_variables` for the list -> set conversion, apologies if this wasn't explicit.   This likely to be fairly minor, so I'll just keep as is and merge, possibly revisiting this for #727 and other similar places in code which use lists.

lukecampbell:
Thank you

Issue #782, opened 2020-03-26
check_child_attr_data_types error from cf tests


ancillary:
We are getting the following error message from the cf:1.7 and cf:1.6 tests run on this [netcdf file](https://gyre.umeoce.maine.edu/ncei/neracoos/A01/A0142.ocean.001m.merged.nc) (and all of our other netcdf files).

Running Compliance Checker on the datasets from: ['A0101.ocean.001m.merged.nc']
WARNING: The following exceptions occured during the cf:1.7 checker (possibly indicate compliance checker issues):
cf:1.7.check_child_attr_data_types: 'bytes' object has no attribute 'dtype'
  File "c:\apps\anaconda3\envs\ioosalt\lib\site-packages\compliance_checker\cf\cf.py", line 1245, in check_child_attr_data_types
    self._parent_var_attr_type_check(att_name, var, ctx)
  File "c:\apps\anaconda3\envs\ioosalt\lib\site-packages\compliance_checker\cf\cf.py", line 442, in _parent_var_attr_type_check
    val_type = attr_val.dtype.type
    
Currently using 4.3.3rc2, but we have seen the same message with 4.3.3rc1, and 4.3.2.

Is there some issue with the dataset I am missing?  Although this example uses scale_factor = 1. and add_offest = 0. attributes, we get the same error if these attributes are omitted.

benjwadams:
Curious.  For some reason the string-like value `_FillValue` is getting side as `bytes` here.  In pretty much all string-like attributes, `str` is returned under Python 3.

In the offending variable, `missing_value` is also "0", but is a `str` type when read through netCDF4-Python:

```
ipdb> var
<class 'netCDF4._netCDF4.Variable'>
|S1 conductivity_qc_tests(time, number_tests)
    long_name: conductivity QARTOD Individual Tests
    standard_name: conductivity status_flag
    short_name: CONDQTST
    intent: data_quality
    _FillValue: b'0'
    missing_value: 0
    flag_values: 1 2 3 4 9
    flag_meanings: PASS NOT_EVALUATED SUSPECT FAIL MISSING
    comments: 14-character array with results of individual QARTOD tests.  1:Gap, 2:Syntax, 3:Location, 4:Gross Range, 5:Climatology, 6:Spike, 7:Rate of Change, 8:Flat-line, 9:Multi-variate,
 10:Attenuated Signal, 11:Neighbor, 12:Associated variable, 13:Manual, 14:Other
unlimited dimensions: time
[...]

PR #783, opened 2020-03-27
Handle both string and bytes in parent variable check

benjwadams:
Handles both string and bytes in parent variable check of attribute
types when parent variable is a string type.

Addresses #782

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -435,7 +435,7 @@ class CFBaseCheck(BaseCheck):
         """
         attr_val = var.getncattr(attr_name)
 
-        if isinstance(attr_val, str):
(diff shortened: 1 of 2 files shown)


benjwadams:
Draft PR as I'd like to make some proper tests for this prior to merging.

compliance_checker/cf/cf.py
```diff
@@ -435,7 +435,7 @@ def _parent_var_attr_type_check(self, attr_name, var, ctx):
         """
         attr_val = var.getncattr(attr_name)
 
-        if isinstance(attr_val, str):
+        if isinstance(attr_val, (str, bytes)):
```
daltonkell: I was always under the impression one would need

```
if isinstance(attr_val, str) or isinstance(attr_val, bytes):
```

this is great!

Issue #784, opened 2020-03-27
JSON/HTML output renders results with messages even if result passes (result display rendering logic discrepancies)


benjwadams:
The HTML and JSON formats will output any result objects which contain messages, even if the result itself passes.  This behavior is different than the plaintext output, which does not appear to output results when a given result passes.   Ideally the rendering of all results should use consistent rules.

Related to, and probably causing issues in #780, #757, #760.

mwengren:
Resolved by #787, closing.

Issue #780, opened 2020-03-25
Incomplete notification message for 'qartod_variables flags' test


mwengren:
Testing with dataset: http://erddap.secoora.org/erddap/tabledap/edu_usf_marine_comps_1407d550 with 4.3.3rc2, the following messages are produced for the 'qartod_variables flags' test:

Test| Message
--------|----------
qartod_variables flags | s.air_temperature_qc_agg's flag_values must be an array of values not                                                                                                        s.air_pressure_qc_agg's flag_values must be an array of values not                                                                                                        s.sea_water_electrical_conductivity_qc_agg's flag_values must be an array of values not                                                                                                        s.lwe_thickness_of_precipitation_amount_qc_agg's flag_values must be an array of values not                                                                                                        s.relative_humidity_qc_agg's flag_values must be an array of values not ...

**Issue**: need to complete the sentence in output message.

daltonkell:
When I test against the same dataset, I get

```
qartod_variables flags
* air_temperature_qc_agg's flag_values must be an array of values not <class 'str'>
* air_pressure_qc_agg's flag_values must be an array of values not <class 'str'>
* lwe_thickness_of_precipitation_amount_qc_agg's flag_values must be an array of values not <class 'str'>
* relative_humidity_qc_agg's flag_values must be an array of values not <class 'str'>
* sea_water_electrical_conductivity_qc_agg's flag_values must be an array of values not <class 'str'>
* sea_surface_height_above_sea_level_geoid_navd88_qc_agg's flag_values must be an array of values not <class 'str'>
* sea_water_temperature_qc_agg's flag_values must be an array of values not <class 'str'>
* sea_water_temperature_lower_well_qc_agg's flag_values must be an array of values not <class 'str'>
* sea_water_temperature_upper_well_qc_agg's flag_values must be an array of values not <class 'str'>
* wind_speed_of_gust_qc_agg's flag_values must be an array of values not <class 'str'>
* wind_speed_of_gust_sonic_qc_agg's flag_values must be an array of values not <class 'str'>
* wind_speed_samples_sonic_qc_agg's flag_values must be an array of values not <class 'str'>
* wind_speed_qc_agg's flag_values must be an array of values not <class 'str'>
* wind_from_direction_qc_agg's flag_values must be an array of values not <class 'str'>
* wind_speed_sonic_qc_agg's flag_values must be an array of values not <class 'str'>
[...]

Issue #757, opened 2020-03-09
Supress reporting on 'required, if applicable' attributes.  Fix gts_ingest notification messaging.  


mwengren:
Looking at some output from some example datasets we've listed in #723, I ran a test on PacIOOS' [AWS-HIMB](https://pae-paha.pacioos.hawaii.edu/erddap/info/AWS-HIMB/index.html) dataset.  

Output is available here for reference: [HTML](https://gist.github.com/mwengren/82496d8114bab0776ae34ef70510b2a1), [JSON](https://gist.github.com/mwengren/df78535eb260133df8ac6b92e1fb72d2).

Both the `gts_ingest` (variable) check and `wmo_platform_code` (global) checks should probably not produce notification messages if not present.  Reason is that not all datasets will be destined for GTS.  This is why the phrase 'required, if applicable' is used: each is not explicitly required for a dataset to comply with the Profile.  

We probably want the test instead to check for presence of a global `wmo_platform_code` attribute, and if present, confirm it meets format requirements (5 alphanumeric or 7 numeric digits).

Same for any variables with `gts_ingest` attribute (it should either be `true|false`).  

If neither is present, no notification message.

As these both are now, it results in cluttered output messages that aren't really relevant for many datasets.  

Also, second issue on the `gts_ingest` variable test: the notification message contains text from the [Requirements for GTS Ingest](https://ioos.github.io/ioos-metadata/ioos-metadata-profile-v1-2.html#requirements-for-ioos-dataset-gts-ingest) section; these requirements shouldn't be included here.  The `gts_ingest` attribute represents a subset of these larger requirements, so it doesn't make sense to report them on this specific test.

If it's possible to write a separate test to check that a dataset includes any variables that meet the GTS ingest attribution requirements, that would be cool and worth adding.  Separate issue to come.

daltonkell:
The `wmo_platform_code` piece of this issue is addressed by #766

mwengren:
Testing this dataset as of 4.3.3rc2:

http://erddap.sensors.ioos.us/erddap/tabledap/ssbn7-sun2wave-sun2w-sunset-n

Shows that the `gts_ingest` output is still appearing in results.  

For both **'required, if applicable'** global `gts_ingest` and `wmo_platform_code` attributes, as mentioned above, we should skip reporting on them if they are omitted in the dataset.  

Only if each attribute is present, should syntax checks be done and output messages generated.

daltonkell:
[...]

PR #787, opened 2020-03-30
Ioos convert to conditional results

daltonkell:
Only use messages if the check fails.


benjwadams:
Related #784, #780, #757, #760.

Issue #748, opened 2020-02-26
IOOS:1.2 `platform` check


daltonkell:
Calling the `ioos:1.2` checker on this [SECOORA dataset](http://erddap.secoora.org/erddap/info/edu_usf_marine_comps_1407d550/index.html) yields the following messages:

```
platform
* Gridded model datasets are not required to declare a platform
* The global attribute "platform" must be a single string containing no blank characters; it is None
```

This dataset is not a gridded model, and contains the attribute `platform: buoy`.

daltonkell:
__Update__

Upon further investigation, the `ncdump -h` of this dataset reveals that `platform` is actually not included in the header:

```
netcdf secoora_edu_usf_marine_comps {
dimensions:
        row = 647 ;
        Row_Type_strlen = 9 ;
        Variable_Name_strlen = 56 ;
        Attribute_Name_strlen = 28 ;
        Data_Type_strlen = 6 ;
        Value_strlen = 452 ;
variables:
        char Row_Type(row, Row_Type_strlen) ;
                Row_Type:_Encoding = "ISO-8859-1" ;
        char Variable_Name(row, Variable_Name_strlen) ;
                Variable_Name:_Encoding = "ISO-8859-1" ;
        char Attribute_Name(row, Attribute_Name_strlen) ;
                Attribute_Name:_Encoding = "ISO-8859-1" ;
        char Data_Type(row, Data_Type_strlen) ;
                Data_Type:_Encoding = "ISO-8859-1" ;
        char Value(row, Value_strlen) ;
                Value:_Encoding = "ISO-8859-1" ;

// global attributes:
                :id = "edu_usf_marine_comps_1407d550_info_1480046041" ;
}
```
The dataset was obtained via 
```
$ wget http://erddap.secoora.org/erddap/info/edu_usf_marine_comps_1407d550/index..nc -O secoora_edu_usf_marine_comps.nc
```


This may be why the checker thinks it's a gridded model.

daltonkell:
Indeed - after getting the full `.nc` file

```
[...]

PR #791, opened 2020-04-08
IOOS  1.2  platform  DSG

daltonkell:
Separate platform checks from the CF DSG checks and add corresponding unit tests, per #748.

### DIFF
--- a/compliance_checker/ioos.py
+++ b/compliance_checker/ioos.py
@@ -829,13 +829,15 @@ class IOOS1_2Check(IOOSNCCheck):
         Check that the dataset follows the restrictions for CF Discrete
         Sampling Geometries set by the IOOS Metadata Profile.
 
-        Examine the featureType of the dataset. If the featureType is
-        [profile, trajectory, timeSeriesProfile, trajectoryProfile] and cf_role
-        in [profile_id, trajectory_id], the dimensionality of the variable containing
-        cf_role must be 1. If featureType==timeSeries, a warning will be
-        triggered to inform the user that the Metadata Profile restrict features to share the
-        same lat/lon position (e.g., multiple features should be on the same
-        platform). If featureType==point, it can have whatever dimension.
+        Examine each variable with a cf_role. For each (featureType, cf_role):
+            - (timeSeries, timeseries_id)
+            - (timeSeriesProfile, timeseries_id)
+            - (trajectory, trajectory_id)
+            - (trajectoryProfile, trajectory_id)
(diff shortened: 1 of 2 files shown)


compliance_checker/ioos.py
```diff
@@ -815,62 +806,114 @@ def check_single_platform(self, ds):
...
+                    "profile restricts {feature_type} datasets to a "
+                    "single platform (ie. station/trajectory/profile) per dataset."
+                ).format(
+                    cf_role_var=var.name, cf_role=cf_role,
+                    dim=shp, feature_type=feature_type
+                )
+ 
+                _val = shp==1
+ 
+            elif feature_type=="point": # do nothing
```
benjwadams, line 900: Please leave spaces around binary operators and non-keyword assignments per suggestions in PEP8.

Bobfrat: if you're going to enforce pep8, might be worth letting travisCI catch those violations. Likewise, @daltonkell there are tools you can use that provide a pre-commit hook to check your code for pep8 compliance.

benjwadams: @Bobfrat, The thing is with CI is that a lot of the existent codebase is not necessarily PEP8 compliant already.  We have a "coding-standards" test in Travis already which at a cursory glance appears to check PEP8 compliance:  https://travis-ci.org/github/ioos/compliance-checker/jobs/672560182

It allows for failure, but to my knowledge it's running across the entire code base, so we'd have to do some real refactoring of the code base if we didn't want that test to fail and to provide valuable input.

Thus, until we push our code through something like `black` and refactor, I'd prefer to handle this on the coding side.

Bobfrat: Not sure what you had in mind for refactoring but you can get the code base a pretty good ways to pep8 compliant with autopep8. I am definitely not an autopep8 expert but the couple times I've used it I was happy with the results. 

```
find ./compliance_checker -name \*.py -exec autopep8 --recursive --aggressive --in-place {} +
```

benjwadams: @ocefpaf might know more about some of the tooling here.  Perhaps we should open a new issue to address coding standards/PEP8, etc.

ocefpaf: Do you want to use `pre-commit` hooks do to this? I can setup GH actions as a CI and those who have the pre-commit locally will get the code adjustments "for free." I'll open a PR to demonstrate that.

PR #793, opened 2020-04-09
Adds ability to fetch latest CF standard name table

benjwadams:
Adds an option to the `-d` flag to download the latest CF standard name
table without needing to explicitly specify a version.  This is expected
to be useful in cases with non-interactive runs of the compliance
checker (e.g. scheduled checker runs) where an up to date standard name
table is desired.

### DIFF
--- a/compliance_checker/cf/util.py
+++ b/compliance_checker/cf/util.py
@@ -264,7 +264,8 @@ class StandardNameTable(object):
             with io.open(cached_location, 'r', encoding='utf-8') as fp:
                 resource_text = fp.read()
         elif os.environ.get('CF_STANDARD_NAME_TABLE') and os.path.exists(os.environ['CF_STANDARD_NAME_TABLE']):
-            with io.open(os.environ['CF_STANDARD_NAME_TABLE'], 'r', encoding='utf-8') as fp:
+            with io.open(os.environ['CF_STANDARD_NAME_TABLE'], 'r',
+                         encoding='utf-8') as fp:
                 resource_text = fp.read()
         else:
             resource_text = get_data("compliance_checker", "data/cf-standard-name-table.xml")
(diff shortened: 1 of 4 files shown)


compliance_checker/cf/util.py
```diff
@@ -327,15 +328,28 @@ def download_cf_standard_name_table(version, location=None):
...
         location = resource_filename('compliance_checker', 'data/cf-standard-name-table.xml')
 
-    url = "http://cfconventions.org/Data/cf-standard-names/{0}/src/cf-standard-name-table.xml".format(version)
-    r = requests.get(url, allow_redirects=True)
-    if r.status_code == 200:
-        print("Downloading cf-standard-names table version {0} from: {1}".format(version, url), file=sys.stderr)
-        with open(location, 'wb') as f:
-            f.write(r.content)
+    if version == 'latest':
+        from bs4 import BeautifulSoup
```
daltonkell, line 332: I've always been confused by in-line imports like this. Not saying this is bad, but I've seen sources in the past say that inline imports are organizationally bad. It seems that it actually makes more sense to have them like this if the option is rarely used.

compliance_checker/cf/util.py
```diff
@@ -327,15 +328,28 @@ def download_cf_standard_name_table(version, location=None):
...
+        tables_req = requests.get("http://cfconventions.org/documents.html")
+        tables_req.raise_for_status()
+        # use bs4 due to HTML parse errors under lxml
+        html_content = BeautifulSoup(tables_req.text, features='lxml')
+        # the first a element found with href ending in
+        # "cf-standard-name-table.xml should be the latest CF table
+        html_elem = html_content.find('a',
+                     {'href': lambda s: s.endswith("cf-standard-name-table.xml")})
+        url = urljoin("http://cfconventions.org", html_elem.attrs['href'])
     else:
```
daltonkell, line 343: What is the behavior when users request `... -d banana-town`? Does this have an ugly error on invalid table versions?

benjwadams: Requests would try to fetch the version from the CF Conventions webpage and it would 404.  `raise_for_status()` then raises an exception.

compliance_checker/cf/util.py
benjwadams: The consensus from the Python community seems to be that it's OK to use it's for a feature that's not used all the time.  For example, Pandas has certain code paths for Excel output of a dataframe which appear to import the libraries when the exporting function is called.  

In this particular case, since people won't necessarily always be updating their standard name tables every time, it's not necessary to import the library for BeautifulSoup every time.

daltonkell:
Good to go from my perspective 👍

compliance_checker/cf/util.py
Bobfrat: What's pep8 have to say about it?

benjwadams: Hmm, my linter didn't consider that an error.  I'm going to rewrite this code to use plain lxml without the BeautifulSoup extension.

benjwadams:
Closing to implement in lxml instead and avoid introducing a new library dependency.

PR #756, opened 2020-03-09
remove versioneer and add a pyproject.toml

ocefpaf:
Sadly the `versioneer` project is dead and modern python packaging standards won't work with it. This PR starts a slow transition towards modern python packaging techniques following PEPs 517 and 518.

In this PR we:

- switch from `versioneer` to `setuptools_scm` as the version number manager;
- added a "bare bones" `pyproject.toml`.

Just by adding an empty `pyproject.toml` we already adhere to PEP 517 and make installing compliance-checker easier with latest `pip`. I also added a minimum build system requirements expected to build compliance-checker.

---
Note that we will get some errors locally due to the lack of `cython` but that is a bug in cf-units and should be fixed in https://github.com/SciTools/cf-units/pull/156.

---
### with pyproject.toml
```
> pip install -e . --no-deps --force-reinstall
Obtaining file:///home/filipe/Dropbox/pymodules/01-forks/compliance-checker
  Installing build dependencies ... done
  Getting requirements to build wheel ... done
    Preparing wheel metadata ... done
Installing collected packages: compliance-checker
  Attempting uninstall: compliance-checker
    Found existing installation: compliance-checker 4.3.3rc2.dev11+g3a7c3cd.d20200309
    Uninstalling compliance-checker-4.3.3rc2.dev11+g3a7c3cd.d20200309:
      Successfully uninstalled compliance-checker-4.3.3rc2.dev11+g3a7c3cd.d20200309
  Running setup.py develop for compliance-checker
Successfully installed compliance-checker
```

### without pyproject.toml
```
> pip install -e . --no-deps --force-reinstall
Obtaining file:///home/filipe/Dropbox/pymodules/01-forks/compliance-checker
Installing collected packages: compliance-checker
  Attempting uninstall: compliance-checker
    Found existing installation: compliance-checker 0.0.0
    Uninstalling compliance-checker-0.0.0:
      Successfully uninstalled compliance-checker-0.0.0
  Running setup.py develop for compliance-checker
Successfully installed compliance-checker
```

Notice that the main difference is that, with the `pyproject.toml`, pip will:

- Install build dependencies separately from the actual package dependencies
- Get requirements to build wheel also unrelated to the package dependencies
- Prepare a proper wheel metadata

After this is merged I'll start working on a PR that moves the metadata from the `setup.py` to the `pyproject.toml`. Just don't want to do it all at once to avoid breakages.

PR #794, opened 2020-04-10
add pre-commit hooks

ocefpaf:
Once merge this will add a GitHub action that runs the pre-commit hooks configuration.

The checks are:

https://github.com/pre-commit/pre-commit-hooks
- trailing space
- check-ast
- debug-statements

https://github.com/pre-commit/mirrors-isort
https://github.com/asottile/seed-isort-config
- isort with third party from seed-isort

https://github.com/psf/black
- black

https://github.com/pre-commit/mirrors-mypy
- mypy

I use all of those in my projects but we may not need of them at once here.

Fixing all code style checks cab be done incrementally or in a single massive commit. While the latter is harder I'd rather do it once and then we just need to keep the code tidy. Also, a massive commit makes git-blame easier thanks to git >=2.23 `--ignore-revs`', see https://git-scm.com/docs/git-blame#Documentation/git-blame.txt---ignore-revs-fileltfilegt.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
   hooks:
   - id: isort
     additional_dependencies: [toml]
-    args: [--project=erddapy, --multi-line=3, --lines-after-imports=2, --lines-between-types=1, --trailing-comma, --force-grid-wrap=0, --use-parentheses, --line-width=88]
+    args: [--project=compliance_checker, --multi-line=3, --lines-after-imports=2, --lines-between-types=1, --trailing-comma, --force-grid-wrap=0, --use-parentheses, --line-width=88]
 
 - repo: https://github.com/asottile/seed-isort-config
   rev: v1.9.4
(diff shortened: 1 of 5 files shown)


ocefpaf:
If you have `pre-commit` locally you can test the changes that will be needed by running `pre-commit run --all-files`. Also, this will be a bit easier if we merge https://github.com/ioos/compliance-checker/pull/756.

We should also recommend devs to install `pre-commit` with `pre-commit install` so the checks are always run with every commit command.

.pre-commit-config.yaml
```diff
@@ -0,0 +1,39 @@
...
+    - id: flake8
+      exclude: docs/source/conf.py
+      args: [--max-line-length=105]
+
+- repo: https://github.com/pre-commit/mirrors-isort
+  rev: v4.3.21
+  hooks:
+  - id: isort
+    additional_dependencies: [toml]
+    args: [--project=erddapy, --multi-line=3, --lines-after-imports=2, --lines-between-types=1, --trailing-comma, --force-grid-wrap=0, --use-parentheses, --line-width=88]
```
daltonkell, line 21: What is the `--project=erddapy` flag used for?

ocefpaf: For copy-n-pasta corrections ;-p

PR #795, opened 2020-04-10
Adds ability to fetch latest CF standard name table

benjwadams:
Adds an option to the `-d` flag to download the latest CF standard name
table without needing to explicitly specify a version.  This is expected
to be useful in cases with non-interactive runs () of the compliance
checker (e.g. scheduled checker runs) where an up to date standard name
table is desired.

Supersedes #793 and eliminates the need for BeautifulSoup4 library.

### DIFF
--- a/compliance_checker/cf/util.py
+++ b/compliance_checker/cf/util.py
@@ -6,7 +6,9 @@ import sys
 from copy import deepcopy
 from collections import defaultdict
 from lxml import etree
+import lxml.html
 from cf_units import Unit
+from urllib.parse import urljoin
 from netCDF4 import Dimension, Variable
 from pkgutil import get_data
 from pkg_resources import resource_filename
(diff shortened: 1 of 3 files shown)


compliance_checker/cf/util.py
```diff
@@ -327,15 +330,25 @@ def download_cf_standard_name_table(version, location=None):
...
-    url = "http://cfconventions.org/Data/cf-standard-names/{0}/src/cf-standard-name-table.xml".format(version)
-    r = requests.get(url, allow_redirects=True)
-    if r.status_code == 200:
-        print("Downloading cf-standard-names table version {0} from: {1}".format(version, url), file=sys.stderr)
-        with open(location, 'wb') as f:
-            f.write(r.content)
+    if version == 'latest':
+        tables_tree = lxml.html.parse("http://cfconventions.org/documents.html")
+        end_str = "cf-standard-name-table.xml"
+        xpath_expr = ("//a[substring(@href, string-length(@href) - "
```
benjwadams, line 336: NB: However verbose this is, this really is how you do something equivalent to something like Python's `str.endswith` when using XPath 1.0.  XPath 2.0 has an `ends-with` function, which is considerably more concise, but LXML only uses XPath 1.0.

PR #796, opened 2020-04-13
Fix coding standards

ocefpaf:
Follow up on #794 

This "fixes" all the coding standards that the pre-commit would catch. I left the `mypy` checks for later for simplicity.

While this PR is massive the changes are almost all automated.


ocefpaf:
I recommend to merge any pending PR before this one to avoid conflicts. It is easier to fix this one than the others.

ocefpaf:
All the mypy failure above https://github.com/ioos/compliance-checker/pull/796/checks?check_run_id=583115305#step:6:89 will go away with https://github.com/ioos/compliance-checker/pull/756

ocefpaf:
@benjwadams this caused conflicts with the opened PRs. In my experience the easiest way to fix that is to keep the PR's changes and let pre-commit "fix the style" again.

PR #797, opened 2020-04-15
Add vertical coordinate variable checks for IOOS 1.2

benjwadams:
Checks vertical variables for IOOS 1.2 profile.  Checks any variable
with a "Z" axis that the following criteria are met:

- positive attribute is either "up" or "down"

- units attribute is one of "meter", "inch", "foot", "yard",
  "US_survey_foot", or "fathom", or equivalent abbreviations or plural
  forms.  Note that unlike CF checks, other length units recognized by
  UDUNITS are not accepted.


benjwadams:
@mwengren, I based this check off the correspondence we had via email.  As the requested implementation hasn't made it into the main IOOS 1.2 spec yet, I have marked this PR as a draft for the time being.

mwengren:
@benjwadams Looks reasonable.  Good call on draft status, we can leave as is until the guidance is posted in the IOOS Profile.  It's possible we may need to add additional accepted units, but that looks fairly easy to do in your code.  I'm hoping to get some feedback on this from NDBC in the near future.

compliance_checker/ioos.py
```diff
@@ -950,6 +953,56 @@ def _check_gts_ingest_val(self, val):
...
+
+        :param netCDF4.Dataset ds: An open netCDF dataset
+        :rtype: list
+        :return: List of results
+        '''
+        ret_val = []
+        z_variables = cfutil.get_z_variables(ds)
+        for name in z_variables:
+            variable = ds.variables[name]
+            standard_name = getattr(variable, 'standard_name', None)
```
mwengren, line 974: Maybe can skip this as it's not used anywhere.

compliance_checker/ioos.py
```diff
@@ -6,11 +6,14 @@
                                      attr_check)
 from owslib.namespaces import Namespaces
 from lxml.etree import XPath
+from cf_units import Unit
 from compliance_checker.acdd import ACDD1_3Check
+from compliance_checker import cfutil
```
daltonkell, line 11: Why not just import `get_z_variables` in the line below with the other methods?

compliance_checker/ioos.py
daltonkell: Agreed

benjwadams: Oops, meant to do that.

compliance_checker/ioos.py
benjwadams: This isn't an exceptional use case of relative vs. qualified imports.  I'm probably going to keep it the same since it largely comes down to semantics.

benjwadams:
I'm curious why CI does not appear to be testing this branch properly -- is it because it's a draft PR?

compliance_checker/ioos.py
benjwadams: Oh, I see.  There's already some relative imports from cfutil.  I'll change it over.

mwengren:
@benjwadams 

I'm going to update the Metadata Profile this week, hopefully, but I think we should be a little more flexible in the restriction of distance units allowed.  Can you update this PR accordingly?

Going from: https://github.com/nco/nco/blob/master/data/udunits.dat, let's include:

meter
metre
inch
foot
yard
mile
US_survey_foot
US_survey_feet
US_survey_yard
US_survey_mile
US_statute_mile
fathom 
international_inch
international_inches
international_foot
international_feet
international_yard
international_mile
inches
in
feet
ft
yd
mi

I'll also post again here when I add these guidelines in the profile.

mwengren:
@benjwadams Guidance for vertical coordinate variable added to the IOOS profile:

https://ioos.github.io/ioos-metadata/ioos-metadata-profile-v1-2.html#requirements-for-vertical-coordinate-variable

Please update and merge PR accordingly.

benjwadams:
@mwengren, will do.

benjwadams:
Accepted units have been updated.  Newest commit just adds mile, since the other units have aliases, plural forms, or abbreviations already handled by UDUnits.

TravisCI failure has to do with downloading of CF standard name tables -- it doesn't appear that this test introduced any code that would cause a test failure.

Issue #798, opened 2020-04-16
Requests to ERDDAP TableDAP and Desired Response Format


daltonkell:
Per the discussion in the ERDDAP Google Group,

https://groups.google.com/d/msg/erddap/kyYF8wcNeME/p5TSngqPAwAJ

The default `ERDDAP` response for a format-agnostic TableDAP request is essentially a raw OPeNDAP sequence. The resulting dataset is one with `s.` pre-pended to all the variable names:

```
<xarray.Dataset>
Dimensions:                                                  (s: 27617)
Dimensions without coordinates: s
Data variables:
    s.time                                                   (s) datetime64[ns] ...
    s.latitude                                               (s) float64 ...
    s.longitude                                              (s) float64 ...
    s.z                                                      (s) float64 ...
    s.backscatter_intensity_2651_a                           (s) float64 ...
    s.sea_water_velocity_to_direction_2651ds_a               (s) float64 ...
    s.sea_water_speed_2651ds_a                               (s) float64 ...
    s.eastward_sea_water_velocity_2651ds_a                   (s) float64 ...
    s.northward_sea_water_velocity_2651ds_a                  (s) float64 ...
    s.sea_water_pressure_cm_time__standard_deviation_2651_a  (s) float64 ...
    s.sea_water_pressure_2651_a                              (s) float64 ...
    s.sea_water_temperature_2651_a                           (s) float64 ...
```

(dataset used in Google group example, courtesy R. Signell)

This can cause several checks to fail for at least two reasons:

1. If a variable `s.a` has an attribute which refers to another variable by name, `b`, and must use `b` in a check, `b` no longer exists but instead is replaced by `s.b`

2. The dimensions of the OPeNDAP sequence are almost guaranteed to not be CF DSG-compliant

This raises the all-important question: _What format should we request when supplying an ERDDAP URL to the Compliance Checker?_

ERDDAP can provide the same dataset in 44 different formats. Typically, when running the checker against a dataset, the `.ncCF` (NetCDF Contiguous Ragged Array) data format or `.ncCFMA` (NetCDF CF Discrete Sampling Geometries File) format work best, since the binary data is encoded in such a way that the features have a single _instance dimension_. Other formats, such as the `.nc` (plain, table-like NetCDF-3 format) align the variables using different dimensions (this does _not_ alter the geospatial location of the data in any way) which may cause the dataset to fall out of compliance.

[...]

PR #799, opened 2020-04-20
Enable ncCF format requests to TableDAP

daltonkell:
__ERDDAP TableDAP Requests__

When making a request to `TableDAP`, ensure the `.ncCF` (Contiguous ragged array) format is returned. In order for the data to be returned, the full URl must be generated listing all the variables as the `TableDAP` query. The binary data is fetched from the server, and then instantiated as a `netCDF4.Dataset` representation, allowing proper checking.

Users are only required to supply the base URL to the dataset, e.g.

```
$ compliance-checker -t ioos "http://data.glos.us/erddap/tabledap/glerlwe2"
```

This commit addresses #798.

It should be noted that this approach takes a significant amount of time to fetch the data, far longer than downloading the file and running the checker on a local file.

Unit tests are forthcoming after deliberating on this approach.

### DIFF
--- a/compliance_checker/protocols/erddap.py
+++ b/compliance_checker/protocols/erddap.py
@@ -1,0 +1,42 @@
+import io
+import urllib.request
+import compliance_checker.protocols.opendap as opendap
+
+def is_tabledap(url):
+    """
+    Identify a dataset as an ERDDAP TableDAP dataset.
+
+    Parameters
+    ----------
+    url (str) : URL to dataset
+
+    Returns
+    -------
+    bool
+    """
+
+    if "tabledap" in url:
+        return True
+    return False
+
+def get_tabledap_bytes(url, ftype):
+    """
+    ERDDAP TableDAP returns an OPeNDAP "sequence" response by default
+    when no file extensions are provided. If a user wishes to get a dataset
+    from an ERDDAP TableDAP URL, append the desired file extension and return
+    a byte buffer object containing the data.
+
+    Parameters
+    ----------
+    url (str)   : URL to TableDAP dataset
+    ftype (str) : file format extension
+
+    Return
+    ------
+    io.BytesIO buffer object
+    """
+
+    vstr = opendap.create_DAP_variable_str(url) # variable str from DDS
+    _url = f'{".".join([url, ftype])}?{vstr}'
+    with urllib.request.urlopen(_url) as resp:
+        return io.BytesIO(resp.read())
--- a/compliance_checker/protocols/opendap.py
+++ b/compliance_checker/protocols/opendap.py
@@ -4,8 +4,45 @@ compliance_checker/protocols/opendap.py
 
 Functions to assist in determining if the URL is an OPeNDAP endpoint
 '''
+import io
 import requests
+import urllib.parse
+import urllib.request
 
+def create_DAP_variable_str(url):
+    """
+    Create a URL-encoded string of variables for a given DAP dataset.
+    Works on OPeNDAP datasets.
+
+    Parameters
+    ----------
+    url (str): endpoint to *DAP dataset
+
+    Returns
+    -------
+    str
+    """
+
+    # get dds
+    with urllib.request.urlopen(f"{url}.dds") as resp:
+        strb = io.StringIO(resp.read().decode())
+
+    strb.seek(8) # remove "Dataset "
+    x = strb.read()
+    strb.close()
+
+    # remove beginning and ending braces, split on newlines
+    lst = list(filter(lambda x: "{" not in x and "}" not in x, x.split("\n")))
+
+    # remove all the extra space used in the DDS string
+    lst = list(filter(None, map(lambda x: x.strip(" "), lst)))
+
+    # now need to split from type, grab only the variable and remove ;
+    lst = list(map(lambda x: x.split(" ")[-1].strip(";"), lst))
+
+    # encode as proper URL characters
+    varstr = urllib.parse.quote(",".join(lst))
+    return varstr
 
 def is_opendap(url):
     '''
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -15,7 +15,7 @@ from compliance_checker.base import fix_return_value, Result, GenericFile
 from compliance_checker.cf.cf import CFBaseCheck
 from owslib.sos import SensorObservationService
 from owslib.swe.sensor.sml import SensorML
-from compliance_checker.protocols import opendap, netcdf, cdl
+from compliance_checker.protocols import opendap, netcdf, cdl, erddap
 from compliance_checker.base import BaseCheck
 from compliance_checker import MemoizedDataset
 from collections import defaultdict
@@ -723,7 +723,13 @@ class CheckSuite(object):
         :param str ds_str: URL to the remote resource
         '''
 
-        if opendap.is_opendap(ds_str):
+        if erddap.is_tabledap(ds_str):
+            return Dataset(
+                ds_str,
+                mode="r",
+                memory=erddap.get_tabledap_bytes(ds_str, "ncCF").getbuffer()
+            )
+        elif opendap.is_opendap(ds_str):
             return Dataset(ds_str)
         else:
             # Check if the HTTP response is XML, if it is, it's likely SOS so


compliance_checker/protocols/erddap.py
```diff
@@ -0,0 +1,42 @@
...
+    ----------
+    url (str)   : URL to TableDAP dataset
+    ftype (str) : file format extension
+
+    Return
+    ------
+    io.BytesIO buffer object
+    """
+
+    vstr = opendap.create_DAP_variable_str(url) # variable str from DDS
```
ocefpaf, line 39: Minor comment: you called `vstr` `varstr` in the `create_DAP_variable_str` function, it could helps others reading your code to use the same name here. Also, your variable names are too short :smile:

compliance_checker/protocols/opendap.py
```diff
@@ -4,8 +4,45 @@
...
+
+    # remove all the extra space used in the DDS string
+    lst = list(filter(None, map(lambda x: x.strip(" "), lst)))
+
+    # now need to split from type, grab only the variable and remove ;
+    lst = list(map(lambda x: x.split(" ")[-1].strip(";"), lst))
+
+    # encode as proper URL characters
+    varstr = urllib.parse.quote(",".join(lst))
+    return varstr
```
ocefpaf, line 45: I wonder if there is an ERDDAP trick to request the JSON with out any data, they this could be simply a request to get `columnNames`.

compliance_checker/suite.py
```diff
@@ -723,7 +723,13 @@ def load_remote_dataset(self, ds_str):
         :param str ds_str: URL to the remote resource
         '''
 
-        if opendap.is_opendap(ds_str):
+        if erddap.is_tabledap(ds_str):
+            return Dataset(
```
ocefpaf, line 727: If this is slower than downloading the file and running the checks locally as you say, could something like a temporary file be an alternative faster solution (I did not run any tests)?

Downloading the file locally and removing it when done could be something like:

```python
@contextmanager
def tempnc(data: BinaryIO) -> Generator[str, None, None]:
    from tempfile import NamedTemporaryFile

    tmp = None
    try:
        tmp = NamedTemporaryFile(suffix=".nc", prefix="compliance-checker_")
        tmp.write(data)
        tmp.flush()
        yield tmp.name
    finally:
        if tmp is not None:
            tmp.close()
```

compliance_checker/protocols/erddap.py
daltonkell: Guilty as charged... I might be getting influenced by too many scientists 😆

daltonkell:
@ocefpaf The JSON and tempfile comments are very intriguing, I'll try to write some stuff up for that to see if it's feasible. Thanks for the feedback!

compliance_checker/protocols/erddap.py
ocefpaf: Wait! I'm a scientists!! Am I doing my job wrong ;-p

Bobfrat: you're cut from a different cloth felipe!

daltonkell:
### Update

Before I went and tested the `tempfile` route, I thought to test against a few more remote datasets. It's possible that the length of time is a direct result of where the data is coming from. For instance,

```
$ time python cchecker.py -t ioos "http://testing.erddap.axds.co/erddap/tabledap/sun2wave_timeseries"

...

real    4m52.888s
user    0m1.538s
sys     0m2.044s
```

That's reasonable. 

This PacIOOS dataset seems to take longer though:

```
$ time python cchecker.py -t ioos "https://pae-paha.pacioos.hawaii.edu/erddap/tabledap/WQB-04"

 ...

real    0m23.115s
user    0m1.810s
sys     0m4.256s
```

In fact, in earlier tries this morning, this dataset took over _4 minutes_ to load up, obviously an unacceptable timeframe. 23 seconds is certainly not ideal, but...

I have a hunch this is a mixture of IO and computation. When ERDDAP receives a request for a particular format, it must generate that format at request-time. If that data were not cached, ERDDAP must then open all the necessary files and then generate the requested format, _then_ send the bytes over -- much like `TDS`.

My first try at a SECOORA dataset:

```
$ time python cchecker.py -t ioos http://erddap.secoora.org/erddap/tabledap/edu_usf_marine_comps_1407d550

...

real    2m15.368s
user    0m1.204s
sys     0m6.769s
```

compliance_checker/suite.py
mwengren: Do we need some way to limit the amount of data requested?  The goal of the check is really to check the dataset structure, not so much the contents of the whole dataset.  I'm not really sure how this part of CC works, but is there a row limit or something that can be included in the ERDDAP request, if we're not doing that already?

ocefpaf: There are a few tests that requires the data/coordinates but a `.cdl` only test (metadata only) for online uses is something I've been wanting for a long time. However, do we have an ERDDAP response that returns only the metadata?

mwengren: That's true, need to check the data for some tests.  Not sure about this one though, mostly attribute-only, with some dimension checks like this one.  Possibly forgetting some things tho.  

There's the `.das` response, for example: http://erddap.sensors.ioos.us/erddap/tabledap/ssbn7-sun2wave-sun2w-sunset-n.das.  

This is what Bob pointed me to as most useful for checking dataset metadata.

daltonkell: With the OPeNDAP protocol, dimension information is typically be found in the `DDS`, at the `.dds` format, and would look something like this:

```
...
    Float64 mask_v[eta_v = 290][xi_v = 332];
    Float64 mask_psi[eta_psi = 290][xi_psi = 331];
    Float32 zeta[time = 762][eta_rho = 291][xi_rho = 332];
    Float32 u[time = 762][s_rho = 20][eta_u = 291][xi_u = 331];
    Float32 v[time = 762][s_rho = 20][eta_v = 290][xi_v = 332];
...
```

However, a request to an ERDDAP source doesn't bring back the same structure:

```
dalton@ubuntu:~$ curl -L "http://erddap.sensors.ioos.us/erddap/tabledap/ssbn7-sun2wave-sun2w-sunset-n.dds"                                               
Dataset {
  Sequence {
    Float64 time;
    Float64 latitude;
    Float64 longitude;
    Float64 z;
    Float64 sea_water_velocity_to_direction;
    Int32 sea_water_velocity_to_direction_qc_agg;
    String sea_water_velocity_to_direction_qc_tests;
    Float64 sea_water_speed;
    Int32 sea_water_speed_qc_agg;
    String sea_water_speed_qc_tests;
    Float64 sea_water_temperature;
    Int32 sea_water_temperature_qc_agg;
    String sea_water_temperature_qc_tests;
    Float64 peak_wave_period;
    Int32 peak_wave_period_qc_agg;
    String peak_wave_period_qc_tests;
    Float64 sea_surface_wave_significant_height;
    Int32 sea_surface_wave_significant_height_qc_agg;
    String sea_surface_wave_significant_height_qc_tests;
    Float64 sea_surface_wave_from_direction;
    Int32 sea_surface_wave_from_direction_qc_agg;
    String sea_surface_wave_from_direction_qc_tests;
    String station;
  } s;
} s;
```

I'll dig around in the documentation to see if that's a possibility.

For the time being, the logic in this commit produces the desired results of requesting the `.ncCF` format.

daltonkell: Ah, the `.ncCFHeader` might yield what we're looking for:

```
$ curl -L "http://erddap.sensors.ioos.us/erddap/tabledap/ssbn7-sun2wave-sun2w-sunset-n.ncCFHeader"
netcdf ssbn7-sun2wave-sun2w-sunset-n_e698_69e3_e309.nc {
  dimensions:
    timeseries = 1;
    profile = 127371;
    obs = 254742;
    station_strlen = 1;
    sea_water_velocity_to_direction_qc_tests_strlen = 1;
    sea_water_speed_qc_tests_strlen = 1;
    sea_water_temperature_qc_tests_strlen = 1;
    peak_wave_period_qc_tests_strlen = 1;
    sea_surface_wave_significant_height_qc_tests_strlen = 1;
    sea_surface_wave_from_direction_qc_tests_strlen = 1;
  variables:
    double latitude(timeseries=1);
      :_CoordinateAxisType = "Lat";
      :axis = "Y";
      :ioos_category = "Location";
      :long_name = "Latitude";
      :standard_name = "latitude";
      :units = "degrees_north";

    double longitude(timeseries=1);
      :_CoordinateAxisType = "Lon";
      :axis = "X";
      :ioos_category = "Location";
      :long_name = "Longitude";
      :standard_name = "longitude";
      :units = "degrees_east";
...
```

mwengren: Oh, yeah.  Duh, I've been staring at those forever, should have thought of that.  

For the `.das` response, the reason that may not show dimensions like other DAP servers is that the different ERDDAP output formats have different dimensionality, so it can't account for each of them.

compliance_checker/protocols/erddap.py
```diff
@@ -0,0 +1,42 @@
...
+    Parameters
+    ----------
+    url (str) : URL to dataset
+
+    Returns
+    -------
+    bool
+    """
+
+    if "tabledap" in url:
```
Bobfrat, line 18: I'd suggest using url.lower() here just to be safe.

ocefpaf:
> Before I went and tested the tempfile route, I thought to test against a few more remote datasets. It's possible that the length of time is a direct result of where the data is coming from.

I'd there is no speed gain I prefer the approach you have here. Dealing with tempfiles is always troublesome, specially on Windows.

compliance_checker/protocols/erddap.py
benjwadams: Just use `return "tabledap" in url.lower()` to simplify the return since it's returning boolean type.

benjwadams: You also can't necessarily guarantee "tabledap" will be in the URL, e.g. Apache/Nginx setups.  Usually it will though.

compliance_checker/protocols/opendap.py
```diff
@@ -4,8 +4,45 @@
...
+    Returns
+    -------
+    str
+    """
+
+    # get dds
+    with urllib.request.urlopen(f"{url}.dds") as resp:
+        strb = io.StringIO(resp.read().decode())
+
+    strb.seek(8) # remove "Dataset "
```
benjwadams, line 30: Consider using a context manager/`with` here.

compliance_checker/suite.py
```diff
@@ -723,7 +723,13 @@ def load_remote_dataset(self, ds_str):
...
+                ds_str,
+                mode="r",
+                memory=erddap.get_tabledap_bytes(ds_str, "ncCF").getbuffer()
+            )
+        elif opendap.is_opendap(ds_str):
```
benjwadams, line 732: Consider adding a case for files that have a mimetype of `application/x-netcdf` so that you can hit a netCDF file directly.  Below is an example that only grabs part of the data.

E.g.:
```
curl -I 'https://gliders.ioos.us/erddap/tabledap/amelia-20180501T0000.ncCF?&time%3E=max(time)-1%20day'
HTTP/1.1 200 
Server: nginx/1.16.1
Date: Fri, 24 Apr 2020 19:36:26 GMT
Content-Type: application/x-netcdf
Connection: keep-alive
Last-Modified: Fri, 24 Apr 2020 19:36:26 GMT
xdods-server: dods/3.7
erddap-server: 2.02
Content-Disposition: attachment;filename=amelia-20180501T0000_dec7_6846_3101.nc
Content-Encoding: identity
Access-Control-Allow-Origin: *
```

compliance_checker/protocols/erddap.py
```diff
@@ -0,0 +1,42 @@
...
+    url (str)   : URL to TableDAP dataset
+    ftype (str) : file format extension
+
+    Return
+    ------
+    io.BytesIO buffer object
+    """
+
+    vstr = opendap.create_DAP_variable_str(url) # variable str from DDS
+    _url = f'{".".join([url, ftype])}?{vstr}'
```
benjwadams, line 40: We want to maintain backwards compatibility with Python 3 versions less than 3.6.  For this reason, `str.format()` should be used instead of f-strings.

compliance_checker/protocols/opendap.py
```diff
@@ -4,8 +4,45 @@
...
+    x = strb.read()
+    strb.close()
+
+    # remove beginning and ending braces, split on newlines
+    lst = list(filter(lambda x: "{" not in x and "}" not in x, x.split("\n")))
```
benjwadams, line 35: Try to express this as a nested loop rather than reassigning to the same variable.  If there are issues with the parsing, it makes it much easier to track down the issues because the variable reference isn't changed.

compliance_checker/suite.py
benjwadams: @ocefpaf: there's also the option to read a netCDF file into memory.  It should probably be preferred over writing to a tempfile when possible and when the netCDF-C library has been compiled with in-memory support.

ocefpaf: Indeed. But if we are aiming for a broader audience we should not rely on an option that may not be available, right?

PS: I believe that the conda-forge pkg is compiled with in-memory support but I'm not 100% sure. I'll check and fix it if not.

mwengren: @daltonkell where do we stand with merging this PR and resolving this dimension check issue?    Interested to include this in the next RC, if possible.   

Can we implement the dimension check, and maybe most or all of the IOOS checks, by requesting the `.ncCFHeader` response instead of the full `.ncCF` output?  For some ERDDAP datasets, if there aren't limits placed on the request, any of the `.nc`, `.ncCF`, `.ncCFMA`, or even `.csv` output types could end up requesting a lot of data that could contribute to poor performance.

That may be too big a change though.  If so, let's go with `.ncCF` and see what results are for next RC.

daltonkell: I think including this in the next RC is appropriate, but we'll have to leave the `.ncCFHeader` request out for another edition. It's a great idea and perhaps we can get some good contributions for it, but it requires finding a way to "switch off" the checks which examine data (which would obviously fail).

I'm currently working on merging concepts from this PR and Ben's latest, #800, because his implements a useful abstraction for handling any remote netCDF resource.

compliance_checker/suite.py
benjwadams: Resolved by #801

PR #800, opened 2020-04-27
Process Content-Type header values of application/x-netcdf

benjwadams:
Issues an HTTP HEAD request and handles "Content-Type: application/x-netcdf"
headers by attempting to open a netCDF file in memory.  Opening in
memory will require the netCDF-C library to be compiled with that
feature.

### DIFF
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -765,20 +765,43 @@ class CheckSuite(object):
         :param str ds_str: URL to the remote resource
         """
 
-        if opendap.is_opendap(ds_str):
-            return Dataset(ds_str)
+
+        # Some datasets do not support HEAD requests!  The vast majority will,
+        # however, support GET requests
+        try:
+            head_req = requests.head(ds_str, allow_redirects=True, timeout=10)
+            head_req.raise_for_status()
+        except:
+            warnings.warn("Could not complete HEAD request for resource "
+                          "{}.  Skipping content-type headers".format(ds_str))
+            content_type = None
         else:
+            content_type = head_req.headers.get("content-type")
+
+        # if the Content-Type header returned was "application/x-netcdf",
+        # or a netCDF file (not OPeNDAP) we can open this into an
+        if content_type == "application/x-netcdf":
+            response = requests.get(ds_str, allow_redirects=True,
+                                    timeout=60)
+            # TODO: handle case when netCDF C libs weren't compiled with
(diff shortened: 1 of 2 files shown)


benjwadams:
Meant to mark this as draft.

@ocefpaf, I recently saw your snippet for writing to tempfiles on #799.  I will probably integrate it or something similar in case a given user's netCDF-C library wasn't compiled with in-memory support.

compliance_checker/suite.py
```diff
@@ -765,20 +765,43 @@ def load_remote_dataset(self, ds_str):
...
-                return self.process_doc(response.content)
-
-            raise ValueError(
-                "Unknown service with content-type: {}".format(
-                    response.headers["content-type"]
-                )
-            )
+         # some SOS servers don't seem to support HEAD requests.
+         # Issue GET instead if we reach here and can't get the response
+        response = requests.get(ds_str, allow_redirects=True,
```
benjwadams, line 796: This is ugly, but something like this is necessary if HEAD requests aren't being accepted, which was the case for a few of the SOS servers referenced in the integration tests.

compliance_checker/suite.py
```diff
@@ -765,20 +765,43 @@ def load_remote_dataset(self, ds_str):
...
+        # or a netCDF file (not OPeNDAP) we can open this into an
+        if content_type == "application/x-netcdf":
+            response = requests.get(ds_str, allow_redirects=True,
+                                    timeout=60)
+            # TODO: handle case when netCDF C libs weren't compiled with
+            # in-memory support by using tempfiles or other means
+            return MemoizedDataset(response.content, memory=response.content)
+
+        elif opendap.is_opendap(ds_str):
+            return Dataset(ds_str)
```
benjwadams, line 791: Should we use the MemoizedDataset form here?  It's potentially faster if we know the data isn't changing underneath us.   I don't fully understand the OPeNDAP protocol though, so I'm leaving this as is for now.

daltonkell:
Before I review @benjwadams, can you explain how this ties back to #799? Is this to be used in conjunction with the methods (which will be improved) that I began writing to fetch a `.ncCF` dataset?

benjwadams:
This is mostly to be able to point at a non-OPeNDAP netCDF resource and be able to open it and read it.   This is useful for ERDDAP TableDAP datasets.

benjwadams:
Superseded by #801.

PR #823, opened 2020-06-12
Add the timestamp of report generation and version

daltonkell:
Reports (all formats) now include the version of the Compliance Checker
as well as the date the report was run.

### DIFF
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -581,18 +581,25 @@ class CheckSuite(object):
         check_name = self._get_check_versioned_name(check_name)
         check_url = self._get_check_url(check_name)
         width = 2 * self.col_width
+
+        # NOTE: printing and use of .center()
+        # Nested .format() calls should be avoided when possible.
+        # As a future enhancement, a string.Template string might work best here
+        # but for the time being individual lines are printed and centered with
+        # .center()
+
         print("\n")
         print("-" * width)
-        print("{:^{width}}".format("IOOS Compliance Checker Report", width=width))
-        print("{:^{width}}".format("Version {}".format(__version__), width=width))
-        print("{:^{width}}".format("Report generated {}".format(
+        print("IOOS Compliance Checker Report".center(width))
+        print("Version {}".format(__version__).center(width))
+        print("Report generated {}".format(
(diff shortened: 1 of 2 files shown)


compliance_checker/suite.py
```diff
@@ -584,6 +584,10 @@ def standard_output(self, ds, limit, check_name, groups):
         print("\n")
         print("-" * width)
         print("{:^{width}}".format("IOOS Compliance Checker Report", width=width))
+        print("{:^{width}}".format("Version {}".format(__version__), width=width))
+        print("{:^{width}}".format("Report generated {}".format(
+            datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ")
```
ocefpaf, line 589: Came here to [paste this](https://blog.ganssle.io/articles/2019/11/utcnow.html) but I'm glad you probably already know about it!

daltonkell:
Actually @ocefpaf this is a great chance to ask -- how do we get the `pre-commit` tests to pass? I'm not familiar with them, but I looked at the build logs and they're quite verbose.

ocefpaf:
> Actually @ocefpaf this is a great chance to ask -- how do we get the `pre-commit` tests to pass?

Merge #756 :smile: 

> I'm not familiar with them, but I looked at the build logs and they're quite verbose.

In the GH action they show all the offending cases but if you run them locally most of them are automatically fixed. Just install pre-commit in your dev env and run `pre-commit run --all-files`. If you it to run with every commit, recommended but annoying, you have to run `pre-commit install`.

compliance_checker/suite.py
```diff
@@ -584,6 +584,10 @@ def standard_output(self, ds, limit, check_name, groups):
         print("\n")
         print("-" * width)
         print("{:^{width}}".format("IOOS Compliance Checker Report", width=width))
```
benjwadams, line 586: I realize this was legacy code included with the project, but I'm not really a fan of these repeated print functions and mutliple calls to format for centering text in a plaintext table.   For things like f-strings, multiple levels could even be an injection hazard with user supplied strings.

  Could you combine the text to be centered into a single multiline string and then call `str.center(width)` instead?

daltonkell: Should be fixed to satisfaction with latest commit

Issue #818, opened 2020-05-27
ioos 1.2 test: possible issue with multiple contributors in glabal attributes


ancillary:
using v. 4.3.3rc2+64.g71a4228 and running ioos:1.2 on a netcdf file with multiple contributors (e.g. [A0141.ocean.003m.merged.nc](http://gyre.umeoce.maine.edu/ncei/neracoos/A01/A0141.ocean.003m.merged.nc) 
generates Corrective Action Recommendations: 

contributor_email
* contributor_email must be a valid email address
contributor_role
* contributor_role should be from NERC or NOAA-NCEI

In this case contributor global attributes are:
:contributor_email = "info@neracoos.org,info@gmri.org,dan.codiga@mwra.org,croesler@bowdoin.edu" ;
:contributor_name = "NERACOOS,Gulf of Maine Research Institute,Massachusetts Water Resource Authority,Bowdoin University" ;
:contributor_role = "funder,distributor,funder,collaborator" ;
:contributor_role_vocabulary = "https://vocab.nerc.ac.uk/collection/G04/current/" ;
:contributor_url = "http://neracoos.org,http://www.gmri.org,http://www.mwra.org,http://www.bowdoin.edu" ;

I am not seeing any problems with punctuation, spelling, number of contributors, and I see the contributors in the vocabulary.   I have not been able to get any files with two or more contributors to pass without these messages.  

I see the implementation mentioned in https://github.com/ioos/compliance-checker/pull/739 but I wondered if anyone has successful examples of multiple contributors in a netcdf file.

ancillary:
let's make that subject refer to "global" rather than "glabal" attributes.  (sigh)

mwengren:
@ancillary I can duplicate these results with your source file on version 4.3.3.

We'll have to see how Compliance Checker handles verifying the content of those comma-separated attributes.  

Thanks for the report!

PR #830, opened 2020-07-08
Allow multiple contributor_role/contributor_role_vocabulary values

daltonkell:
The attributes contributor_role and contributor_role_vocabulary were
being checked as single values only. Instead, split the CSV string
and check each value given. Maintain inheritance model flexibility in
the Checker class by moving the attributes which the check validates
against into instance attributes. This enables the class to be extended
and attributes redfined without having to override the method.

Addresses #818.


compliance_checker/ioos.py
```diff
@@ -486,6 +486,44 @@ def __init__(self):
             )
         )
 
+        # geospatial vars must have the following attrs:
+        self.geospat_check_var_attrs = self._default_check_var_attrs
+
+        # valid contributor_role values
+        self.valid_contributor_roles = set(
+            [  # NERC and NOAA
```
benjwadams, line 494: Just use `{ }` here instead of constructing a list first.

benjwadams (changes requested):

PR #834, opened 2020-07-20
fixed "str has no kind attr"; fixed #809

### LINKED ISSUE #809: check_data_types exception during cf test: type object 'str' has no attribute 'kind'
Using version 4.3.3rc2+54.gb086b57, I am getting the following exception for all of our netcdf files under cf:1.6 and cf:1.7 test... an example netcdf file: 
http://gyre.umeoce.maine.edu/ncei/neracoos/A01/A0141.ocean.001m.merged.nc

WARNING: The following exceptions occured during the cf:1.7 checker (possibly in
dicate compliance checker issues):
cf:1.7.check_data_types: type object 'str' has no attribute 'kind'
  File "c:\apps\anaconda3\envs\cc-test-dev\lib\site-packages\compliance_checker-
4.3.3rc2+54.gb086b57-py3.8.egg\compliance_checker\cf\cf.py", line 1326, in check
_data_types
    if v.dtype.kind != "S" and all(

Any thoughts appreciated.

neumannd:


daltonkell:
Actually, it would also be great if you could add a quick unit test to confirm that string variables are not tested. It's not required, but could help identify unseen pitfalls or edge cases.

neumannd:
I will add one unit test later on today.

compliance_checker/cf/cf.py
```diff
@@ -1323,23 +1323,24 @@ def check_data_types(self, ds):
...
-                    np.dtype("i2"),
-                    np.dtype("i4"),
-                    np.float32,
-                    np.double,
-                )
-            ):
-                fails.append(
-                    "The variable {} failed because the datatype is {}".format(
-                        k, v.datatype
+            if v.dtype is not str:
```
lukecampbell, line 1326: I might be reading the indents wrong, but since this is an `if` statement followed by another `if` statement with no `else` conditions, I would recommend using a single `if` statement in lieu of nested `if`s to improve readability slightly.
```python
if v.dtype is not str \
        and v.dtype.kind != 'S' \
        and all(
            v.dtype.type != t
            for t in (
                np.character,
                np.dtype("|S1"),
                np.dtype("b"),
                np.dtype("i2"),
                np.dtype("i4"),
                np.float32,
                np.double
            )
        ):
    fails.append(
        "The variable {} failed because the datatype is {}".format(
            k,
            v.datatype,
        )
    )
```

neumannd: Thanks. Would be better to merge both if statements. I will update the PR.

neumannd:
There were some more locations in the code where string-type variables could cause issues. Therefore, I updated all location that I could find. I only looked for those in the CF Checker Plugin.

I didn't run `pre-commit` locally and, therefore, ran into several code-style issues. I'll do it in future to reduce the number of PR-updates.

compliance_checker/cf/cf.py
neumannd: PR updated accordingly

benjwadams:
This looks good, but I think it'd be good to squash down when merging.   Original issue occurs with variable length strings, which are not NumPy data types.

Issue #839, opened 2020-07-24
Add checks for new IOOS Metadata Profile 1.2 attributes: precision, accuracy, resolution, etc


mwengren:
We made some new additions to IOOS Metadata Profile 1.2.  Specific commit can be found [here](https://github.com/ioos/ioos-metadata/commit/1eb32fa679b1b015493e75383808a8920e42b94e) for reference.

As I mentioned in https://github.com/ioos/compliance-checker/issues/819#issuecomment-657763655, we should add new checks for the following attributes (all recommended):
* `geophysical_variable:accuracy`
* `geophysical_variable:precision`
* `geophysical_variable:resolution`
* `instrument_variable:calibration_date`
* `instrument_variable:make_model`
* `instrument_vocabulary`

daltonkell:
@mwengren In the specification, it is not mandated that `accuracy, precision, resolution` be numeric. Should this be tested for, or is it insignificant? Similarly, formats for `instrument_variable:calibration_date` (default ISO-8601?) and `instrument_variable:make_model` are not specified either.

mwengren:
@daltonkell 

For `accuracy, precision, resolution` I think we only state they be the same units as the variables themselves.  Can this be tested easily? @jessicaaustin @lukecampbell what do you think - single value in same units OK?

For `calibration_date` I think ISO-8601 is appropriate.  I'll add that to the description unless anyone objects to that change.

`make_model` I would expect to be string values.  We don't want to over prescribe here I think, so that field should probably not be checked for compliance with a data type or pattern.

daltonkell:
I think not over-prescribing is the right move :+1:

mwengren:
Updated `calibration_date` description.  Are we good with the others?

daltonkell:
Does "being the same units" mean "must include a units string"? E.g.

```
0.0025 degree_c
   |      |
quantity units
```

in which case the entire attribute would have to be a string.

mwengren:
Hmmm.  This may be getting over my head, but I assumed by saying that  we were intending the value of the attribute to be numeric, and the units to match those specified in the `units` attribute.

daltonkell:
Nope, that's what how I read it, just wanted to be crystal clear!

jessicaaustin:
@daltonkell @mwengren Sorry for the delay in commenting, I was out on vacation. Here's my two cents:

* `make_model` -- i agree this could be any arbitrary string, so the only validation i could think of is check that it's not empty if it's provided
[...]

PR #840, opened 2020-07-30
Add new geophysical attributes accuracy, precision, resolution

daltonkell:
Addresses https://github.com/ioos/compliance-checker/issues/839


daltonkell:
@ocefpaf It looks like AppVeyor failed to connect to GitHub -- likely a temporary issue.

https://ci.appveyor.com/project/ocefpaf/compliance-checker/builds/34392980

image


I notice that this seems to be your AppVeyor, or is that just how it's set up for the public repo? Don't know much about the AppVeyor configuration.

ocefpaf:
> I notice that this seems to be your AppVeyor, or is that just how it's set up for the public repo? Don't know much about the AppVeyor configuration.

We need to move away from AppVeyor, there is no granularity on the permissions and it is quite slow. Let me check if we should to Travis-CI Windows provider or something else.

compliance_checker/ioos.py
```diff
@@ -740,10 +743,39 @@ def check_geophysical_vars_have_attrs(self, ds):
...
+                if isinstance(val, Number):
+                    r = True
+                else:
+                    r = False
+
+                results.append(
+                    Result(
+                        BaseCheck.MEDIUM,
+                        r,
+                        f"geophysical_variable:{att}",
```
benjwadams, line 767: Use `str.format()` instead of f-strings for backwards compatibility with older Python 3 versions.

benjwadams (changes requested):

compliance_checker/ioos.py
ocefpaf: f-strings are new on Python 3.6 and, with most [scientific software adopting NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html), that Python is no longer supported. That bares the question should we support Python <=3.6?

benjwadams: Hmmm.  Not that long ago we dropped Python 2.  I wasn't aware of this NEP.  My NumPy version appears to be higher than the Python 3.6 supported version, so it's likely safe.  I don't think we've explicitly pinned a NumPy version, but rather it's going to be dependent on some of the other packages such as NetCDF.  To be on the safe side, I'd say convert to `str.format` for now, with a view towards supporting Python 3.6 as a minimum baseline in the near future.

benjwadams: Going to target 3.6+ from now on, so this is OK.

PR #841, opened 2020-07-30
Use GH actions instead of AppVeyor for Windows tests

ocefpaf:
This is just a test. I'm experimenting with an alternative for appveyor.

Travis-CI windows image did not work b/c it does not support Python and I could not download miniconda in it. If GH actions also fails we can try AzurePipellines.


ocefpaf:
I'm not an expert on GH actions I don't know if we need to merge this to get it to run :-/

ocefpaf:
@daltonkell the logs are a bit messy but the tests are running on Windows with GH actions. See https://github.com/ioos/compliance-checker/pull/841/checks?check_run_id=929156660#step:4:8154

ocefpaf:
> Nice use of the workflow, @ocefpaf. However, similar to the TravisCI build, should we test on multiple Python environments?

We can. I want to "reproduce" what we had in AppVeyor first but the main reason we only tested for latest Python was the slowness of AppVeyor. I'll send a few more commits here soon.

ocefpaf:
It is nice that the matrix items are shown here in the status. Maybe we should move everything to GH actions ;-p

PR #846, opened 2020-08-14
Small addition to neumannd drop zero dim vars for feature type search

daltonkell:
@neumannd put in a great PR in #835; this tweaks it ever so slightly to use a more idiomatic `numpy` expression per review of @benjwadams


neumannd:
Thanks for improving it.

PR #852, opened 2020-11-06
Integration test fixes

benjwadams:
Hopefully fixes outstanding integration test errors.


benjwadams:
Travis logs seem to indicate there are a few problems with the `test_protocols`:

https://travis-ci.org/github/ioos/compliance-checker/jobs/741968596
The Hyrax endpoint throws a 500 error.

The SOS response tested fails since `text/xml;charset=ISO-8859-1` is used rather than `text/xml`

PR #853, opened 2020-11-29
fix GitHub Actions on Windows

ocefpaf:
Using a more robust GH Action for conda.

### DIFF
--- a/.github/workflows/miniconda-win.yml
+++ b/.github/workflows/miniconda-win.yml
@@ -6,25 +6,30 @@ on:
     branches: [master]
 
 jobs:
-  windows:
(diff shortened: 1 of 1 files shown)


ocefpaf:
This will fix the Windows failure in #848.

I'm planning to move all testing to GitHub Actions in a future PR. Is that OK @benjwadams and @daltonkell? Travis-CI is dying slow death since it was bought and lately it has become quite unreliable.

benjwadams:
Sounds good to me, @ocefpaf.  Thanks for taking care of this.  Was going to ask about this, but looks like you've already got the fix in.

PR #862, opened 2021-05-14
run all tests on GitHub Actions

ocefpaf:
Just a draft for now. I want to add the downstream tests too and then remove the .travis.yml file.


ocefpaf:
This one is ready for review. The failures here are the same we have in current `master`.

pytest.ini
```diff
@@ -1,5 +0,0 @@
-[pylama]
-format = pylint
-skip = */doc/*,*/data/*
-linters = pep8,pyflakes
-ignore = E501,E251,E221,E201,E202,E203
```
ocefpaf, line 5: Need to migrate these to the pre-commit hooks. I'll  that in another PR and fix the pre-commits failures at the same time.

ocefpaf:
The NCEI plugin was failing before and it seems to be related with the plugin code and not compliance-checker. BTW, the plugins tests where "hidden" in Travis-CI under the allow failure. I'm not sure we can do that with GitHub Actions but I believe it is better this way, so the failures are explicitly when we decide to merge.

@daltonkell and @benjwadams this PR is ready for review!

benjwadams:
Looks good to me.

PR #864, opened 2021-05-18
fix failing tests

ocefpaf:
@daltonkell can you take a quick look? Should unblock the a few PRs.

### DIFF
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -829,7 +829,7 @@ class CheckSuite(object):
         # Issue GET instead if we reach here and can't get the response
         response = requests.get(ds_str, allow_redirects=True, timeout=60)
         content_type = response.headers.get("content-type")
-        if content_type == "text/xml":
+        if content_type.startswith("text/xml"):
             return self.process_doc(response.content)
         else:
             raise ValueError(
--- a/compliance_checker/tests/resources.py
+++ b/compliance_checker/tests/resources.py
@@ -16,7 +16,7 @@ def get_filename(path):
 
 
 def generate_dataset(cdl_path, nc_path):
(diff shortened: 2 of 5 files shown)


compliance_checker/suite.py
```diff
@@ -829,7 +829,7 @@ def load_remote_dataset(self, ds_str):
         # Issue GET instead if we reach here and can't get the response
         response = requests.get(ds_str, allow_redirects=True, timeout=60)
         content_type = response.headers.get("content-type")
-        if content_type == "text/xml":
+        if content_type.startswith("text/xml"):
```
ocefpaf, line 832: This was failing in cases where the string is `text/xml;charset=ISO-8859-1`.

compliance_checker/tests/resources.py
```diff
@@ -16,7 +16,7 @@ def get_filename(path):
 
 
 def generate_dataset(cdl_path, nc_path):
-    subprocess.call(["ncgen", "-o", nc_path, cdl_path])
+    subprocess.call(["ncgen", "-4", "-o", nc_path, cdl_path])
```
ocefpaf, line 19: We can either pin out netcdf-c to <4.8.0 or use `-4` everywhere.

ocefpaf:
Looks like Windows is not OK with the changes here :-/, investigating...

compliance_checker/tests/resources.py
daltonkell: I'm in favor of keeping this more explicit - especially since there are some odd Windows path issues with libnetcdf-4.8.0 currently.

ocefpaf: > I'm in favor of keeping this more explicit

What do you mean? Pin to `<4.8.0` for now?

ocefpaf:
I move the commits here to #862. Closing this.

Issue #842, opened 2020-08-05
CF: incorrect warnings for variable attributes scale_factor and add_offset


mhidas:
In #646 a new [check](https://github.com/ioos/compliance-checker/pull/646/files#diff-f14722526b669b2144710ff4feeff8d4R419) was added to the CF suite to check that certain variable attributes have the same data type as their parent variable. I'm pretty sure the attributes `scale_factor` and `add_offset` should **not** be included in this check.

The [NetCDF User Guide](https://www.unidata.ucar.edu/software/netcdf/docs/attribute_conventions.html) (referenced in the doc string) says
> When scale_factor and add_offset are used for packing, the associated variable (containing the packed data) is typically of type byte or short, whereas the unpacked values are intended to be of type float or double. The attributes scale_factor and add_offset should both be of the type intended for the unpacked data, e.g. float or double.

This means in general these two attributes (type float/double) will be of a different type to the variable (byte/short).

CF section 8.1 [Packed Data](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#packed-data) restricts this somewhat, but still allows the attributes to have different data type than the variable.
> However, if the scale_factor and add_offset attributes are of a different data type from the variable (containing the packed data) then the unpacked data should match the type of these attributes, which must both be of type float or both be of type double . An additional restriction in this case is that the variable containing the packed data must be of type byte , short or int . It is not advised to unpack an int into a float as there is a potential precision loss.

mhidas:
Sorry I missed this earlier. Haven't had much time to look at checker issues in the past year. We're finally looking to upgrade to version 4.3.x (still using 4.1.1. now), so now we're testing it against various IMOS data.

daltonkell:
Thanks for pointing this out, @mhidas. So really, the check should be specifically for `scale_factor` and `add_offset`:

```
if (scale_factor_dtype == packed_var_dtype) and (add_offset_dtype == packed_var_dtype):
    continue # all good
else:
    # check dtype of scale_factor and add_offset must be both byte or float
```

and repeat for `add_offset`. I think this is more accurate to the specification. Any thoughts?

mhidas:
Yes, something like that, except in the `else` clause above, you should
* check that `scale_factor` and `add_offset` are both of type **float** or both be of type **double**; and also
* check that the variable with these attributes is of type byte, short or int

benjwadams:
[...]

PR #869, opened 2021-07-13
Relax type checking with scale_factor and add_offset variable attributes

benjwadams:
Don't require exact match for `scale_factor` and `add_offset` attributes as compared WRT parent variable.


benjwadams:
Related: #842

Issue #845, opened 2020-08-13
CF: incorrect errors for ragged array structure


mhidas:
We have some files structured according to the [CF indexed ragged array representation](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_indexed_ragged_array_representation). The basic structure looks like this:
```
netcdf indexed_ragged_array {
dimensions:
        OBSERVATION = 12 ;
        INSTRUMENT = 3 ;
variables:
        double TIME(OBSERVATION) ;
                TIME:axis = "T" ;
        float TEMP(OBSERVATION) ;
                TEMP:coordinates = "TIME LATITUDE LONGITUDE NOMINAL_DEPTH" ;
        int instrument_index(OBSERVATION) ;
                instrument_index:instance_dimension = "INSTRUMENT" ;
                instrument_index:long_name = "which instrument this obs is for" ;
        double LATITUDE(INSTRUMENT) ;
                LATITUDE:axis = "Y" ;
        double LONGITUDE(INSTRUMENT) ;
                LONGITUDE:axis = "X" ;
        float NOMINAL_DEPTH(INSTRUMENT) ;
                NOMINAL_DEPTH:axis = "Z" ;
}
```
As far as I can tell, this is a correct implementation of this valid CF representation.
When I run the CF checker on it, I get the following errors and warnings:
```
                                    Errors                                     
--------------------------------------------------------------------------------
§5 Coordinate Systems
* dimensions for auxiliary coordinate variable LATITUDE (INSTRUMENT) are not a subset of dimensions for variable TEMP (OBSERVATION)
* dimensions for auxiliary coordinate variable LONGITUDE (INSTRUMENT) are not a subset of dimensions for variable TEMP (OBSERVATION)
* dimensions for auxiliary coordinate variable NOMINAL_DEPTH (INSTRUMENT) are not a subset of dimensions for variable TEMP (OBSERVATION)

§9.1 Features and feature types
* Unidentifiable feature for variable TEMP
* Unidentifiable feature for variable instrument_index


                                    Warnings                                    
--------------------------------------------------------------------------------
§3.1 Units
* units (None) attribute of 'instrument_index' must be a string compatible with UDUNITS
```
The `Unidentifiable feature` seems to be a known issue (#719) not specific to this structure.
The other two are checks that need to make an exception for this kind of file.

mhidas:
Note that the indexed ragged array representation should be easy to identify, as stated in CF:
[...]

PR #870, opened 2021-07-15
Refactor featur type checking update tests

openSourcerer9000:
Here I added some integration tests for issue #845, which Dalton fixed in the refactor_featureType branch (has been merged).

There were 5 failing integration tests, each checking for multiple things at once. Isolating which message they were failing on, I saw that they all failed a check for a message produced by the check_all_features_are_same_type method, which was removed in this branch. I removed those messages from the checklist, and added an explicit diff for when multiple messages get checked. This way no one will have to go rummaging around in the code looking for which piece of the test tripped the failure.

Before:

```
E       AssertionError: assert False
E        +  where False = all(<generator object TestCFIntegration.test_bad_cf_roles.<locals>.<genexpr> at 0x0000026DA0F16820>)

compliance_checker\tests\test_cf_integration.py:412: AssertionError
```

After (with regression inserted):
```
E       AssertionError: Failed to find the following messages:
E         GRatitude variable 'lat' should define standard_name='latitude' or axis='Y'
E
E                 These were the messages captured:
E         §2.6.2 global attribute title should exist and be a non-empty string
E         §2.6.2 global attribute history should exist and be a non-empty string
E         §2.6.1 Conventions field is not present
E         Attribute long_name or/and standard_name is highly recommended for variable lat
E         Attribute long_name or/and standard_name is highly recommended for variable lon
E         Attribute long_name or/and standard_name is highly recommended for variable time
E         latitude variable 'lat' should define standard_name='latitude' or axis='Y'
E         longitude variable 'lon' should define standard_name='longitude' or axis='X'
E                     Please check wording and section names if messages have been altered since this test was written
E       assert False
E        +  where False = all([True, True, True, True, True, True, ...])

compliance_checker\tests\test_cf_integration.py:341: AssertionError
```

A more permanent solution could be something like a special message object, whose __repr__ would be the message to display, but would also have a message ID associated with it, which could be easily testable and wouldn't cause problems if wording changes. At this point that may not be worth the trouble, but just a thought I wanted to throw out.


.vscode/settings.json
```diff
@@ -0,0 +1,10 @@
+{
+    "workbench.colorTheme": "Solarized Dark",
+    "workbench.colorCustomizations": {
+        "activityBar.foreground": "#ff0000",
+        "activityBarBadge.foreground": "#000000",
+        "activityBarBadge.background": "#ff000095",
+        "sideBar.background": "#200000",
+    },
+    "python.pythonPath": "C:\\Users\\sean.micek\\AppData\\Local\\Continuum\\anaconda3\\envs\\cchecka\\python.exe"
+}
```
benjwadams: Could you remove this file and squash the commit down?

benjwadams (changes requested):

.vscode/settings.json
openSourcerer9000: Ben, I had removed that file and added to gitignore before this latest PR https://github.com/ioos/compliance-checker/pull/870/files

benjwadams: OK, I'm going to squash down these commits and merge.

Issue #868, opened 2021-07-12
Move plugin tests to respective repositories


benjwadams:
There are a number of unit tests related to other compliance-checker plugins, such as the NCEI and Glider plugins which are located in the main compliance checker repo.   These should be moved to their respective repos so that any test issues can be addressed there, rather than in the core repo.

Also, there are failing tests for the glider plugin on this repo, but I am not able to reproduce them from a install using pip.

ocefpaf:
@benjwadams I'll let you decide b/c the burden to maintain is on you. However, the plugin tests here are more on an "informative note" than a real failure for compliance-checker. It is mostly a "downstream test." The plugins do have an "upstream test" too.

With that said there is a bug in the testing, here it should be against tagged releases of the plugins, downstream it should be with development version of compliance-checker. Otherwise it does not make sense. I'll fix that ASAP.

benjwadams:
Related: #763

benjwadams:
OK, @ocefpaf, I see some of the rationale behind this now.  The downstream tests are failing.  The test failures look to be related to the `setup.py` entry points.  Namely, the CI setup with GitHub actions appear to not be finding them, whereas as previously mentioned, a local build using `pip` does find them.  With recent pull request merges, the downstream NCEI checks are the only remaining outstanding CI failure, so I'd like to either remove them or find out what's occurring that's causing the NCEI check suites not to be registered.

Issue #763, opened 2020-03-12
Downstream plugin failures


ocefpaf:
The following plugins are failing with the latest 4.3.3 RC release:

- glider
- ncei
- ugrid

only sgrid is passing.

I did not inspect if the API changes broke the plugins and/or if the plugins just need some updates. More info on https://travis-ci.org/github/ioos/compliance-checker/builds/650628887

cc @mwengren

benjwadams:
I think some of this may have to do with changes to the class hierarchy introduced when the CF 1.7 checker was added -- the class hierarchy changed.

benjwadams:
@ocefpaf, the glider plugin, does have the issue I mention above along with the referenced fix.

I can't reproduce issues for ncei and ugrid on my development virtual environment running Python 3.6.  The tests pass and the few files I tried running on seem to work OK.  Any suggestions on reproducing the test failures?

ocefpaf:
> Any suggestions on reproducing the test failures?

The tests are run on a git checkout of latest master of the plugin and run it against compliance-checker we are testing. Here are the CI lines:

https://github.com/ioos/compliance-checker/blob/master/.travis.yml#L73-L77


I tried, from the compliance-checker work folder,

```
git checkout 4.3.3rc1
conda create --name CC python=3.6 --file requirements.txt --file test_requirements.txt
conda activate CC
```

and then, from the `cc-plugin-ncei` work folder,

```
conda install --file requirements.txt --file requirements-dev.txt
pytest -s -rxs -v
```

I was able to reproduce the same failures with this setup.

benjwadams:
I'm not presently using conda, but I wasn't able to reproduce this when checked out against 4.3.3rc1, doing `python setup.py develop`, and running the tests.

I'll see if I can test this out under conda.

ocefpaf:
> I'll see if I can test this out under conda.

It should not matter unless we have some broken packages. I'll try the pip route too and see if I can debug that.

PS: the only reason I use `conda` is b/c that is the recommended way to get `cf_units` b/c of `udunits` and b/c it is what we use on the CIs.

daltonkell:
I tried to debug using `pip` as well:

```
$ git checkout 4.3.3rc1
$ python3.8 -m venv CC_BUG
$ pip install -r requirements.txt -r requirements-dev.txt
$ pytest -s -rxs -v
```

for glider, ncei, and ugrid; for sgrid:

```
$ python3.8 -m venv CC_BUG
[...]

PR #873, opened 2021-07-21
Remove NCEI check from GitHub Actions

benjwadams:
Removes routinely failing NCEI plugin check for the time being.


benjwadams:
Related to #868, #763

Issue #877, opened 2021-08-02
Files incorrectly identified as ragged array


mhidas:
I'm testing the latest code on master (ff93392a), running the `cf:1.6` checks against some of our files. 

For [this file](https://s3-ap-southeast-2.amazonaws.com/imos-data/IMOS/SOOP/SOOP-XBT/DELAYED/Line_IX28_Dumont-d-Urville-Hobart/2019/IMOS_SOOP-XBT_T_20190108T073100Z_IX28_FV01_ID-89011149.nc), which certainly _is_ a `profile`, I get the following warning:
```
§9.1 Features and feature types
* TEMP is not a profile, it is detected as a ragged-array
```
Also found similar warnings for [timeseries](https://s3-ap-southeast-2.amazonaws.com/imos-data/IMOS/SOOP/SOOP-TMV/VLST_Spirit-of-Tasmania-1/realtime/mooring/1sec/2019/IMOS_SOOP-TMV_TSUB_20190101T074611Z_VLST_FV00_mooring-DEV_END-20190101T101530Z.nc) and [trajectory](https://s3-ap-southeast-2.amazonaws.com/imos-data/IMOS/SOOP/SOOP-TMV/VLST_Spirit-of-Tasmania-1/realtime/transect/1sec/2019/IMOS_SOOP-TMV_TSUB_20190101T101531Z_VLST_FV00_transect-D2M_END-20190101T192740Z.nc) files, which are also mis-identified as `ragged-array`.

I think the root of the issue is a confusion between the [feature type](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#_features_and_feature_types) of a file or variable (which refers to the geographic and temporal distribution of the physical values being reported), and the [representation](http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html#representations-features) (ragged array or otherwise - which refers to the structure of the file). 

None of the above linked files are actually ragged array representations, but even if they were, their correct feature type would still be profile, timeseries or trajectory. So the checks need to consider these two concepts independently.

benjwadams:
I think we might want to do away with the following lines entirely, as "ragged array" is not a CF feature type.

https://github.com/ioos/compliance-checker/blob/master/compliance_checker/cf/cf.py#L4444-L4445

benjwadams:
@mhidas, please check with the recently merged PR.   I tested the file and the erroneous error no longer displays.

[...]

PR #878, opened 2021-08-03
Remove ragged-array from featureType determination routines

benjwadams:


benjwadams:
Related to #877

openSourcerer9000:
@benjwadams this seems to fix the issue from @mhidas 's dataset. I ran it with the test I had written and it passed. Since I couldn't ncdump a valid CDL for some reason, I just used it directly on his .nc instead of letting the repo ncgen a new one. 

I'm going to go ahead and leave that test behind, this one seems pretty straightforward and we don't want those strange UTF errors from that CDL causing problems down the road.

Issue #874, opened 2021-07-29
Strip unittest dependency to fully upgrade tests to pytest, so we can provide better quality and coverage of tests


openSourcerer9000:
As things are currently, many of our tests are trying to test multiple cases per test, or don't fail gracefully, providing enough info to diagnose the problem. I believe we can give them a makeover if we can upgrade them to the pytest framework, which will increase our confidence in the codebase and new PR's that pass all tests. Pytest has all the features built in that can easily handle a lot of these common testing blues.

Since we are using pytest, we can run the legacy unittests, but they can't leverage pytest features like fixtures or parameterization, as they are all inheriting from the incompatible unittest TestCase object. 
From a discussion with @benjwadams:

https://github.com/benjwadams/compliance-checker/pull/4
```
[10:11 AM] Sean Micek    
    Ben Adams I noticed all of our test objects inherit from unittest TestCase but we have pytest now, is there any reason we couldn't upgrade them at some point so we can use pytest fixtures/parameterization?
Edited​[10:13 AM] Ben Adams
    Given that TestCase works with pytest, there's no reason why we couldn't change over parts of the library
​[10:14 AM] Sean Micek
    How I understand it, pytest made it backwards compatible, so if you want to switch it will still run all your old unittests. I know nothing about unittest, but when I tried adding pytest params I got an error complaining that the test was a unittest TestCase
​[10:16 AM] Sean Micek
    I guess it all stems from tests/_.init_.py
I would think we could just drop the unittest inheritance and it should work, pytest should run any method in the tests files with test_ prefix
```
https://github.com/ioos/compliance-checker/blob/e14be12c130d1100c6a6a7d700c539106049dc5e/compliance_checker/tests/__init__.py#L1-L9

I tried simply removing the TestCase inheritance from the BaseTestCase object, and got errors as all these unittest setUp and tearDown helper functions stopped working without the dependency. We can convert them all to pytest fixtures, but beyond that I'm not sure what other features we may be using from unittest, I'm just not familiar with the library. 

This probably isn't an urgent upgrade, but I feel that it would allow us to diagnose and patch issues more quickly and merge PR's more confidently.

![image](https://user-images.githubusercontent.com/61931826/127524607-8d434674-9039-49c6-b093-0c6f2e71b4d8.png)
![image](https://user-images.githubusercontent.com/61931826/127527977-d5dbb4ba-b7c8-4293-a38c-f051f2df5b24.png)

benjwadams:
[...]

PR #880, opened 2021-08-27
Remove unittest dependency for CF Integration tests

openSourcerer9000:
I've begun the migration away from unittest towards pure pytest, for reasons outlined in #874 . I didn't have the stomach to refactor the ~7K lines of test code all at once, so I've only upgraded test_cf_integration.py for now, as it's the main file I imagine we'll be using to patch test data from github issues as they come in. We can upgrade more test classes as needed when we find ourselves writing new tests in them. 

I've replaced the hardcoded resources.py file with a simple function, abstracted away in a pytest.fixture. There is now an assertion in there that ncgen was successful as well, making it easy to diagnose in case we have hangups with it again in the future. 

In order to add a new CDL and integration test, the CDL can simply be placed somewhere below the `tests/data` directory and the list `dataset_stem__expected_messages` appended with a tuple of the CDL's file stem (which will be become the test case's id), and the set of expected messages to test for:

```
#test_cf_integration.py

dataset_stem__expected_messages = [
    (
        "sldmb_43093_agg",
        [
            u"attribute time:_CoordianteAxisType should begin with a letter and be composed of letters, digits, and underscores",
            u"attribute lat:_CoordianteAxisType should begin with a letter and be composed of letters, digits, and underscores",
            u"attribute lon:_CoordianteAxisType should begin with a letter and be composed of letters, digits, and underscores",
            u"§2.6.2 global attribute history should exist and be a non-empty string",
            u"standard_name temperature is not defined in Standard Name Table v{}".format(
                std_names._version
            ),
            u"temperature's auxiliary coordinate specified by the coordinates attribute, precise_lat, is not a variable in this dataset",
            u"temperature's auxiliary coordinate specified by the coordinates attribute, precise_lon, is not a variable in this dataset",
        ],
    ),
    (
        "usgs_dem_saipan",
        ['§2.6.1 Conventions global attribute does not contain "CF-1.7"'],
    ),

...
```
The parameterized test cases will report like this:
```
test_cf_integration.py::TestCFIntegration::test_cf_integration[sldmb_43093_agg-expected_messages0] PASSED     [  5%]
test_cf_integration.py::TestCFIntegration::test_cf_integration[usgs_dem_saipan-expected_messages1] PASSED     [ 11%]
test_cf_integration.py::TestCFIntegration::test_cf_integration[l01-met-expected_messages2] PASSED             [ 17%]
test_cf_integration.py::TestCFIntegration::test_cf_integration[sp041-expected_messages3] PASSED               [ 23%]
test_cf_integration.py::TestCFIntegration::test_cf_integration[3mf07-expected_messages4] PASSED               [ 29%]
test_cf_integration.py::TestCFIntegration::test_cf_integration[ooi_glider-expected_messages5] PASSED          [ 35%]
test_cf_integration.py::TestCFIntegration::test_cf_integration[swan-expected_messages6] PASSED                [ 41%]
test_cf_integration.py::TestCFIntegration::test_cf_integration[kibesillah-expected_messages7] PASSED          [ 47%]
test_cf_integration.py::TestCFIntegration::test_cf_integration[pr_inundation-expected_messages8] PASSED       [ 52%]
test_cf_integration.py::TestCFIntegration::test_cf_integration[ww3-expected_messages9] PASSED                 [ 58%]
test_cf_integration.py::TestCFIntegration::test_cf_integration[glcfs-expected_messages10] PASSED              [ 64%]
test_cf_integration.py::TestCFIntegration::test_cf_integration[bad_cf_role-expected_messages11] PASSED        [ 70%]
test_cf_integration.py::TestCFIntegration::test_cf_integration[ocos-expected_messages12] PASSED               [ 76%]
```
Conversely, there is a second parameterized test function for asserting that incorrect messages don't come up. 

As you can see, not only does pytest reduce repetitive code, but it also reduces repetitive code. The only caveat is if there was anyone still using the unittest runner. If it was still compatible with unittest before, it won't be anymore. In order to run the test suite, devs (not users) will need to pip/conda install pytest, cd to the repo, and run `$ pytest` .


compliance_checker/tests/conftest.py
```diff
@@ -0,0 +1,106 @@
...
+def flatten_list(listOfLists):
+    """or list of tups"""
+    return reduce(iconcat, listOfLists, [])
+
+
+def glob_down(pth, suffix, lvls):
+    """globs down up to (lvls: int) levels of subfolders\n
+    suffix in the form ".ipynb"\n
+    pth: Path"""
+    return flatten_list([list(pth.glob(f'*{"/*"*lvl}{suffix}')) for lvl in range(lvls)])
```
benjwadams, line 26: Minor point, but if this is a list of lists, why not `itertools.chain(*pth.glob...)` here?

benjwadams (commented):
Looks much cleaner overall and a lot less boilerplate.

compliance_checker/tests/conftest.py
openSourcerer9000: Ok cool, I didn't realize that does the same thing; I just switched it to chain.

PR #885, opened 2021-09-29
CF 1.8 geometries

benjwadams:
Draft implementation of CF 1.8 geometries.  @greenteadigital, please review both this and the relevant spec section to see if it's up to snuff.  Please don't merge until we decide all's OK.

PR #888, opened 2021-10-29
Debug gha

ocefpaf:
Looks like we have conflicts and some path issues.


benjwadams:
I was going to do `develop` branch for new CF versions, but we can move fixes into the main branch for now.

ocefpaf:
> I was going to do `develop` branch for new CF versions, but we can move fixes into the main branch for now.

This is only to try to debug this. I'll send a commit to the PRs you have open if I can find the conflict and we can close this one.

ocefpaf:
We had to update our libnetcdf and to do that we need to regenerate the remaining netcdf files and add te `"-4` option to the ncgen calls. I added the commits here to #885. Hope that is fix is OK to move us forward.

Issue #806, opened 2020-05-08
Fuzzy search/recommendations for close matches


benjwadams:
It'd be nice to have something like a fuzzy search/suggestion on certain checks that take lists of values, such as the check for CF standard names.   In this way, if someone put "sea_water_temp" as a standard name, "sea_water_temperature" would be suggested in the error message as a correction.  @Bobfrat has already sent me a snippet using the `fuzzywuzzy` fuzzy string search/Levenshtein distance library that was used for a limited subset of the CF standard names.

   This would be useful in other checkers as well, such as the Glider DAC checker, where the are often slight variations of project names and institutions in submitted metadata which differ from the forms used by the NCEI authority tables.

benjwadams:
Closing as implemented by #925 for CF standard names.  Other functionality for suggestions can be made on a case-by-case basis for individual checkers.

PR #889, opened 2021-11-11
Taxa names

benjwadams:
Draft commit for taxa names, Section 6.1.2 of CF Conventions 1.8 document.

Done/To-do:

- [x] Implementation of LSID, WoRMS, and ITIS data checking
- [x] Unit tests for "happy path" on WoRMS LSID using mocked network calls.
- [x] Unit tests for LSIDs referring to ITIS
- [x] Unit test for nodata LSID values (skip the check, can be used if taxon is not present in databases)
- [x] Unit tests on not-so-happy-paths such as badly formed LSIDs, nonexistent IDs, non-matching taxon names when compared with database contents, etc.
- [ ] (*optional for now*) Integration tests with a "gold standard" bio data file and network calls to lsid.info, WoRMS, ITIS, etc.  (@MathewBiddle, if you know where to find good datasets for this purpose, please let us know)


compliance_checker/cf/cf_18.py
```diff
@@ -2,11 +2,17 @@
...
 from compliance_checker.base import BaseCheck, BaseNCCheck, Result, TestCtx
+import requests
+from lxml import etree
+import pyworms
 from shapely.geometry import Polygon
 import numpy as np
-from compliance_checker.cf.util import reference_attr_variables
+import re
+from compliance_checker.cf.util import (reference_attr_variables,
+                                        VariableReferenceError)
```
benjwadams, line 15: Extraneous, probably can remove.

compliance_checker/cf/util.py
```diff
@@ -573,6 +573,16 @@ def is_vertical_coordinate(var_name, var):
         satisfied |= getattr(var, "positive", "").lower() in ("up", "down")
     return satisfied
 
+def string_from_var_type(variable):
```
benjwadams, line 576: This is needed for char arrays where the `_Encoding` attribute is not declared, which is virtually always, in my experience.   Still, assuming UTF-8 may have pitfalls, so if you have better ideas please let me know.

compliance_checker/cf/cf_18.py
```diff
@@ -2,11 +2,17 @@
 from compliance_checker.base import BaseCheck, TestCtx, Result
 from compliance_checker import MemoizedDataset
 from compliance_checker.cf.cf import CF1_7Check
+from cf.util import reference_attr_variables, string_from_var_type
 from netCDF4 import Dataset
 from compliance_checker.base import BaseCheck, BaseNCCheck, Result, TestCtx
+import requests
+from lxml import etree
+import pyworms
```
benjwadams, line 10: I'm not sure whether to use `pyworms` or stick to making REST API calls against the WoRMS services directly with `requests`.  `pyworms` does use `requests` under the hood and it has some basic caching functions.  `pyworms` uses the `lru_cache` decorator, but we could in practice cache HTTP requests with a package such as `requests-cache`.

compliance_checker/cf/cf_18.py
```diff
@@ -132,7 +140,189 @@ def check_taxa(self, ds: MemoizedDataset):
...
+
+
+        for taxon_quantifier_variable in taxa_quantifier_variables:
+            valid_taxa = TestCtx(BaseCheck.HIGH, self.section_titles["6.1.2"])
+            if not isinstance(getattr(taxon_quantifier_variable,
+                                      "coordinates", None), str):
+                valid_taxa.add_failure(f"{taxon_quantifier_variable.name} must have a string valued \"coordinates\" attribute")
+                continue
+
+            #coord_var_names = taxon_quantifier_variable.coordinates.split(" ")
```
benjwadams, line 178: Delete

compliance_checker/cf/cf_18.py
```diff
@@ -132,7 +140,189 @@ def check_taxa(self, ds: MemoizedDataset):
...
+                continue
+
+            coordinate_vars = [ds.variables[var_name] for var_name in
+                                coordinate_var_names]
+
+
+            coord_var_standard_names = {var:
+                                        getattr(var, 'standard_name', None)
+                                        for var in coordinate_vars}
+            #coord_vars = reference_attr_variables(ds, "coordinates", " ")
```
benjwadams, line 197: Delete

benjwadams (commented):
General comments on the clarity of messages returned in error cases would be appreciated as well.

benjwadams:
Blah, `pyworms` doesn't appear to be available on PyPI and the Git repo syntax in `requirements.txt` is being pulled in to `setup.py`, where it appears it's invalid syntax.  See also comment above about whether to use `pyworms` or not.

greenteadigital:
You can coment it out of requiremets.txt, and put it in setup.py. I forget the syntax though, but it's doable.

compliance_checker/cf/cf_18.py
```diff
@@ -132,7 +140,189 @@ def check_taxa(self, ds: MemoizedDataset):
...
+            # if we have no authority, we can't check validity of the name -- assume it's OK
+            standard_name_set = set(coord_var_standard_names.values())
+            if set(coord_var_standard_names.keys()) == {"biological_taxon_name"}:
+                # TODO: Check for at least binomial nomenclature?
```
benjwadams, line 201: Should WoRMS/ITIS still be checked if a variable with "biological_taxon_lsid" is not specified?   The CF Standard Name Table entry mentions in the entry for `biological_taxon_name`:

> "Biological taxon" is a name or other label identifying an organism or a group of organisms as belonging to a unit of classification in a hierarchical taxonomy. The quantity with standard name biological_taxon_name is the human-readable label for the taxon such as Calanus finmarchicus. The label should be registered in either WoRMS (http://www.marinespecies.org) or ITIS (https://www.itis.gov/) and spelled exactly as registered. See Section 6.1.2 of the CF convention (version 1.8 or later) for information about biological taxon auxiliary coordinate variables.

However, the CF interpretation at https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#taxon-names-and-identifiers seems a little less strict:

> It is strongly recommended in CF that the authority chosen is World Register of Marine Species (WoRMS) for oceanographic data and Integrated Taxonomic Information System (ITIS) for freshwater and terrestrial data. WoRMS LSIDs are built from the WoRMS AphiaID taxon identifier such as "urn:lsid:marinespecies.org:taxname:104464" for AphiaID 104464. This may be converted to a URL by adding prefixes such as ​http://www.lsid.info/. ITIS LSIDs are built from the ITIS Taxonomic Serial Number (TSN), such as "urn:lsid:itis.gov:itis_tsn:180543".

> The biological_taxon_name auxiliary coordinate variable included for human readability is mandatory. The biological_taxon_lsid auxliary coordinate variable included for software agent readability is optional, but strongly recommended. If both are present then each biological_taxon_name coordinate must exactly match the name resolved from the biological_taxon_lsid coordinate. If LSIDs are available for some taxa in a dataset then the biological_taxon_lsid auxiliary coordinate variable should be included and missing data given for those taxa that do not have an identifier.

benjwadams: Should we also search for the value(s) of `biological_taxon_name` in WoRMS and ITIS if `biological_taxon_lsid` is not present?

compliance_checker/cf/cf_18.py
```diff
@@ -132,7 +140,189 @@ def check_taxa(self, ds: MemoizedDataset):
...
+            #coord_var_names = taxon_quantifier_variable.coordinates.split(" ")
+            coordinate_var_names = (taxon_quantifier_variable.coordinates
+                                    .split(" "))
+            invalid_coord_vars = set(coordinate_var_names) - ds.variables.keys()
+            if invalid_coord_vars:
+                valid_taxa.add_failure("The following values for \"coordinates\" attributes were not found in the dataset's variables "
+                                       f"{invalid_coord_vars}")
+
+            if len(coordinate_var_names) > 2:
+                valid_taxa.add_failure("coordinates attribute for taxon data must either reference one or two varaible name")
```
benjwadams, line 187: variable name**s**

MathewBiddle:
@benjwadams once we finalize the [ATN template](https://github.com/ioos/ioos-atn-data/blob/main/templates/atn_trajectory_template.cdl) that should be a good gold standard to work with. I'll keep an eye out for others.

compliance_checker/cf/cf_18.py
MathewBiddle: I think so... But, keep in mind it has to be spelled exactly as registered in ITIS or WoRMS. 

Also wondering how computationally expensive this check could be. Thinking of the case where there's hundreds of species to look up in both services. 

Maybe this [global names resolver](http://resolver.globalnames.org/api) example can help? https://gist.github.com/adyork/5983a1ed9763ffb37fb7cd15df24e895#file-match_names_gnrd-ipynb

benjwadams: The API looks interesting, especially the fuzzy matching.  I mentioned this as a possible desirable feature for CF standard names in #806.  The API looks to be alpha based upon the link, but I'll keep an eye on it.  In general, caching might be possible for web requests.  On one of these I also saw that there was a database dump, but it was about 173 MiB compressed, so that's rather a rather large file, given that many users aren't even checking bio data.

benjwadams:
@greenteadigital, this is set for further review.

compliance_checker/cf/cf_18.py
```diff
@@ -132,7 +137,203 @@ def check_taxa(self, ds: MemoizedDataset):
...
+                json_contents = itis_response.json()
+                combined_name = json_contents["scientificName"]["combinedName"]
+
+                if taxon_name_str != combined_name:
+                    messages.append("Supplied taxon name and ITIS scientific name do not match. "
+                                    f"Supplied taxon name is '{taxon_name_str}', ITIS scientific name "
+                                    f"for TSN {taxon_match['object_id']} is '{combined_name}.'")
+
+            else:
+                warnings.warn("Compliance checker only supports checking valid "
```
greenteadigital, line 328: `warnings` is not defined

compliance_checker/tests/test_cf.py
```diff
@@ -38,6 +38,9 @@
     MockTimeSeries,
     MockVariable,
 )
+import requests_mock
```
greenteadigital, line 41: module not found after `pip install`ing requirements

greenteadigital: oops, my bad, need to install `test_requirements`, got it

greenteadigital (approved):

PR #891, opened 2021-11-18
Fix exclude logic

benjwadams:


benjwadams:
This functionality is needed for some upcoming Glider DAC checks.

benjwadams:
TODO: also need to document interface changes to `suite.py` somewhere aside from docstrings for plugin developers.  The `run` function still exists, but throws a deprecation warning and makes a call to the new `run_all` function.   I could not add in the include checks to the `run` function without fundamentally breaking the interface of that function.

Issue #641, opened 2019-04-11
Add test coverage report


benjwadams:
It'd be nice to have a test coverage report so we could see which checks have adequate test case coverage.

benjwadams:
Closed by #900

PR #892, opened 2021-12-01
Merge in CF 1.8 implementation

benjwadams:
CF 1.8 implementation.   Needs to be merged, along with any desired changes, prior to next release.


greenteadigital (commented):
cf.py:4102
"boundary_variable" is not defined

ioos.py:959
"good_result" is not defined

benjwadams:
> cf.py:4102
   "boundary_variable" is not defined

These were introduced in other commits unrelated to CF 1.8.  I am very much in favor of fixing these and adding the proper test coverage where needed prior to the next tagged release, even possibly revisiting #641.  However, I think they should be dealt with in a separate issue, as the commit change set for this PR didn't add them.

In terms of splitting out the CF versions into separate modules, I'm in favor of it overall as long as we still import the classes to the `cf.py` module.  There are a number of plugins for compliance checker which rely on the classes being importable from that module.

greenteadigital:
OK, I can live w/ that.

.github/workflows/default-tests.yml
```diff
@@ -26,7 +26,7 @@ jobs:
     - name: Python ${{ matrix.python-version }}
       shell: bash -l {0}
       run: |
-        conda create --name TEST python=${{ matrix.python-version }} pip --file requirements.txt --file test_requirements.txt --strict-channel-priority
+        conda create --name TEST python=${{ matrix.python-version }} pip "libnetcdf<4.8.0" --file requirements.txt --file test_requirements.txt --strict-channel-priority
```
ocefpaf, line 29: We should probably revert this change. We need to unpin libnetcdf.

benjwadams: Sorry, I accidentally merged in some stuff related to 1.9 into develop, which I didn't mean to do.  Some of the commits above are my attempts to rectify some of the issues.  Others are just from merging master.  I'll unpin this.

benjwadams:
@greenteadigital , will break apart the CF modules once I'm satisfied tests are working again.  Some of the integration tests seem to have had URLs change and now cause failures.

benjwadams:
Looks like something is breaking in `ncgen` for generating string variables.

benjwadams:
OK @greenteadigital, I have refactored the versioned CF checker classes into their own respective modules, which are then imported into `cf.py` for the sake of backwards compatibility.  Looking at this, I think it may make sense to import them into `__init__.py` instead, but that would break backwards compatibility with some plugins.  It'd be "nicer" to be able to do `from compliance_checker.cf import CF1_8Check` instead of `from compliance_checker.cf.cf import CF1_8Check`.  Perhaps if we do decide to move to `__init__.py`, we can have a deprecation warning for the import from `cf.py`.

PR #893, opened 2021-12-06
Fix undefined function in scale_factor/add_offset, typo

benjwadams:
Probable fix to #842

cc @mhidas

### DIFF
--- a/compliance_checker/cf/cf.py
+++ b/compliance_checker/cf/cf.py
@@ -1414,10 +1414,11 @@ class CF1_6Check(CFNCCheck):
         Reusable function for checking both add_offset and scale_factor.
         """
 
-        msg = (
+        msgs = []
+        error_msg = (
             f"Variable {variable.name} and {attr_name} must be equivalent "
-            "data types or {variable.name} must be of type byte, short, or int "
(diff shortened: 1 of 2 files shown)


compliance_checker/cf/cf.py
```diff
@@ -1455,7 +1457,7 @@ def check_add_offset_scale_factor_type(self, ds):
             results.extend(
                 list(
                     map(
-                        lambda x: self._check_scale_factor_add_offset(
+                        lambda x: self._check_add_offset_scale_factor(
```
mhidas, line 1460: This one doesn't exist either. Do you mean `_check_add_offset_scale_factor_type`? (the function defined directly above this one?

benjwadams:
Tests were non-existent for `check_add_offset_scale_factor_type` (note lack of leading underscore), so tests were switched to call that rather than the internal function, as the latter is called from the former anyhow.

compliance_checker/cf/cf.py
benjwadams: Erroneous function name has been addressed.

greenteadigital (approved):

PR #897, opened 2022-01-05
Data types for CF 1.9

benjwadams:
Adds new data types considered valid by CF 1.9 while refactoring old code.

### DIFF
--- a/compliance_checker/tests/test_cf.py
+++ b/compliance_checker/tests/test_cf.py
@@ -2303,6 +2303,16 @@ class TestCF1_9(BaseTestCase):
     def setUp(self):
         self.cf = CF1_9Check()
 
+    def test_check_data_types(self):
+        dataset = MockTimeSeries()
+        dataset.createVariable("ubyte", "u1", ())
+        dataset.createVariable("ushort", "u2", ())
+        dataset.createVariable("uint", "u4", ())
+        dataset.createVariable("int64", "i8", ())
+        dataset.createVariable("uint64", "u8", ())
+        result = self.cf.check_data_types(dataset)
+        assert result.value[0] == result.value[1]
+
     def test_domain(self):
         dataset = MockTimeSeries()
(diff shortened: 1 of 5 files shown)


compliance_checker/tests/test_cf.py
```diff
@@ -2449,6 +2452,59 @@ def test_taxonomy_skip_lsid(self):
         assert len(results[0].msgs) == 0
         assert results[0].value[0] == results[0].value[1]
 
+        
+class TestCF1_9(BaseTestCase):
+    def setUp(self):
+        self.cf = CF1_9Check()
+
+    def test_check_data_types(self):
+        dataset = MockTimeSeries()
```
benjwadams, line 2461: Ideally other data types should also be checked.  I wanted to keep DRY, but couldn't figure out how to inherit the tests from CF 1.6 data type checks without inheriting from the test case for 1.6.  Any suggestions would be welcome.

compliance_checker/tests/test_cf.py
```diff
@@ -2449,6 +2452,59 @@ def test_taxonomy_skip_lsid(self):
...
+        dataset.createVariable("ubyte", "u1", ())
+        dataset.createVariable("ushort", "u2", ())
+        dataset.createVariable("uint", "u4", ())
+        dataset.createVariable("int64", "i8", ())
+        dataset.createVariable("uint64", "u8", ())
+        result = self.cf.check_data_types(dataset)
+        assert result.value[0] == result.value[1]
```
benjwadams, line 2468: TODO: Add failing case.  Since all the simple types supported by netCDF4 module should be handled, I think we'd have to create a compound type.  See https://unidata.github.io/netcdf4-python/#Dataset.createVariable for more details.

benjwadams:
Integration tests failing due to a resource that is temporarily unavailable.

Pre-commit spell checks indicate that the surname "Claus" should be "clause" instead.  This is wrong as it's part of a citation from a taxonomy example.

PR #899, opened 2022-01-07
Refactor Python package setup into GHA composite action

benjwadams:


benjwadams:
@ocefpaf, here you see some of my initial attempts to refactor the GitHub Actions code for greater code reuse when installing the core Python packages in use by compliance-checker.   Unfortunately, something's not quite right, as most of the GHA checks have disappeared... :ghost:

ocefpaf:
> most of the GHA checks have disappeared...

Usually, when that happens, it is b/c the yaml is not OK. I don't know any good yaml linter for GHA but I found this one: https://github.com/marketplace/actions/yaml-checker (Warning! I never used that one and must be added as a GHA, which is annoying.)

Issue #898, opened 2022-01-06
Undefined Variables


greenteadigital:
cf.py:4102
"boundary_variable" is not defined
![MicrosoftTeams-image (4)](https://user-images.githubusercontent.com/1993231/148292818-d10d71d5-a696-471d-a580-9f98a5d88e02.png)

ioos.py:959
"good_result" is not defined
![MicrosoftTeams-image (2)](https://user-images.githubusercontent.com/1993231/148292525-97794e5e-6c83-4240-b6e4-dbf115c7894a.png)

I don't see the point in `git blame`ing anyone, but can these get defined or removed, pls?

_Originally posted by @greenteadigital in https://github.com/ioos/compliance-checker/pull/892#pullrequestreview-845066683_

benjwadams:
Some of these code paths don't have proper test coverage.  Although linter should have caught them, it's worthwhile to check code coverage too on a number of these, so considered #641 as related here.

benjwadams:
Related: #900

ocefpaf:
BTW, flake8 tests are commented out and those should catch these errors. I don't recall why but probably b/c most of the offending errors were ignored in the past. We should re-enable those and skip what we don't want to use.

ocefpaf:
See #901 for the flake8 results. I can fix most of those but I'd like to check with @benjwadams first and agree on what checks we want to keep before causing a lot of churn in the code.

benjwadams:
Fixed by #903, closing.

PR #903, opened 2022-01-17
Fix undefined variables/tests, linter fixes

benjwadams:
Fixes a number of undefined variables and add some tests where coverage is not present.  Also address various linter messages.
Addresses #898.


benjwadams:
@greenteadigital, it looks like integration test is failing due to some URL changing and returning an error response rather than the expected sensor response.

PR #905, opened 2022-01-25
Remove Python 2.7 warning / contributors image added in README

benjwadams:
Python 2.7 has stopped being supported long time ago, so a notice that it is no longer supported is fairly superfluous.   PyPI/setup.py should indicate supported versions Python versions in any event.

Many other people have made contributions to compliance checker, so it's recommended that this section be removed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,6 @@ for completeness and community standard compliance of local or remote
 file standards. The python module can be used as a command-line tool or as a
 library that can be integrated into other software.
 
-## Python 2.7 Deprecation Notice
-As of release 4.3.0, the __Compliance Checker no longer officially supports Python 2.7__.
-
 A [web-based version](https://data.ioos.us/compliance/index.html) of the Compliance
 Checker was developed to enable a broader audience and improve accessibility for the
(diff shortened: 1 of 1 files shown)


benjwadams:
I'm open to further discussion, especially regarding the removal of contributors section and possible replacements therein.
 I'm not against acknowledging contributors, but folks like @ocefpaf who have also contributed a lot to the project aren't mentioned.   If we do wish to keep contributor acknowledgements on the README/front matter of the GitHub page, we could do so in a more modern fashion by using something like a contribution badge I've seen on some other GitHub projects.  Some of the email information removed in this PR no longer reflected current contact information for these contributors, and while I do want to still recognize the many contributions they made to this project, I feel that using their GitHub identities would be better in the long run.

ocefpaf:
I don't have a strong opinion on the contribution part. It is stale though and a modern one would be nice. All the other changes are very welcome! Thanks for doing this @benjwadams!

benjwadams:
Added a dynamic image badge for the contributors section instead of completely removing.  Seems to be a good compromise for all, provided it works as I can't preview it at this moment.

PR #911, opened 2022-03-15
Fix integration tests

benjwadams:
Fixes (hopefully) integration tests, improves clarity of some CF related messages, and removes a redundant message.


benjwadams:
Hang on, having some additional difficulties with the units handling.

Issue #904, opened 2022-01-24
Integration test protocol failures


benjwadams:
**Version of compliance checker running:**

`develop` branch

**Describe the checker this affects:**
_The ACDD, IOOS, and CF checkers are currently considered in scope for the purposes of the issue tracking on this repository. Issues with other plugins should be placed into their respective repositories._

N/A / integration tests.

**Attach a minimal CDL or NetCDF file which is able to reproduce the issue**

N/A , URLs

**To Reproduce:**
_List the steps or command line to run to reproduce the issue_

Run `py.test compliance_checker/tests/test_integration.py`

**Describe the issue below:**
_Any additional detail or information relevant to the issue_

Integration tests are currently failing due to HTTP 403 or 404 errors.

```
compliance_checker/tests/test_protocols.py::TestProtocols::test_erddap PASSED
compliance_checker/tests/test_protocols.py::TestProtocols::test_hyrax FAILED
compliance_checker/tests/test_protocols.py::TestProtocols::test_netcdf_content_type PASSED
compliance_checker/tests/test_protocols.py::TestProtocols::test_sos FAILED
compliance_checker/tests/test_protocols.py::TestProtocols::test_thredds PASSED
```

See GitHub Actions log for extended test details https://github.com/ioos/compliance-checker/runs/4900074527?check_suite_focus=true

In summation, there seem to be two issues:

1) ingria.coas.oregonstate.edu appears unreachable.   

2) ncSOS appears deactivated on https://data.oceansmap.com/thredds/sos/caricoos_ag/VIA/VIA.ncml

Action items are:

1) Contact OSU/OOI about the status of their subdomain.

2)  Contact OceansMap team at RPS regarding status of ncSOS services on data.oceansmap.com/thredds

benjwadams:
Closed by #912

PR #912, opened 2022-03-21
Fix protocol tests that had broken URLs

benjwadams:
Fixes protocol test failures mentioned in #904.


benjwadams:
Note: this is only for `test_protocols.py`.   The other outstanding test failures will be addressed in separate pull requests.

PR #916, opened 2022-04-08
Adds checks for .nc filename extensions

benjwadams:


compliance_checker/cf/cf_1_6.py
```diff
@@ -49,6 +49,13 @@ def __init__(self, options=None):  # initialize with parent methods and data
     ###############################################################################
     # Chapter 2: NetCDF Files and Components
     ###############################################################################
+    def check_filename(self, ds):
+        """Checks that the filename ends with .nc"""
+        # IMPLEMENTS CONFORMANCE 2.1
+        filename_suffix = TestCtx(BaseCheck.HIGH, self.section_titles["2.1"])
+        filename_suffix.assert_true(ds.filepath().endswith("nc"),
```
ocefpaf, line 56: Probably a corner case but a file named `filenc` would pass this check but should not, right? Maybe we should use pathlib:

```
In [1]: Path("fname.nc").suffix
Out[1]: '.nc'

In [2]: Path("fnamenc").suffix
Out[2]: ''
```

PR #918, opened 2022-04-12
Handle nonstandard calendar and leap time in time variables

benjwadams:
Handles nonstandard calendars which need to define leap time definitions.

PR #920, opened 2022-04-13
Add checks/tests for external variables

benjwadams:


benjwadams:
Test failures were addressed in #918

compliance_checker/cf/cf_1_7.py
```diff
@@ -75,8 +75,47 @@ def __init__(self, options=None):
...
+                external_vars_ctx.out_of += 1
+                bad_msg = ("Global attribute external_variables should not "
+                           "have any variable names which are present in the dataset. "
+                           "Currently, the following names appear in both external_variables "
+                           f"and the dataset's variables: {bad_external_var_names}")
+                external_vars_ctx.messages.append(bad_msg)
+
+        # string/global attributes are handled in Appendix A checks
+        except (AttributeError, ValueError):
+            pass
```
benjwadams, line 109: NB: currently one has to check the Appendix A attributes in CF checker to get an error on the type:

```
python cchecker.py -O cf:enable_appendix_a_checks -t cf /tmp/foo2.nc 
Running Compliance Checker on the datasets from: ['/tmp/foo2.nc']


                                Version unknown                                 
                     Report generated 2022-04-13T18:45:57Z                      
                                     cf:1.8                                     
http://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html
                               Corrective Actions                               
foo2.nc has 2 potential issues


                                     Errors                                     
§2.6 Attributes
* external_variables must be a string

§3.3 Standard Name
* Attribute long_name or/and standard_name is highly recommended for variable ext3
```

Also note that the section does not correspond to 2.6.3.   Historical reasons for this were related to backwards compatibility, which I disagreed with, so an option was supplied for this behavior.  In any event, there is some redundancy in CF with respect to Appendix A specifying the types for global attributes as well as some checks calling for attribute types.  It'd be best to check this by default, but have it only checked once and have it grouped with the corresponding section.

PR #924, opened 2022-04-29
Flag fixes

benjwadams:
Fixes to flags:

- Checks that 0 does not appear in `flag_masks`
- Fixes `flag_values` checking when attribute is single-valued/scalar (fixes #992)


benjwadams:
@pp-mo, I can't add you as a reviewer presently, but please see if this fixes issues you were having with #992.

pp-mo:
> see if this fixes issues you were having with #992.

Thanks, yes it does. 
Closed #924 in anticipation.

PR #925, opened 2022-05-18
Standard name recommendations

benjwadams:
Suggests nearby corrections for incorrect standard names.  Implements #806.

### DIFF
--- a/compliance_checker/cf/cf_1_6.py
+++ b/compliance_checker/cf/cf_1_6.py
@@ -22,6 +22,8 @@ from compliance_checker.cf.appendix_f import (grid_mapping_attr_types16,
                                               grid_mapping_dict16)
 
 from compliance_checker.cf.cf_base import CFNCCheck, appendix_a_base
+import difflib
+
 logger = logging.getLogger(__name__)
 
 class CF1_6Check(CFNCCheck):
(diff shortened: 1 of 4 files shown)


compliance_checker/cf/cf_1_6.py
```diff
@@ -936,13 +938,23 @@ def check_standard_name(self, ds):
...
+                        .format(
+                            standard_name or "undefined",
+                            self._std_names._version))
+                    close_matches = difflib.get_close_matches(standard_name,
+                                                              self._std_names)
+                    if close_matches:
+                        if len(close_matches) == 1:
+                            err_msg += f" Possible close match: {close_matches}"
+                        else:
+                            err_msg += f" Possible close matches: {close_matches}"
```
ocefpaf, line 955: Why not `f" Possible close match(es):`? I know, not as elegant to the user but a few less complexity in the code.

ocefpaf (approved):

compliance_checker/cf/cf_1_6.py
benjwadams: Was considering that, I'll revise.

ocefpaf: No strong opinion on my side BTW. My user brain wants the original commit, my programmer brain wants the change.

PR #931, opened 2022-05-24
mark ERDDAP test with vcr

ocefpaf:
@benjwadams this should help you with the ERDDAP server restrictions. The network interaction is recorded in the cassettes and then re-played for the test, reducing the server load and speeding up the tests.

Note that I only recorded the ERDDAP tests but, if you think it is OK, we can record all network interactions in this module.

BTW, I took this opportunity to make this test module a bit more pytest-y by removing the unittest class


ocefpaf:
The failing tests are:

```
compliance_checker/tests/test_protocols.py::test_hyrax FAILED
compliance_checker/tests/test_protocols.py::test_sos FAILED
```

both tests are using URLs that are down at the moment, http://ingria.coas.oregonstate.edu/opendap/hyrax/aggregated/ocean_time_aggregation.ncml and https://data.oceansmap.com/thredds/sos/caricoos_ag/VIA/VIA.ncml.

The odd thing about the SOS test is that it seems that they removed that endpoint but did not update the page accordingly.

SOS

benjwadams:
Try rebasing on master -- develop should have a merge commit against current master as well.  There were a number of fixes made to CI and integration tests in master that should fix some of these failures.

ocefpaf:
> Try rebasing on master -- develop should have a merge commit against current master as well. There were a number of fixes made to CI and integration tests in master that should fix some of these failures.

Done!

PR #932, opened 2022-05-24
update GHAs, clean up yml, use micromamba

ocefpaf:
In this PR:

1. update the GitHub Actions versions
2. move from conda to micromamba to get faster builds
3. clean up the plugin tests a bit (we created a matrix when we did not really needed it)
4. run the tests on `master`, `main` and `develop` (we should probably get rid of `master` BTW)

This is not ready for review yet. I'll work on caching the environment first to make it easier to re-run tests in a PR.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -10,30 +10,29 @@ jobs:
     runs-on: ubuntu-latest
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
 
-    - name: Setup Conda
-      uses: s-weigand/setup-conda@v1
+    - name: Setup Micromamba
+      uses: mamba-org/provision-with-micromamba@main
       with:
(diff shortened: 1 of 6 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -16,8 +16,6 @@ jobs:
       uses: mamba-org/provision-with-micromamba@main
       with:
         environment-file: false
-        cache-env: true
-        environment-name: TEST
```
ocefpaf: We could use downloads-cache instead but according to their docs it can be even slower than direct download. See https://github.com/mamba-org/provision-with-micromamba/pull/38#discussion_r808837618

benjwadams:
Closing due to merge of #980

ocefpaf:
Done in #980

ocefpaf:
> Closing due to merge of #980

We probably clicked almost at the same time :-)

PR #944, opened 2022-09-08
Removes deprecated NumPy types

benjwadams:


compliance_checker/cf/cf_1_6.py
```diff
@@ -156,7 +156,7 @@ def _check_add_offset_scale_factor_type(self, variable, attr_name):
             ) or (  # will short-circuit or if first condition is true
                 isinstance(att, (np.float32, np.float64, float))
                 and variable.dtype
-                in (np.byte, np.short, np.int16, np.int, np.int32, int)
+                in (np.byte, np.short, np.int16, np.int32, np.int64, int)
```
ocefpaf, line 159: I think this can be expressed as `is np.issubdtype(variable.dtype, np.integer)`.

benjwadams: `int64` shouldn't actually be here as this is CF 1.6.

benjwadams: Also, `issubdtype` on types such as `np.uint8` will return true, while we don't actually want this for CF 1.6.

ocefpaf: Indeed. I was approaching this from 'catch all ints' point of view, not CF.

benjwadams:
Please squash when ready to merge.

leilabbb (approved):

Issue #960, opened 2022-10-03
Improve CF Conformance: 2.2 Data Types


jamesdoyle21:
2.2 Data Types
Requirements:

 

- [x] CF attributes that take string values must be 1D character arrays or single atomic strings.
~~
Unclear, but I think this is what they are looking for: 
[also mentioned below under section 6.1 Labels]
- If the variable is of type string it must have at most one dimension, which must match one of those of the data variable. 
- If the variable is of type char it must have one or two dimensions, where the trailing (CDL order) or sole dimension is for the maximum string length. If there are two dimensions, the leading dimension (CDL order) must match one of those of the data variable.
The following has been verified:

 

- [x] The netCDF data types char, byte, short, int, float or real, and double are all acceptable. NetCDF4 allows string as data type, which is also acceptable
- [x] ~~

benjwadams:
Should be done for attributes only, not variables.

benjwadams:
It looks like netCDF3/classic, which doesn't have advanced string types is cast to `str` when interacting with a file.  This means that this section:
> CF attributes that take string values must be 1D character arrays or single atomic strings.
will pass by default.

Thus, I'm closing this issue.

PR #961, opened 2022-10-04
Only allow fixed length string variables to be one or two-dimensional

benjwadams:
Implements missing point of #960 1/2 .

### DIFF
--- a/compliance_checker/tests/test_cf.py
+++ b/compliance_checker/tests/test_cf.py
@@ -153,6 +153,35 @@ class TestCF1_6(BaseTestCase):
         dataset = self.load_dataset(STATIC_FILES["bad_data_type"])
         result = self.cf.check_data_types(dataset)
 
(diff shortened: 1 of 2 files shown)


compliance_checker/tests/test_cf.py
```diff
@@ -153,6 +153,35 @@ def test_check_data_types(self):
...
+        dataset.createDimension("extra", 9)
+
+        # TEST CONFORMANCE 2.2 1/2
+        # create strings with no dimensions -- fail
+        dataset.createVariable("zero_dimensional", "S1", ())
+        # create strings with one dimension -- valid
+        dataset.createVariable("one_dimensional", "S1", ("max_string_length"))
+        # create strings with two dimensions - valid
+        # NB: string length is supposed to be last, but this can't be inferred
+        #     easily with
```
benjwadams, line 169: ... the code we have presently

Issue #953, opened 2022-10-03
Improve CF Conformance: 7.2 Cell Measures


jamesdoyle21:
7.2 Cell Measures
Requirements:

 

- [x] The type of the cell_measures attribute is a string whose value is list of blank separated word pairs in the form measure: var. The valid values for measure are area or volume. The var token specifies a variable that must either exist in the file or be named by the external_variables attribute. The dimensions of the variable specified by var must be the same as, or be a subset of, the dimensions of the variable to which they are related.

- [x]  A measure variable must have units that are consistent with the measure type, i.e., square meters for area measures and cubic meters for volume measures.

PR #966, opened 2022-10-19
updated check_cell_measures

leilabbb:
This is in reference to [issue #953](https://github.com/ioos/compliance-checker/issues/953)

Updated the function "check_cell_measures" to include:
- [ ] A measure variable must have units that are consistent with the measure type, i.e., square meters for area measures and cubic meters for volume measures.

### DIFF
--- a/compliance_checker/cf/cf_1_6.py
+++ b/compliance_checker/cf/cf_1_6.py
@@ -2663,20 +2663,20 @@ class CF1_6Check(CFNCCheck):
                         )
                     else:
                         # IMPLEMENTATION CONFORMANCE REQUIRED 2/2
-                        # verify this combination {area: 'm2', volume: 'm3'} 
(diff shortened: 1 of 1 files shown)


benjwadams:
pre-commit.ci autofix

compliance_checker/cf/cf_1_6.py
```diff
@@ -2661,6 +2661,22 @@ def check_cell_measures(self, ds):
                                 cell_meas_var_name
                             )
                         )
+                    else:
+                        # IMPLEMENTATION CONFORMANCE REQUIRED 2/2
+                        # verify this combination {area: 'm2', volume: 'm3'}
+                        dic_expected = {"area": "m2", "volume": "m3"}
```
benjwadams, line 2667: This should be able to take an arbitrary measure value, not just meters.

Issue #986, opened 2023-01-10
CF: exception '`cftime._cftime.DatetimeGregorian` object has no attribute `set_fill_value`


mhidas:
**Version of compliance checker running:** current master (e.g. at 799bf15318)

**Describe the checker this affects:** CF-1.6

**Attach a minimal CDL or NetCDF file which is able to reproduce the issue**
https://s3-ap-southeast-2.amazonaws.com/imos-data/IMOS/SOOP/SOOP-XBT/DELAYED/Line_IX28_Dumont-d-Urville-Hobart/2019/IMOS_SOOP-XBT_T_20190108T073100Z_IX28_FV01_ID-89011149.nc

**To Reproduce:**
`compliance-checker -t cf:1.6 IMOS_SOOP-XBT_T_20190108T073100Z_IX28_FV01_ID-89011149.nc`

**Describe the issue below:**
The file passes the checks, but an exception has been caught:
```
WARNING: The following exceptions occurred during the cf:1.6 checker (possibly indicate compliance checker issues):
cf:1.6.check_calendar: 'cftime._cftime.DatetimeGregorian' object has no attribute 'set_fill_value'
  File "/sw/chef/src/compliance-checker/compliance_checker/cf/cf_1_6.py", line 1949, in check_calendar
    ret_val.append(check_standard_calendar_no_cross(time_var))
  File "/sw/chef/src/compliance-checker/compliance_checker/cf/cf_1_6.py", line 1912, in check_standard_calendar_no_cross
    times.set_fill_value(crossover_date)
```

mhidas:
In case it's helpful, I'm running the checker in a Python 3.8.13 virtualenv with `cftime` version 1.6.2
([full pip freeze output](https://github.com/ioos/compliance-checker/files/10378806/pip_freeze.csv))

benjwadams:
Strange, I upgraded `cftime` to 1.6.2 and did not encounter this error.   Possibly it's some change in numpy-ish API, I'll keep checking.

benjwadams:
Fixed by #990

PR #990, opened 2023-01-27
Fix calendar crossing 1582 issue on some builds

benjwadams:
@mhidas, should address issue mentioned in #986 by using an alternative implementation for excluding _FillValue time data.


leilabbb (commented):
no comments

PR #1003, opened 2023-04-04
[CF 1.9] Ensure time coordinate variables have a calendar attribute s…

benjwadams:
…pecified

### DIFF
--- a/compliance_checker/cf/cf_1_9.py
+++ b/compliance_checker/cf/cf_1_9.py
@@ -1,5 +1,6 @@
 from netCDF4 import Dataset
 
+from compliance_checker import cfutil
 from compliance_checker.base import BaseCheck, TestCtx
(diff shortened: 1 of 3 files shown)


compliance_checker/cf/cf_1_9.py
leilabbb, line 1: The test does not check for the value of the coordinate attribute, it only checks if a time coordinate variable has a calendar attribute.
The test does not check if the time variable is the only variable with a coordinate attribute.

benjwadams: This only checks for a time coordinate variable and see if it has a `calendar` attribute. The remaining checks inherited from should check the value of calendar attribute, etc.

benjwadams: Namely:
> A time coordinate variable should have a calendar attribute

leilabbb (approved):

PR #1005, opened 2023-04-05
Fixes dead ERDDAP link in integration tests

benjwadams:


benjwadams:
Only remaining integration test failure is AOOS THREDDS, which has a 502 error reporting as of the time of this post.  I imagine it will come back up, so I'm not going to attempt to move to a different source for that.

PR #1007, opened 2023-04-12
Build api docs

ocefpaf:
@benjwadams and @leilabbb this is not ready for merging yet but it is a start. Here we have:

- basic sphinx + pydata theme configuration
- autodoc for the API
- GitHub Actions to publish the docs as a gh-pages

I think that the preview will work only if we do step https://github.com/rossjrw/pr-preview-action#grant-actions-permission-to-read-and-write-to-the-repository (I do not have enough permission to enable it.)


ocefpaf:
@mwengren, @benjwadams, and @leilabbb I consolidated the README as a "quick intro" and front page, added the autogenerated API docs, and the compliance-checker-web FAQ here. We need better sections, intro, and something to tie this all together but before we do that we can discuss this PR and what/how it does.

PR #1009, opened 2023-04-14
Fix numpy deprecations that are errors in 1.24.2

ocefpaf:
I also run pre-commit on the files I touched so we slowly get them all back into compliance.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,17 +34,17 @@ repos:
     - id: seed-isort-config
 
 - repo: https://github.com/psf/black
-  rev: 22.12.0
+  rev: 23.3.0
   hooks:
   - id: black
     language_version: python3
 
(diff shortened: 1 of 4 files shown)


ocefpaf:
The integration tests have been broken for quite a while so I won't touch those now. @benjwadams this one is ready for review and merge. It will unblock #1008.

ocefpaf:
@benjwadams please let me know if you have cycles to review this one. It is not a bug change but it is blocking other PRs. The pre-commit failures are already addressed in #1010 and I want to tackle the integration test after this one is merged.

benjwadams (approved):

ocefpaf:
> SOS datasets.

We should check with folks but my feeling is that we could probably drop the SOS one and mark then as no longer supported. I need to check what they are testing though to be sure.

PR #1016, opened 2023-04-19
Remove unused helper functions for axes

benjwadams:


benjwadams:
@ocefpaf, there are probably still more unused code segments especially in some of the `*util.py` segments.  Will keep looking and pruning as necessary.

ocefpaf:
Ideally we should remove and release quickly.  That way, if a user downstream is still using dead code, they would let us know ASAP.

benjwadams:
I just checked the "mainline" IOOS-related Compliance Checker plugins and I don't see that they'll be affected.  Merging.

PR #1024, opened 2023-05-17
move to pyproject.toml

ocefpaf:
~~Work in progress and not ready for review.~~

This is the new standard for Python packaging but there are some details missing to accommodate everything we do in compliance-checker, like shipping scripts is deprecated won't be support for much longer.


pyproject.toml
```diff
@@ -6,11 +6,90 @@ requires = [
...
+[project]
+name = "compliance-checker"
+description = "Checks Datasets and SOS endpoints for standards compliance"
+readme = "README.md"
+license = {text = "Apache-2.0"}
+maintainers = [
+    {name = "Dave Foster", email = "dave@axiomdatascience.com"},
+    {name = "Benjamin Adams"},
+    {name = "Luke Campbell"},
+    {name = "Filipe Fernandes"},
```
ocefpaf, line 18: @benjwadams are you OK with this change? Most projects are moving away from "author" in liue for maintainers.

pyproject.toml
```diff
@@ -6,11 +6,90 @@ requires = [
...
+dynamic = [
+  "dependencies",
+  "version",
+]
+[project.urls]
+documentation = "https://ioos.github.io/compliance-checker"
+homepage = "https://compliance.ioos.us/index.html"
+repository = "https://github.com/ioos/compliance-checker"
+[project.scripts]
+compliance-checker = "cchecker:main"
```
ocefpaf, line 46: This is deprecated and in the future. We need a better solution to ship that script.

pyproject.toml
```diff
@@ -6,11 +6,90 @@ requires = [
...
+"ioos-1.1" = "compliance_checker.ioos:IOOS1_1Check"
+"ioos-1.2" = "compliance_checker.ioos:IOOS1_2Check"
+"ioos_sos" = "compliance_checker.ioos:IOOSBaseSOSCheck"
+
+[tool.setuptools]
+packages = ["compliance_checker"]
+license-files = ["LICENSE"]
+zip-safe = false
+include-package-data = true
+script-files = ["cchecker.py"]
```
ocefpaf, line 63: See https://github.com/ioos/compliance-checker/pull/1024/files#r1209407884

ocefpaf:
Publishing to PyPI using Python 3.12 failed with:

```
ValueError: Require to set UDUNITS2_XML_PATH for a cf-units wheel build.
```

That is likely a problem with the cf_units wheel. I'll revert to use 3.11 b/c compliance-checker is pure Python but there is a change that folks using Python 3.12 will hit that issue.

PR #1026, opened 2023-05-17
Enable bugbear...

ocefpaf:
... and fix almost all possible bugs found.

Note that we are explicitly skipping two tests:

```toml
"compliance_checker/__init__.py" = ["B019"]
"compliance_checker/cfutil.py" = ["B028"]
```

The former should be upstream'ed and fixed there. The latter is actually a false positive b/c the stacklevel is there.

### DIFF
--- a/compliance_checker/acdd.py
+++ b/compliance_checker/acdd.py
@@ -188,7 +188,7 @@ class ACDDBaseCheck(BaseCheck):
             msgs = []
             # Check units and dims for variable
             unit_check = hasattr(ds.variables[variable], "units")
-            no_dim_check = getattr(ds.variables[variable], "dimensions") == ()
+            no_dim_check = ds.variables[variable].dimensions == ()
             # Check if we have no dimensions.  If no dims, skip test
             if no_dim_check:
                 continue
@@ -256,7 +256,7 @@ class ACDDBaseCheck(BaseCheck):
 
         # identify lat var(s) as per CF 4.1
         lat_vars = {}  # var -> number of criteria passed
-        for name, var in ds.variables.items():
+        for _name, var in ds.variables.items():
             # must have units
             if not hasattr(var, "units"):
                 continue
@@ -354,7 +354,7 @@ class ACDDBaseCheck(BaseCheck):
 
         # identify lon var(s) as per CF 4.2
         lon_vars = {}  # var -> number of criteria passed
-        for name, var in ds.variables.items():
+        for _name, var in ds.variables.items():
             # must have units
             if not hasattr(var, "units"):
                 continue
@@ -779,7 +779,7 @@ class ACDD1_3Check(ACDDNCCheck):
         if not hasattr(ds, "metadata_link"):
             return
         msgs = []
-        meta_link = getattr(ds, "metadata_link")
+        meta_link = ds.metadata_link
         if "http" not in meta_link:
             msgs.append("Metadata URL should include http:// or https://")
         valid_link = len(msgs) == 0
@@ -793,7 +793,7 @@ class ACDD1_3Check(ACDDNCCheck):
         """
         if not hasattr(ds, "id"):
             return
-        if " " in getattr(ds, "id"):
+        if " " in ds.id:
             return Result(
                 BaseCheck.MEDIUM,
                 False,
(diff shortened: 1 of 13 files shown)


compliance_checker/acdd.py
```diff
@@ -256,7 +256,7 @@ def check_lat_extents(self, ds):
 
         # identify lat var(s) as per CF 4.1
         lat_vars = {}  # var -> number of criteria passed
-        for name, var in ds.variables.items():
+        for _name, var in ds.variables.items():
```
benjwadams, line 259: Surprised it didn't change to just `.values()` here instead.

compliance_checker/base.py
```diff
@@ -549,7 +549,7 @@ def attr_check(kvp, ds, priority, ret_val, gname=None, var_name=None):
     # starting with "check".  Avoid naming check functions
     # starting with check if you want to pass them in with
     # a tuple to avoid them being checked more than once
-    elif hasattr(other, "__call__"):
+    elif callable(other):
```
benjwadams, line 552: I wrote this, didn't even realize that this was preferred style...

compliance_checker/cf/cf_1_7.py
```diff
@@ -794,7 +794,8 @@ def _evaluate_towgs84(self, val):
             return (True, msg)
 
     def check_grid_mapping(self, ds):
-        super().check_grid_mapping.__doc__
+        # FIXME: Looks like this is not needed.
+        # super().check_grid_mapping.__doc__
```
benjwadams, line 798: Docstring is taken from `super` here or?

I guess Python 3.5 introduced docstring inheritance for subclass methods?  Probably wrote this back when we were supporting Python 2.7.

compliance_checker/acdd.py
ocefpaf: It probably doesn't have the notion of the tuple output just checks for unused vars. But we can change that to values.

compliance_checker/base.py
ocefpaf: IMO what you had there is "better style" but it can fail in some corner cases. The callable check won't.

compliance_checker/cf/cf_1_7.py
ocefpaf: To be honest I'm not sure what to do here.

benjwadams (approved):

compliance_checker/cf/cf_1_7.py
ocefpaf: This is the behaviour of the docs without this line:

```python
Signature: CF1_7Check.check_grid_mapping(self, ds)
Docstring:
5.6 When the coordinate variables for a horizontal grid are not
longitude and latitude, it is required that the true latitude and
longitude coordinates be supplied via the coordinates attribute. If in
addition it is desired to describe the mapping between the given
coordinate variables and the true latitude and longitude coordinates,
the attribute grid_mapping may be used to supply this description.

This attribute is attached to data variables so that variables with
different mappings may be present in a single file. The attribute takes
a string value which is the name of another variable in the file that
provides the description of the mapping via a collection of attached
attributes. This variable is called a grid mapping variable and is of
arbitrary type since it contains no data. Its purpose is to act as a
container for the attributes that define the mapping.

The one attribute that all grid mapping variables must have is
grid_mapping_name which takes a string value that contains the mapping's
name. The other attributes that define a specific mapping depend on the
value of grid_mapping_name. The valid values of grid_mapping_name along
with the attributes that provide specific map parameter values are
described in Appendix F, Grid Mappings.

When the coordinate variables for a horizontal grid are longitude and
latitude, a grid mapping variable with grid_mapping_name of
latitude_longitude may be used to specify the ellipsoid and prime
meridian.


In order to make use of a grid mapping to directly calculate latitude
and longitude values it is necessary to associate the coordinate
variables with the independent variables of the mapping. This is done by
assigning a standard_name to the coordinate variable. The appropriate
values of the standard_name depend on the grid mapping and are given in
Appendix F, Grid Mappings.

:rtype: list
:return: List of results
Source:   
    def check_grid_mapping(self, ds):
        # FIXME: Looks like this is not needed.
        # super().check_grid_mapping.__doc__
        prev_return = super().check_grid_mapping(ds)
        grid_mapping_variables = cfutil.get_grid_mapping_variables(ds)
        for var_name in sorted(grid_mapping_variables):
:
```

Is that OK?

PR #1027, opened 2023-05-17
Fix a broken link and drop a removed one.

ocefpaf:
This should fix the docs build.


compliance_checker/cf/cf_1_6.py
```diff
@@ -26,8 +26,7 @@ class CF1_6Check(CFNCCheck):
     """CF-1.6-specific implementation of CFBaseCheck; supports checking
     netCDF datasets.
     These checks are translated documents:
```
ocefpaf, line 28: The Links below are broken. I replaced one with the CF docs but the other one seems to be gone.

Screenshot from 2023-05-17 17-00-10

PR #1029, opened 2023-05-18
remove MemoizedDataset

ocefpaf:
The improvements only show up when calling `get_variables_by_attributes` with the same `kwargs` multiple times and I'm not sure if that is a big improvement for users. (Maybe for cc tests it is? Not sure.)

Regardless, the way this is implement it can lead to [memory leaks](https://bugs.python.org/issue19859) and this change should be upstream anyway.  See an attempt to do that in https://github.com/Unidata/netcdf4-python/pull/1253

Another reason to remove this is to make `netcdf4` optional and allowing us to swap it with `h5netcdf` when building a WASM version of CC.


benjwadams:
Relevant discussion to this issue is located in this now closed issue: https://github.com/ioos/compliance-checker/issues/70

ocefpaf:
> So I have mixed feelings about this.

We don't have to merge this now. Our options are:

1. Wait for this feature to be upstreamed, merge, and pin the latest netcdf4-python that has this cached;
2. Refactor to use a simple function call that we can safely cache.

Option 1, IMO, is the better one but we don't control the timeline. Option 2 can be more portable b/c I plan to introduce the `h5netcdf` as an alternative and I'm not sure how `get_variables_by_attributes` work in that library.

Changing this to draft anyway as we ponder on these options.

benjwadams:
> Wait for this feature to be upstreamed, merge, and pin the latest netcdf4-python that has this cached;

It might be considered a niche feature and only really works as long as there are simple cases where the cache isn't invalidated, such as opening a NetCDF file as read-only.

ocefpaf:
> It might be considered a niche feature

Not sure I agreed with that. It is definitely worth upstreaming.

> and only really works as long as there are simple cases where the cache isn't invalidated, such as opening a NetCDF file as read-only.

That is the tricky part and probably why my PR is failing upstream. Do you want to take a look at it?

https://github.com/Unidata/netcdf4-python/pull/1253

ocefpaf:
> Not sure I agreed with that. It is definitely worth upstreaming.

This is upstream now: https://github.com/Unidata/netcdf4-python/pull/1253

We can work on this PR when there is a new release of netCDF4-python.

PR #1042, opened 2023-09-05
let's skip this one for now

ocefpaf:
We should revisit that URL and check if it is back online in the near future. For it is safe to skip so we can get the docs to build again.


ocefpaf:
Self-merging this one b/c it is pretty harmless and will unlock new PRs. However, the integration tests are broken again :-/

Seems to be changes in datasets URLs upstream. I'll check them later.

PR #1039, opened 2023-08-01
Remove pkg resources

ocefpaf:
[pkg_resources is notoriously slow and problematic](https://github.com/pypa/setuptools/issues/926). This PR tries to remove it from compliance-checker.

PR #1056, opened 2024-03-19
Some pytest fixes

jcermauwedu:
* base.py: Sometimes cfutil is None
 * protocols/netcdf.py:
   - Return False right away if content_type is None
   - Support extended type strings: "application/x-netcdf;ver=4"
 * suite.py: If text for xml fails, look for netcdf one more time
 * Convert suite.run() calls to suite.run_all() calls
 * Update hyrax url
 * Update thredds url
 * Add some development notes for troubleshooting tests with pytest using the vcr feature.

For the SOS test in test_protocols.py, I set that to skip.  I searched for awhile for a working SOS service.

Looking forward to seeing https://github.com/ioos/compliance-checker/pull/1039 merged.

PR #1059, opened 2024-04-10
Remove pkg_resources and go fully to pyproject.toml

ChrisBarker-NOAA:
This PR removes the deprecated pkg_resources.

It also updates to a fully pyproject.toml based build / install, building off the work in: #1024 

Looking now, it seems that #1039 is in fact pretty much the same thing -- wish I'd noticed that earlier!

Sorry to put two things in one PR -- I had done one, and didn't realize how much. was involved in pkg_resources.

Anyway, it would be great if either this or #1039 could be merged.


ChrisBarker-NOAA:
Hmm -- test failing, due to bad yaml file:

[Invalid workflow file: .github/workflows/default-tests.yml#L24](https://github.com/ChrisBarker-NOAA/compliance-checker/actions/runs/8624698426/workflow)
The workflow is not valid. .github/workflows/default-tests.yml (Line: 24, Col: 22): The expression is not closed. An unescaped ${{ sequence was found, but the closing }} sequence was not found.

I didn't touch that file, so ???

jcermauwedu:
Using `importlib.metadata.entry_points(group=` forces the minimum python to 3.10.   I was struggling in python 3.9.   The workflow `default-tests.yml` may need to drop 3.8 and 3.9.   I will wait for this PR to continue with the pytest fixes in #1056.

ChrisBarker-NOAA:
Now that #1039 has been merged, I'm going to update this. closing for now.

Thanks for the 3.10 hint -- that's where I was testing, but it wasn't clear from the docs which features were added when to the stdlib. I'll use the backported version for older versions.

ChrisBarker-NOAA:
OK -- here it is: #1060

PR #1060, opened 2024-04-12
removed last of pkg_resources.

ChrisBarker-NOAA:
I've removed the last of pkg_resources, replacing it with:

`importlib.metadata`

for Python >= 3.9 

and the backport:

`import importlib_metadata`

for older versions.

I also added conditional import of: `importlib.resources`, so the backport will only be used if it's needed.

Both the backports are in the requirements files, as I know of no way to make a requirement python-version-specific.

NOTE: do we really need to keep older versions working? With this kind of utility, I'd think folks could user a newish Python.

### DIFF
--- a/compliance_checker/base.py
+++ b/compliance_checker/base.py
@@ -191,10 +191,9 @@ class BaseCheck:
         are cleared before the next checker uses it. Some caches were
         inadvertently mutated by other functions.
         """
-        # odd errors -- module getting deleted before this object?
-        if cfutil is not None:
-            cfutil.get_geophysical_variables.cache_clear()
-            cfutil.get_time_variables.cache_clear()
+
+        cfutil.get_geophysical_variables.cache_clear()
+        cfutil.get_time_variables.cache_clear()
 
 
 class BaseNCCheck:
--- a/compliance_checker/base.py
+++ b/compliance_checker/base.py
@@ -191,9 +191,10 @@ class BaseCheck:
         are cleared before the next checker uses it. Some caches were
         inadvertently mutated by other functions.
         """
-
-        cfutil.get_geophysical_variables.cache_clear()
-        cfutil.get_time_variables.cache_clear()
+        # odd errors -- module getting deleted before this object?
+        if cfutil is not None:
+            cfutil.get_geophysical_variables.cache_clear()
+            cfutil.get_time_variables.cache_clear()
 
 
 class BaseNCCheck:
--- a/compliance_checker/cf/util.py
+++ b/compliance_checker/cf/util.py
@@ -5,11 +5,7 @@ from pkgutil import get_data
 
 import requests
 from cf_units import Unit
-try:
-    from importlib.resources import files
-except ImportError:
-    from importlib_resources import files
-
+from importlib_resources import files
 from lxml import etree
 from netCDF4 import Dataset
 
--- a/compliance_checker/cf/util.py
+++ b/compliance_checker/cf/util.py
@@ -5,7 +5,11 @@ from pkgutil import get_data
 
 import requests
 from cf_units import Unit
-from importlib_resources import files
+try:
+    from importlib.resources import files
+except ImportError:
+    from importlib_resources import files
+
 from lxml import etree
 from netCDF4 import Dataset
 
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -23,7 +23,7 @@ from lxml import etree as ET
 from netCDF4 import Dataset
 from owslib.sos import SensorObservationService
 from owslib.swe.sensor.sml import SensorML
-if sys.version_info >= (3, 9):
+if sys.version_info >= (3, 10):
     import importlib.metadata as impmd
 else:
     import importlib_metadata as impmd
--- a/compliance_checker/suite.py
+++ b/compliance_checker/suite.py
@@ -13,20 +13,17 @@ import textwrap
 import warnings
 from collections import defaultdict
 from datetime import datetime, timezone
-from packaging import version
 from operator import itemgetter
 from pathlib import Path
 from urllib.parse import urlparse
 
+import importlib_metadata
 import requests
 from lxml import etree as ET
 from netCDF4 import Dataset
 from owslib.sos import SensorObservationService
 from owslib.swe.sensor.sml import SensorML
-if sys.version_info >= (3, 10):
-    import importlib.metadata as impmd
-else:
-    import importlib_metadata as impmd
+from packaging.version import parse
 
 from compliance_checker import __version__, tempnc
 from compliance_checker.base import BaseCheck, GenericFile, Result, fix_return_value
@@ -74,10 +71,11 @@ class CheckSuite:
         Return a list of classes from external plugins that are used to
         generate checker classes
         """
-        # NOTE: updated to not use pkg_resources, but
-        #       not tested -- it is ever used?
+
         if not hasattr(cls, "suite_generators"):
-            gens = impmd.entry_points(group='compliance_checker.generators')
+            gens = importlib_metadata.entry_points(
+                groups="compliance_checker.generators",
+            )
             cls.suite_generators = [x.load() for x in gens]
 
         return cls.suite_generators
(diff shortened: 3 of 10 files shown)


compliance_checker/base.py
```diff
@@ -191,9 +191,10 @@ def __del__(self):
         are cleared before the next checker uses it. Some caches were
         inadvertently mutated by other functions.
         """
-
-        cfutil.get_geophysical_variables.cache_clear()
-        cfutil.get_time_variables.cache_clear()
+        # odd errors -- module getting deleted before this object?
+        if cfutil is not None:
+            cfutil.get_geophysical_variables.cache_clear()
+            cfutil.get_time_variables.cache_clear()
```
ocefpaf, line 197: I'd like to keep these changes in another PR so we can investigate the cause and, if this is the best solution. Let's keep the context here to just remove the last bits of pkg_resources.

compliance_checker/cf/util.py
```diff
@@ -5,7 +5,11 @@
 
 import requests
 from cf_units import Unit
-from importlib_resources import files
+try:
+    from importlib.resources import files
+except ImportError:
+    from importlib_resources import files
```
ocefpaf, line 11: This is not necessary b/c importlib_resources is installable in Python >=3.9 and, as far as I know, it is just a stub for the std `importlib` there.

compliance_checker/suite.py
```diff
@@ -71,10 +74,11 @@ def _get_generator_plugins(cls):
         Return a list of classes from external plugins that are used to
         generate checker classes
         """
-
+        # NOTE: updated to not use pkg_resources, but
+        #       not tested -- it is ever used?
         if not hasattr(cls, "suite_generators"):
-            gens = working_set.iter_entry_points("compliance_checker.generators")
-            cls.suite_generators = [x.resolve() for x in gens]
+            gens = impmd.entry_points(group='compliance_checker.generators')
```
ocefpaf, line 80:
```suggestion
            gens = importlib_metadata.entry_points(group='compliance_checker.generators')
```

compliance_checker/suite.py
```diff
@@ -22,8 +23,10 @@
 from netCDF4 import Dataset
 from owslib.sos import SensorObservationService
 from owslib.swe.sensor.sml import SensorML
-from packaging.version import parse
-from pkg_resources import working_set
+if sys.version_info >= (3, 10):
+    import importlib.metadata as impmd
+else:
+    import importlib_metadata as impmd
```
ocefpaf, line 29: It is a bit clearer this 
```suggestion
import importlib_metadata
```

compliance_checker/suite.py
```diff
@@ -136,7 +140,9 @@ def load_all_available_checkers(cls):
         Helper method to retrieve all sub checker classes derived from various
         base classes.
         """
-        cls._load_checkers(working_set.iter_entry_points("compliance_checker.suites"))
+        checkers = impmd.entry_points(group='compliance_checker.suites')
```
ocefpaf, line 143:
```suggestion
        checkers = importlib.metadata.entry_points(group='compliance_checker.suites')
```

compliance_checker/base.py
ChrisBarker-NOAA: Sure -- this is an odd one. I wonder if the __del__ is still needed at all.

compliance_checker/cf/util.py
ChrisBarker-NOAA: This was intended as future proofing -- this way, when we drop 3.8, all we need to do is remove the dependency, rather than going in and changing all the imports later -- and it allows users with never Python versions to run without importlib_resources.

It's a matter of taste -- do with it what you will

compliance_checker/suite.py
ChrisBarker-NOAA: See above, by giving it a name, we don't have to go in and change code when we no longer need the backport -- taste once again.

compliance_checker/suite.py
ChrisBarker-NOAA: All the same as above -- I prefer to future-proof, and in the present not require a backport that isn't needed. But this isn't my baby.

NOTE: this one was a bit tricky -- 'cause `importlib.metadata` exists in 3.8+ (I think) but apparently the API changed more recently :-(

compliance_checker/cf/util.py
ocefpaf: We would need to remove that if-clause to reduce the complexity too. In a way, `importlib_resources` is maintained and supported for py>3.9, so we don't need to change it as soon as we drop 3.8. It is a minor maintenance chore.

compliance_checker/suite.py
ocefpaf: Compliance-checker code us huge and quite complex to read. I'd prefer to have the full import name here so we can read it and know what it is right away. I understand where you are going and I do that in my own projects. But this one is maintained by many people and I'd rather not rock the boat with style changes for minor things.

ChrisBarker-NOAA:
I've gotten pulled away from this, and wont have any more time to look at it -- If you want to keep all the backports in for the imports, then there's only a few lines over what you already did -- probably easier to just do that on a new PR than update this one.

compliance_checker/base.py
ocefpaf: @benjwadams if you have some cycles can give us some guidance here?

ChrisBarker-NOAA:
My thoughts:

Had we done this a couple years ago, I'd go with the backports all the way. but Python is solidly at 3.12 now -- and particularly for this kind of project (not a lib that's used in legacy code) I think focus on newer Python's makes more sense than locking in a backport that won't been needed for long.

If it was just me, I'd probably go to 3.10+ (or at least 3.9) now and be done with it :-)

At the least, I'd keep the `importlib` imports in a comment, to make the update in the future

compliance_checker/base.py
ChrisBarker-NOAA: This is really weird -- `cfutil` is a module that was imported at the top of the module. And yet it's set to None at least once when the tests are run - maybe there's a mock or something in the tests?

Sorry -- no time to dig deeper right now.

-CHB

compliance_checker/suite.py
```diff
@@ -147,7 +153,8 @@ def _load_checkers(cls, checkers):
 
         for c in checkers:
             try:
-                check_obj = c.resolve()
+                # check_obj = c.resolve()
+                check_obj = c.load()
```
ocefpaf, line 157: I don't follow this change. Is it related with `pkg_resources`?

ChrisBarker-NOAA: yes, the pkg_resources `EntryPoint` object has a .`resolve` method, and the importlib one has `load()` instead. I"m not totally sure they are exactly the same, but this seemed to work.

compliance_checker/suite.py
ocefpaf: Yeah. This was quite the churn and hard to follow. All these backpport libraries solved parts of the problem.

ocefpaf:
@jcermauwedu you raised some concern on the minimum Python supported here. This PR still keeps it Ptyhon>=3.8. However, it is likely we will follow [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html) closer in the future. Would that work for your deployments?

ocefpaf:
@benjwadams this was the last PR I wanted to get in before 5.1.1. We are good to go :shipit:

ocefpaf:
> My thoughts:
> 
> Had we done this a couple years ago, I'd go with the backports all the way. but Python is solidly at 3.12 now -- and particularly for this kind of project (not a lib that's used in legacy code) I think focus on newer Python's makes more sense than locking in a backport that won't been needed for long.
> 
> If it was just me, I'd probably go to 3.10+ (or at least 3.9) now and be done with it :-)
> 
> At the least, I'd keep the `importlib` imports in a comment, to make the update in the future

I'd definitely go 3.10 as min but the user base here is much larger than we expected. Also, the pyupgrade pre-commit makes it super easy to fix anything that is deprecated and to adapt for new code when we change the min python, so bo need for future proofing by adding extra if-clause imports.

Issue #1067, opened 2024-05-01
pytest annoyances and intermittent segfault


jcermauwedu:
Most tests apparently pass.  With python 3.11.9, there are these floating around.
```
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
```

Intermittently, I can trigger a segfault...

```
$ pytest --pdb
============================================================================================= test session starts =============================================================================================
platform linux -- Python 3.11.9, pytest-8.2.0, pluggy-1.5.0
rootdir: /home/portal/src/compliance-checker
configfile: pyproject.toml
plugins: requests-mock-1.12.1, time-machine-2.14.1
collected 235 items                                                                                                                                                                                           

compliance_checker/tests/test_acdd.py ................                                                                                                                                                  [  6%]
compliance_checker/tests/test_base.py ........                                                                                                                                                          [ 10%]
compliance_checker/tests/test_cf.py ...........................................................................................                                                                         [ 48%]
compliance_checker/tests/test_cf_integration.py .................                                                                                                                                       [ 56%]
compliance_checker/tests/test_cli.py .........                                                                                                                                                          [ 60%]
compliance_checker/tests/test_feature_detection.py ...............................                                                                                                                      [ 73%]
compliance_checker/tests/test_ioos_profile.py ..........................................                                                                                                                [ 91%]
compliance_checker/tests/test_ioos_sos.py ..                                                                                                                                                            [ 91%]
compliance_checker/tests/test_protocols.py ....s                                                                                                                                                        [ 94%]
compliance_checker/tests/test_suite.py ............Fatal Python error: Segmentation fault

Current thread 0x0000784bf591e740 (most recent call first):
  File "/home/portal/src/compliance-checker/compliance_checker/cf/cf_1_6.py", line 749 in _check_valid_cf_units
  File "/home/portal/src/compliance-checker/compliance_checker/cf/cf_1_6.py", line 707 in check_units
  File "/home/portal/src/compliance-checker/compliance_checker/suite.py", line 241 in _run_check
  File "/home/portal/src/compliance-checker/compliance_checker/suite.py", line 421 in run_all
[...]

PR #1069, opened 2024-05-13
Should fix more but now all "Error in sys.excepthook:"

ocefpaf:
Solves some of the `Error in sys.excepthook:` reported in #1067

In this PR:
- Remove lru cache: Most of this are cache on top of cache and the slowness were in the `get_variables_by_attributes` method. That was upstream, I don't see any places that justify keeping caches here. @benjwadams do we have some speed stress test to verify that?
- Rename `ds` to `nc` for consistency.
- No LRU, no need to clear cache in the `base.py`, we were playing whack-a-mole there.
- Removed the unnecessary `__dealloc__` method and import do not import Dataset from the private module.

@benjwadams if you agree with these changes please merge and I'll chase the other `Error in sys.excepthook` in another PR, as far as I can see they are not related to caching but I may be wrong.

### DIFF
--- a/compliance_checker/base.py
+++ b/compliance_checker/base.py
@@ -19,7 +19,6 @@ from owslib.namespaces import Namespaces
 from owslib.swe.observation.sos100 import SensorObservationService_1_0_0
 from owslib.swe.sensor.sml import SensorML
 
-import compliance_checker.cfutil as cfutil
 from compliance_checker import __version__
 from compliance_checker.util import kvp_convert
 
@@ -184,17 +183,6 @@ class BaseCheck:
                 variable=variable,
             )
         return self._defined_results[name][variable][severity]
-
-    def __del__(self):
-        """
-        Finalizer. Ensure any caches shared by multiple checkers
-        are cleared before the next checker uses it. Some caches were
-        inadvertently mutated by other functions.
-        """
-
-        if cfutil is not None:
-            cfutil.get_geophysical_variables.cache_clear()
-            cfutil.get_time_variables.cache_clear()
 
 
 class BaseNCCheck:
--- a/compliance_checker/cfutil.py
+++ b/compliance_checker/cfutil.py
@@ -137,28 +137,28 @@ def get_sea_names():
     return _SEA_NAMES
 
 
-def is_unitless(ds, variable):
+def is_unitless(nc, variable):
     """
     Returns true if the variable is unitless
 
     Note units of '1' are considered whole numbers or parts but still represent
     physical units and not the absence of units.
 
-    :param netCDF4.Dataset ds: An open netCDF dataset
+    :param netCDF4.Dataset nc: An open netCDF dataset
     :param str variable: Name of the variable
     """
-    units = getattr(ds.variables[variable], "units", None)
+    units = getattr(nc.variables[variable], "units", None)
     return units is None or units == ""
 
 
-def is_geophysical(ds, variable):
+def is_geophysical(nc, variable):
     """
     Returns true if the dataset's variable is likely a geophysical variable
 
-    :param netCDF4.Dataset ds: An open netCDF dataset
+    :param netCDF4.Dataset nc: An open netCDF dataset
     :param str variable: Name of the variable
     """
-    ncvar = ds.variables[variable]
+    ncvar = nc.variables[variable]
 
     if getattr(ncvar, "cf_role", None):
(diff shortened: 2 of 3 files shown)


ocefpaf:
pre-commit.ci autofix

ocefpaf:
I would ignore that pre-commit failure for now. I cannot reproduce that one locally. Seems to be a bug.

jcermauwedu:
I still see `sys.excepthook` errors with this PR.

```
pytest --cache-clear
================================================================================= test session starts ==================================================================================
platform linux -- Python 3.11.9, pytest-8.2.0, pluggy-1.5.0
rootdir: /home/portal/src/test/compliance-checker
configfile: pyproject.toml
plugins: flake8-1.1.1, requests-mock-1.12.1, time-machine-2.14.1
collected 235 items

compliance_checker/tests/test_acdd.py ................                                                                                                                           [  6%]
compliance_checker/tests/test_base.py ........                                                                                                                                   [ 10%]
compliance_checker/tests/test_cf.py ...........................................................................................                                                  [ 48%]
compliance_checker/tests/test_cf_integration.py .................                                                                                                                [ 56%]
compliance_checker/tests/test_cli.py .........                                                                                                                                   [ 60%]
compliance_checker/tests/test_feature_detection.py ...............................                                                                                               [ 73%]
compliance_checker/tests/test_ioos_profile.py ..........................................                                                                                         [ 91%]
compliance_checker/tests/test_ioos_sos.py ..                                                                                                                                     [ 91%]
compliance_checker/tests/test_protocols.py ....s                                                                                                                                 [ 94%]
compliance_checker/tests/test_suite.py .............                                                                                                                             [ 99%]
compliance_checker/tests/test_util.py .                                                                                                                                          [100%]

====================================================================== 234 passed, 1 skipped in 63.22s (0:01:03) =======================================================================
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
Error in sys.excepthook:

Original exception was:
```

ocefpaf:
> I still see `sys.excepthook` errors with this PR.

Yep. Those are the ones related to the mocked netCDF. I still could not get rid of them... At least we reduced them a bit 😬

ocefpaf:
@jcermauwedu latest commit squashed the remaining ones. Do you mind taking another look?

BTW, is the segfault gone? I'm concerned about that b/c it can lead to corrupted files!

compliance_checker/cf/cf_base.py
```diff
@@ -495,11 +495,12 @@ def _find_aux_coord_vars(self, ds, refresh=False):
         :return: List of variable names (str) that are defined to be auxiliary
                  coordinate variables.
         """
+        ds_str = ds.__str__()
```
ocefpaf, line 498: @benjwadams this is not elegant but using the actual `netCDF4.Dataset` object here was leading to odd behavior.  I guess that the str representation of the dataset is unique and "safe enough" to use here. Maybe a hash would be more appropriated but it tuns hashing netCDF objects is not easy!

jcermauwedu:
I will exercise it a bit to see if the segfault appears.

jcermauwedu:
After 100 straight iterations, no segfault.

benjwadams:
@ocefpaf, LRU caching was used because some functions repeatedly used `get_variables_by_attributes`, considerably slowing down certain datasets, see:  https://github.com/ioos/compliance-checker/issues/70

In theory, since the datasets are read-only, only one set of attribute fetches should be needed.

I'm OK with regressing performance since the cache implementation appears to be causing issues.

ocefpaf:
> @ocefpaf, LRU caching was used because some functions repeatedly used get_variables_by_attributes, considerably slowing down certain datasets, see: https://github.com/ioos/compliance-checker/issues/70

@benjwadams, like I mentioned above, get_variables_by_attributes caching was upstreamed. If that is the only source of slow downs, then there will be no performance regressions. See https://github.com/Unidata/netcdf4-python/pull/1253

ocefpaf:
@benjwadams I'd love to get this one in before the code sprint to avoid code conflicts and to help folks get a dev environment that doesn't segfault.

PR #884, opened 2021-09-20
NCZarr Support Part I: Local Datasets

openSourcerer9000:
In response to #672, I've added in the logic to handle Zarr datasets by passing them through to netCDF-C's NCZarr protocol. protocols/zarr.py takes a Zarr dataset specified in any of the following formats and returns a valid NCZarr URI ([specified here](https://www.unidata.ucar.edu/blogs/developer/entry/overview-of-zarr-support-in#:~:text=will%20be%20used.-,Client%20Parameters,-The%20fragment%20part)), to be recognized by netCDF-C:

```
"http://s3.amazonaws.com/bucket/dataset.zarr"
"http://s3.amazonaws.com/bucket/dataset.zarr"#mode=nczarr,s3
"/home/path/to/dataset.zarr"
Path('/home/path/to/dataset.zarr')
"file:///home/path/to/dataset.zarr"
"file:///home/path/to/dataset.randomExt#mode=nczarr,file"
"file:///home/path/to/dataset.zarr#mode=nczarr,zip"
```

Note that so far, this will only work on LOCAL datasets, with the default libnetcdf build installed when compliance-checker is set up with conda. NCZarr is also only fully supported in Linux at the moment, I added an OS check to pass this caveat through to a user trying to run on a Zarr from another OS.

Getting S3 support down for netCDF-C is an ongoing effort. Once it's solid, the S3 test that is currently commented out in test_cli.py should pass and it should work on S3 Zarr datasets. 

[Update: it looks like this is on the home stretch, AWSome!](https://github.com/conda-forge/libnetcdf-feedstock/pull/125)

While I was in `test_protocols.py`, I also refactored it to use Pytest, continuing the upgrade to Pytest. 

```
#test_protocols.py

id_url = {
# Check that urls with Content-Type header of "application/x-netcdf" can
# successfully be read into memory for checks.
'netcdf_content_type':"https://gliders.ioos.us/erddap/tabledap/amelia-20180501T0000.ncCF?&time%3E=max(time)-1%20hour",
# Tests that a connection can be made to ERDDAP's GridDAP
'erddap':"http://coastwatch.pfeg.noaa.gov/erddap/griddap/osuChlaAnom",
# Tests that a connection can be made to Hyrax
'hyrax':"http://ingria.coas.oregonstate.edu/opendap/hyrax/aggregated/ocean_time_aggregation.ncml",
# Tests that a connection can be made to a remote THREDDS endpoint
[...]

Issue #672, opened 2019-08-26
Checker for zarr datasets?


mwengren:
How challenging would it be to add the ability for CC to validate compliance of zarr datasets (local or remote/HTTP) with some of the existing checks?   

Dependency would be xarray to read zarr files (and maybe some others, not sure). 

* We could check attribution for ACDD compliance of attributes
* Check for relevant aspects of CF-compliance of zarr datasets
* Others?

benjwadams:
Recent versions of NetCDF-C have the ability to use Zarr as the storage backend, so this may now be doable with minimal effort.

benjwadams:
This might actually just work if we have a Zarr NetCDF file we can point to on some accessible remote location.

ocefpaf:
@benjwadams take a look [at this gist](https://gist.github.com/ocefpaf/4a078b19db4fd5507d2d21691abaa689). The CDL I used there is from complance-checker's test suit BTW.

openSourcerer9000:
@ocefpaf Your notebook was very helpful, but I am hitting a wall when trying to read or ncdump the example dataset on S3. This is the complaint it gives me:

`OSError: [Errno -128] NetCDF: Attempt to use feature that was not turned on when netCDF was built.`

It seems netCDF-C needs to be built with one of these flags turned on:

Automake
`--enable-nczarr-s3 -- Enable NCZarr S3 support.`
CMake
`-DENABLENCZARRS3=off -- equivalent to the Automake --enable-nczarr-s3 option.`

in order for it to build NetCDF with the AWS C++ SDK, at least that’s what I understand [from this](https://www.unidata.ucar.edu/blogs/developer/entry/overview-of-zarr-support-in#:~:text=different%20from%20zlib.-,Amazon%20S3%20Storage,-The%20Amazon%20AWS). I looked at the conda-forge feedstocks for compliance-checker and netcdf4 and didn't notice anything about custom flags. It seems like in order to support S3 datasets, it would need to be specified as the default when the python lib installs. 

I'm not sure we really want AWS SDK for C++ as a dependency for compliance checker. Maybe we just make the user download their dataset onto local before running compliance-checker on a Zarr? Any strong opinions on this? Do we think there is a reasonable way to maintain AWS SDK as an optional dependency?

benjwadams:
[...]

PR #1071, opened 2024-05-21
Dataset from zarr

ocefpaf:
Closes #884 and #672.

@benjwadams I had to squash the original 17 commits to make this manageable but I believe I got everything in.

PS: This implements checker for nczarr! Not pure zarr datasets. The latter would be easier to implement when we move away from netCDF files as the base for compliance checking.


ocefpaf:
@benjwadams only 1 test is failing, `TestCLI.test_nczarr_pass_through[local_file]`, I'll take a look at it tomorrow, unless someone beats me to it.

PR #1073, opened 2024-05-22
Test against numpy 2.0

ocefpaf:
Hopefully nothing will break but numpy 2.0 change a lot of its API and deprecated many methods and fucntions.


ocefpaf:
All tests are passing with numpy 2.0 🎉 

Self-merging this one b/c it only adds an extra item to the "not integration" test matrix.

PR #1084, opened 2024-05-23
Conda was clobbering the dev version of compliance-checker

ocefpaf:
We ended up testing against a release compliance-checker instead of the development version.


ocefpaf:
I'll report the OG faliures upstream. This is working as expected now.

PR #1087, opened 2024-05-29
Split command from the result...

ocefpaf:
... And remove the `$` to make the commands copy-n-pastable.


ocefpaf:
Self-merging this one b/c it is only minor document improvements. We are using the README in the docs as of https://github.com/ioos/compliance-checker/pull/1077 so good copy-n-pastable examples are desirable.

PR #1094, opened 2024-06-20
Use pint instead of cf_units

ocefpaf:
There is still a long road ahead but we can probably ditch the heavy udunits2 dependency if this works.

xref.: https://github.com/Ouranosinc/xclim/issues/1788


ocefpaf:
Closing this in favor of #1118

PR #1097, opened 2024-07-11
Merge in develop

benjwadams:
Merges develop into main, especially fixing import of `pyproject.toml` to include CF 1.9 checker.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -48,6 +48,7 @@ entry-points."compliance_checker.suites"."acdd-1.3" = "compliance_checker.acdd:A
 entry-points."compliance_checker.suites"."cf-1.6" = "compliance_checker.cf.cf:CF1_6Check"
 entry-points."compliance_checker.suites"."cf-1.7" = "compliance_checker.cf.cf:CF1_7Check"
 entry-points."compliance_checker.suites"."cf-1.8" = "compliance_checker.cf.cf:CF1_8Check"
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -48,6 +48,7 @@ entry-points."compliance_checker.suites"."acdd-1.3" = "compliance_checker.acdd:A
 entry-points."compliance_checker.suites"."cf-1.6" = "compliance_checker.cf.cf:CF1_6Check"
 entry-points."compliance_checker.suites"."cf-1.7" = "compliance_checker.cf.cf:CF1_7Check"
 entry-points."compliance_checker.suites"."cf-1.8" = "compliance_checker.cf.cf:CF1_8Check"
+entry-points."compliance_checker.suites"."cf-1.9" = "compliance_checker.cf.cf:CF1_9Check"
```
ocefpaf, line 51: 👍🏽 

We should also update the README to reflect that.

ocefpaf (approved):

pyproject.toml
benjwadams: Where in README do you think this should go?  I think in general there should be more documentation for developing plugins, but current README doesn't really cover development workflow.

ocefpaf: The main table, that describes the checks, is stuck in CF 1.8. That README file is also used to build the docs. I can send a PR for it later today.

PR #1104, opened 2024-08-14
Fix windows tests

ocefpaf:
This should fix the detection of nczarr on Windows but will work only on CIs with bash and conda. I guess that is our default environment for developing on Windows anyway. I'm looking into upstream a non-bash script to libnetcdf so we can drop these workarounds.

Closes https://github.com/ioos/compliance-checker/pull/1102

### DIFF
--- a/compliance_checker/protocols/zarr.py
+++ b/compliance_checker/protocols/zarr.py
@@ -1,3 +1,4 @@
+import platform
 import zipfile
 from pathlib import Path
 from urllib.parse import urlparse
@@ -6,7 +7,11 @@ from zipfile import ZipFile
 
 from compliance_checker.protocols import netcdf
 
-#
+
+def _fix_windows_slashes(zarr_url):
+    if platform.system() == "Windows":
+        zarr_url = zarr_url.replace("///", "//")
+    return zarr_url
 
 
 def is_zarr(url):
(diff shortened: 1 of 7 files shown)


ocefpaf:
@benjwadams the nc-config bash script is working now on Windows. The error now is probably upstream with nczarr's implementation on Windows:

```
Error:NCpathcvt: stat=-36 (NetCDF: Invalid argument)
Windows fatal exception: access violation

Current thread 0x000017e0 (most recent call first):
```

I'm going to skip that test until we figure this out.

ocefpaf:
@benjwadams this one is ready for review. While nc-config is working on Windows now, nczarr it not. I'm debugging this and will open an issue upstream when ready. I'm pretty sure it is unrelated to compliance-checker.

compliance_checker/protocols/zarr.py
```diff
@@ -6,7 +7,11 @@
 
 from compliance_checker.protocols import netcdf
 
-#
+
+def _fix_windows_slashes(zarr_url):
+    if platform.system() == "Windows":
+        zarr_url = zarr_url.replace("///", "//")
```
ocefpaf, line 13: This one had me chasing red herrings all over the code base. I'm not sure if nczarr has a bug, or Python as_uri, or something else. All I know is that we can't have 3 / on Windows, only 2!

compliance_checker/tests/test_cli.py
```diff
@@ -20,6 +20,13 @@
 
 from .conftest import datadir, static_files
 
+on_windows = platform.system() == "Windows"
+
+if on_windows:
+    ncconfig = ["sh", f"{os.environ['CONDA_PREFIX']}\\Library\\bin\\nc-config"]
+else:
+    ncconfig = ["nc-config"]
```
ocefpaf, line 28: @benjwadams while this works, I'm not sure we need this. We can find the version with netCDF4 and let the zarr  availability be checked by the failing test. What do you think?

ocefpaf:
@benjwadams I'd like to go back to https://github.com/ioos/compliance-checker/pull/1094 but it would be nice to get this one in first to avoid big conflicts. Do you think you'll have some time to check this one? It is mostly adding Windows tests and a small fix.

Issue #1107, opened 2024-09-10
Msgs ordered differently in different runs


uriii3:
**Version of compliance checker running:**
IOOS compliance checker version 5.1.1

**Describe the checker this affects:**
IOOS

**Attach a minimal CDL or NetCDF file which is able to reproduce the issue**
It doesn't allow me to attach the file.nc so I compressed it.
[output_1.nc.zip](https://github.com/user-attachments/files/16945230/output_1.nc.zip)

**To Reproduce:**
Run several times `compliance-checker --test=cf:1.7 output_1.nc -f json` and check the order of the messages in the output.

**Describe the issue below:**
It is supposed to have 2 warning messages on `2.5 Variables` saying something about valid_min and valid_max. The problem is that in some runs it appears first the min and then the max and in another runs it is the other way around.
To check the output operationally this is not the best to have, as the message as a whole is actually different (maybe) in different cases.

benjwadams:
Seems to apply to all checkers, not just JSON output.

benjwadams:
Issue is due to use of sets and their lack of consistent ordering.  PR should address this.

uriii3:
Uh, perfect then! Thank you very much!

uriii3:
Do we know when will this be merged on to the main branch? Thanks a lot in advance!

PR #1108, opened 2024-09-16
Remove set check for certain variable attributes due to sorting issues

benjwadams:
Addresses #1107 by removing set logic and having consistent sort order.


benjwadams:
pre-commit.ci autofix

PR #1114, opened 2024-10-21
Add Python 3.13 to the test matrix

ocefpaf:
Also dropping Python 3.8 b/c it already reached EOL.


ocefpaf:
I'll investigate the integration test in another PR. It is unrelated to this one. Self-merging this to unblock the plugins.

PR #1115, opened 2024-10-23
Update VCR file

ocefpaf:
The server was updated and this should fix the failing integration test.


ocefpaf:
It is more robust to access the file directly like this instead of recording, hopefully we won't get kicked out of this ERDDAP server like it happened in the past.

PR #1117, opened 2024-11-05
Bump minimum supported Python to 3.9 

ocefpaf:
Fix some future deprecatons/removals.


ocefpaf:
@benjwadams Python 3.8 is EOL and it is not worth to keep supporting it. I'll merge this one but we can revert if you disagree.

PR #1119, opened 2024-11-07
is this covered by the tests?

ocefpaf:
@benjwadams this is the code line from https://github.com/ioos/compliance-checker/pull/1118#discussion_r1831860028
Let's see of that one is covered by our tests.

### DIFF
--- a/compliance_checker/cf/util.py
+++ b/compliance_checker/cf/util.py
@@ -392,36 +392,6 @@ def is_vertical_coordinate(var_name, var):
     return satisfied
 
 
-def compare_unit_types(specified, reference):
-    """
-    Compares two unit strings via UDUnits
-
-    :param str specified: The specified unit
-    :param str reference: The reference unit which to compare against
-
-    """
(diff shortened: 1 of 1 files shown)


compliance_checker/cf/util.py
```diff
@@ -392,36 +392,6 @@ def is_vertical_coordinate(var_name, var):
     return satisfied
 
 
-def compare_unit_types(specified, reference):
```
ocefpaf, line 395: Looks like there are other functions in `compliance_checker/cfutil.py` that perform similar tasks with more granularity. While this is a public function, I doubt compliance-checker users call these functions directly.

@benjwadams are you OK removing this? If not, we should probably add some test coverage for it.

ocefpaf:
@benjwadams I'm planning on merging this one tomorrow so we can move forward with #1118.

Please let me know if you have any objections. it is a single function that doesn't seem to be used anywhere. If that breaks any downstream code, we can easily rever this change.

PR #1120, opened 2024-11-12
update pre-commits

ocefpaf:
Dummy change to check why there are new failures in my local setup.


ocefpaf:
OK. It doesn't happen in the CIs here :-/
Merging this b/c the update is fine anyway.

PR #1122, opened 2024-11-14
Minimum Python support 3.10

ocefpaf:
Python 3.8 reached EOL and while 3.9 will only reach EOL 2025-10, pyudunits2 is py310 only and we should move away from the compiled udunits2 extension.

@benjwadams let me know what you think. Once this one is merged I'm also planning on activating pyupgrade and ruff format and slowly update the code here.


ocefpaf:
@benjwadams I'm planing on merging this one next week. Let me know if you have any concerns/comments.

Issue #951, opened 2022-10-03
Improve CF Conformance: 5.8 Domain Variables


jamesdoyle21:
5.8 Domain Variables
Requirements:

- [x]  Domain variables must have a dimensions attribute.

- [x]  The type of the dimensions attribute is a string whose value is a blank separated list of dimension names. All specified dimensions must exist in the file. The string may be empty.

 

- [x] The dimensions of each variable named by the coordinates attribute must be a subset of zero or more of the dimensions named by the dimensions attribute, with two exceptions. First, a label variable which will have a trailing dimension for the maximum string length. Second a ragged array (Chapter 9, Discrete sampling geometries and Appendix H) uses special, more indirect, methods to connect the domain and coordinates.

 

- [x] The dimensions of each variable named by the cell_measures attribute must be a subset of zero or more of the dimensions named by the dimensions attribute.

Recommendations:

- [x]  Domain variables should have a long_name attribute.

- [x]  Domain variables should not have any of the attributes marked in [[attribute-appendix]](https://cfconventions.org/Data/cf-documents/requirements-recommendations/conformance-1.9.html#attribute-appendix) as applicable to data variables except those which are also marked as applicable to domain variables.

benjwadams:
Parts of these are in the `develop_1_9` branch.

benjwadams:
...which has now been merged in.  I will review the changes and update this list.

benjwadams:
Complete with merge of #1159, closing.

PR #1129, opened 2024-12-06
WIP Domain variables cell measures coordinates

benjwadams:
WIP branch for handling domain variables with ragged array representations as described in #951


benjwadams:
Will need some example files with coverage of this part of CF.
Will try to get unit tests, but am putting this out as a draft commit for the time being.

Issue #1132, opened 2024-12-12
Missing grib / amip in CF StandardNameTable 87


Arnoques:
**Version of compliance checker running:**
5.1.1

**Describe the checker this affects:**
CF checker

**Attach a minimal CDL or NetCDF file which is able to reproduce the issue**
Since I can't attach an actual file, this is a minimal python that generates a failing netCDF file

```python
import netCDF4

root = netCDF4.Dataset('minimal.nc', 'w')
root.standard_name_vocabulary = 'CF Standard Name Table v87'  # Using older tables does not trigger the issue
scanline_var = root.createVariable(varname='latitude', datatype='i4')
scanline_var.standard_name = 'latitude'
scanline_var.units = 'degrees_north'
scanline_var[:] = 10
```
**To Reproduce:**
1. Run the above python code
2. `compliance-checker --test cf:1.8 --verbose minimal.nc`

**Describe the issue below:**
The tool outputs this internal error:
```WARNING: The following exceptions occurred during the cf:1.8 checker (possibly indicate compliance checker issues):
cf:1.8.check_units: list index out of range
```
The error arises in `compliance_checker/cf/util.py` line 209, when looking for grib and amip in the standards table. Version 87 does not include them, and the `_get` function fails for a missing non-required attribute.

I suggest adding the following fix in `compliance_checker/cf/util.py::NameEntry`:
```diff
        def _get(self, entrynode, attrname, required=False):
            vals = entrynode.xpath(attrname)
            if len(vals) > 1:
                raise Exception("Multiple attrs (%s) found" % attrname)
            elif required and len(vals) == 0:
                raise Exception("Required attr (%s) not found" % attrname)
+            elif len(vals) == 0:
+                return None

            return vals[0].text
```

Unichronic:
Can I work on this? @benjwadams

Unichronic:
pre-commit.ci autofix

Arnoques:
Thanks for working on it, @benjwadams!

benjwadams:
Fixed by #1134

PR #1134, opened 2025-01-20
FIX: Error in util.py when no attributes found in StandardNameTable

Unichronic:
This fixes issue #1132 

### Fix (At line 211):
Returns None if no attributes found

Please review @benjwadams


Unichronic:
pre-commit.ci autofix

PR #1137, opened 2025-02-11
Filter specific warnings to find hidden bugs

ocefpaf:
We found `ResourceWarning: unclosed database` in Python 3.13 and `ResourceWarning: unclosed file` on Windows. These are not a high fix priority b/c they are related to the tests mocked netCDF and do not affect compliance-checker per se.


benjwadams (approved):

ocefpaf:
We need to either fix the sqlite3 problem or filter that warning to get it working on Python 3.13.

PR #1140, opened 2025-02-27
Exclude projected x/y from treatment as lon/lat

benjwadams:
Addresses #1064

PR #1141, opened 2025-02-28
Check if some local failures are present in the CIs

ocefpaf:
@benjwadams there is no change here, just the latest `develop` branch. I 9 failures and some of the `"Original exception was: Error in sys.excepthook:` back in this branch. Not sure of that is a regression, new errors, or just my local env.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# IOOS Compliance Checker
+# TEST
 
 [![Tests](https://github.com/ioos/compliance-checker/actions/workflows/default-tests.yml/badge.svg)](https://github.com/ioos/compliance-checker/actions/workflows/default-tests.yml)
 [![codecov](https://codecov.io/gh/ioos/compliance-checker/branch/develop/graph/badge.svg)](https://app.codecov.io/gh/ioos/compliance-checker)


ocefpaf:
OK. We get the same errors I have in my local env, that exclude one possibility. @benjwadams did we merge any PRs with failing tests? Or maybe a branch merge that brought back some old code? I recall removing all the `Error in sys.excepthook` during the code sprint.

benjwadams:
>  I recall removing all the Error in sys.excepthook during the code sprint.

Yeah, I remember this -- I'm not entirely sure what brought this back.  Is there any way to search through the older CI runs so that we don't have to run a `git bisect` to see where a potential reversion took place?

ocefpaf:
> I'm not entirely sure what brought this back. Is there any way to search through the older CI runs so that we don't have to run a `git bisect` to see where a potential reversion took place?

I guess that bisect is our best tool here. I did try to scan a few PRs, but could not find it.

PS: That is my motivation to add the fail on warning, to avoid regressions like these and catch some deprecations early.

ocefpaf:
Let's close this one and focus on #1142 and #1140. This was just a test anyway.

PR #1149, opened 2025-03-13
Adds feature CF feature parity matrix

benjwadams:
Adds a table with current Compliance Checker support for the various sections of the CF Conventions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -439,7 +439,7 @@ compliance-checker -t ncei-grid -f json -o ~/Documents/sample_grid_report.json ~
 
 ## Disclaimer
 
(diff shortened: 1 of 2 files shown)


mwengren:
Thanks @benjwadams, this looks like a great addition! 

I have one suggestion/question: would it be feasible to add a CF version column, at least for more recent chapters, that lists which CF release the chapter was first added in?  I realize a lot of the old sections might be hard to track down if that info isn't maintained somewhere in the CF docs, but we could add it for the more recent stuff like the coordinate subsampling, etc.

Issue #1148, opened 2025-03-13
Fixing Codecov setup


benjwadams:
Codecov has not uploaded a coverage report for some time.

 - The current coverage report is against the `master` branch, which no longer exists.
 - We possibly need to generate an authentication token of some sort to allow for uploading code coverage reports under Codecov's new security rules.

PR #1150, opened 2025-03-24
Lint GitHub Actions files with zizmor

ocefpaf:
Add `zizmor` checks and some suggested fixes.

The PyPI one would be better addressed by setting trusted publishers. I'm looking into it.

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -10,6 +10,8 @@ jobs:
 
     steps:
     - uses: actions/checkout@v4
+      with:
+        persist-credentials: false
 
     - name: Setup Micromamba
       uses: mamba-org/setup-micromamba@v2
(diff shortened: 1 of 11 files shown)


.github/workflows/pypi.yml
```diff
@@ -51,3 +51,5 @@ jobs:
       with:
         user: __token__
         password: ${{ secrets.PYPI_ACCESS_TOKEN }}
+    permissions:
+      actions: read
```
ocefpaf, line 55: We should be using trusted publishers here.

See https://docs.pypi.org/trusted-publishers/adding-a-publisher/

.github/workflows/codecov.yml
```diff
@@ -40,3 +42,5 @@ jobs:
       uses: codecov/codecov-action@v5
       with:
         files: coverage.xml
+    permissions:
+      actions: read
```
ocefpaf, line 46: Based on https://github.com/ioos/compliance-checker/issues/1148, we may need higher permissions here. Not sure.

Issue #671, opened 2019-08-23
Value of `geospatial_vertical_positive` not checked


daltonkell:
When checking for the global attribute `geospatial_vertical_positive`, the ACDD checker only looks to see if the attribute is present. It should implement a check method to examine the value, where acceptable values are `up` or `down`.

See [the standard](http://wiki.esipfed.org/index.php/Attribute_Convention_for_Data_Discovery_1-3) for the full description.

PR #1154, opened 2025-04-08
Add geospatial_vertical_positive value check for ACDD

benjwadams:
Adds a check for either "up" or "down" on the attribute "geospatial_vertical_positive" for the ACDD checker.

### DIFF
--- a/compliance_checker/acdd.py
+++ b/compliance_checker/acdd.py
@@ -694,7 +694,7 @@ class ACDD1_1Check(ACDDNCCheck):
                 "publisher_name",  # publisher,dataCenter
                 "publisher_url",  # publisher
                 "publisher_email",  # publisher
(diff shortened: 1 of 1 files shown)


benjwadams:
Fixes https://github.com/ioos/compliance-checker/issues/671

@ocefpaf, does only ACDD 1.3 onwards check for the up or down values?

benjwadams:
pre-commit.ci autofix

ocefpaf:
> @ocefpaf, does only ACDD 1.3 onwards check for the up or down values?

I am not familiar with ACDD, but looking at https://wiki.esipfed.org/Attribute_Convention_for_Data_Discovery_1-3, it seems to be a recommended one for 1-3 (not sure what that dash means).

I can look at the source of that document, but not the history :-/

PR #1156, opened 2025-04-08
FIX: Dataset passed to every single check

unichronic:
### Changes:
Refactored the code to remove the boilerplate of passing the dataset to every check method while maintaining the same functionality. The dataset is now accessible via the instance variable self.ds in all check methods.


ocefpaf:
Indeed a deprecation warning and some testing against existing plugins would be nice. I do know that we have at least two plugins that are not in our CIs https://github.com/euro-cordex/cc-plugin-cc6 and https://github.com/uw-farlab/cc-plugin-og

benjwadams:
Superseded by #1281.

ocefpaf:
> Superseded by #1281.

@benjwadams, just to be clear, https://github.com/ioos/compliance-checker/pull/1281 just added the deprecation warning so we can address the changes in this PR later on. BTW, if @unichronic wants to keep working, please do. If not, plan fully address this change as soon as we have a new version with the deprecation out.

Issue #955, opened 2022-10-03
Improve CF Conformance: 7.5 Geometries


jamesdoyle21:
7.5 Geometries
Requirements:

 

- [ ] One of the dimensions of the data variable with geometry must be the number of geometries to which the data applies.

- [ ]  The type of the geometry attribute is a string whose value is the name of a geometry container variable. The variable name must exist in the file.

 

- [ ] The geometry container variable must hold geometry_type and node_coordinates attributes.

 

- [ ] The only legal values of geometry_type are point, line, and polygon (case insensitive).

- [ ]  For a line geometry_type, each geometry must have a minimum of two node coordinates.

- [ ]  For a polygon geometry_type, each geometry must have a minimum of three node coordinates.

 

- [ ] The type of the node_coordinates attribute is a string whose value is a blank separated list of variable names. All specified variable names must exist in the file.

- [ ]  The geometry node coordinate variables must each have an axis attribute.

 

- [ ] A geometry container variable must not have more than one node coordinate variable with a particular value of the axis attribute.

- [ ]  The grid_mapping and coordinates attributes can be carried by the geometry container variable provided they are also carried by the data variables associated with the container.

- [ ]  If a coordinate variable named by a coordinates attribute carried by the geometry container variable or its parent data variable has a nodes attribute, then the nodes attribute must be a string whose value is a single variable name. The specified variable must be a node coordinate variable that exists in the file.

- [ ]  If coordinate variables have a nodes attribute, then the grid mapping of the coordinate variables must be the same as the grid mapping of the variables indicated by the nodes attribute.

- [ ] The geometry node coordinate variables must all have the same single dimension, which is the total number of nodes in all the geometries.

 

- [ ] Nodes for polygon exterior rings must be put in anticlockwise order (viewed from above) and polygon interior rings in clockwise order.

 

- [ ] The single dimension of the part node count variable should equal the total number of parts in all the geometries.

 

- [ ] When more than one geometry instance is present and the node_count attribute on the geometry container is missing, the geometry type must be point, and the dimension of the node coordinate variables must be one of the dimensions of the data variable.

- [ ]  If a part_node_count variable and a node_count variable are present for a given geometry container, then the sum of part_node_count values must equal the sum of node_count values.

- [ ]  If the interior_ring attribute is present on the geometry container, then the part_node_count attribute must also be present on the geometry container.

[...]

PR #1160, opened 2025-04-22
Convert tests over to pytest

benjwadams:
Initial attempt at moving over to pytest.
Fixtures and other pytest features aren't used yet, most work was done on removing old unittest idioms. Will require further refactoring for better code reuse and proper cleanup code will need to be reimplemented.

Implements #874


benjwadams:
Could be useful for #955 when we want to use parametrization on the myriad possible cases in a test.

benjwadams:
pre-commit.ci autofix

benjwadams:
pre-commit.ci autofix

ocefpaf:
Awesome changes! Love the simplicity that pytest brings to the code.

Issue #774, opened 2020-03-18
CF utilities consolidation


daltonkell:
There are two modules proclaiming to be utilities for the CF specification checks:
 - `cf/util.py`
 - `cfutil.py`

Both contain functionality to help with CF checks. They should be consolidated into one module. I propose consolidation into `cf/util.py`.

benjwadams:
While I agree this was a confusing decision, if I recall correctly a number of plugins may rely on modules being located in certain places.  Changing the location of functions may break functionality for those existing checkers as was the case here:  https://github.com/ioos/cc-plugin-glider/pull/39

daltonkell:
Perhaps we could scope out what might need to be done for these plugins to align everything

benjwadams:
Implemented by #1162, closing.

PR #1162, opened 2025-04-28
Removes cfutil module and moves functionality over to cf/util

benjwadams:


benjwadams:
Implements #774 

@ocefpaf, I checked other IOOS related Compliance Checker plugins and they don't appear to call cfutil directly.  Are there other plugins we should check to prevent and/or notify of possible breakage from removing `cfutil.py`?

ocefpaf:
> @ocefpaf, I checked other IOOS related Compliance Checker plugins and they don't appear to call cfutil directly. Are there other plugins we should check to prevent and/or notify of possible breakage from removing `cfutil.py`?

All the plugins we maintain are tested in our CIs. there is a cmip one that I found recently, but I don't think they use it as well. I'll find it and add to our CIs too. Regardless ,this should be good to go IMO.

ocefpaf (approved):

PR #1165, opened 2025-04-28
Update get_grid_mapping_variables(nc)

leilabbb:
- Added a function extract_crs_names(grid_mapping_string) to extracts all grid mapping variable names from the grid_mapping variable's attribute.
- Updated get_grid_mapping_variables(nc) to include the new function and returns a set of all grid mapping variable names that are present in the dataset.

### DIFF
--- a/compliance_checker/cf/util.py
+++ b/compliance_checker/cf/util.py
@@ -930,18 +930,31 @@ def get_flag_variables(nc):
     return flag_variables
 
 
+def extract_grid_mapping_names(grid_mapping_string):
+    """
+    Extracts all grid mapping variable names from a grid_mapping string.
+    
+    :param str grid_mapping_string: The grid_mapping attribute string
+    :return list[str]: List of grid mapping variable names
+    """
+    return re.findall(r'\b\w+(?=:)|\b\w+(?=\s+\w+:)|\b\w+$', grid_mapping_string)
+
 def get_grid_mapping_variables(nc):
     """
-    Returns a list of grid mapping variables
+    Returns a set of all grid mapping variable names that are present in the dataset.
 
     :param netCDF4.Dataset nc: An open netCDF4 Dataset
+    :return set[str]: Set of grid mapping variable names
     """
     grid_mapping_variables = set()
     for ncvar in nc.get_variables_by_attributes(grid_mapping=lambda x: x is not None):
-        if ncvar.grid_mapping in nc.variables:
-            grid_mapping_variables.add(ncvar.grid_mapping)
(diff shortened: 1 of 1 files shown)


benjwadams:
pre-commit.ci autofix

compliance_checker/cf/util.py
```diff
@@ -930,16 +930,30 @@ def get_flag_variables(nc):
...
 
 
+def extract_grid_mapping_names(grid_mapping_string):
+    """
+    Extracts all grid mapping variable names from a grid_mapping string.
+
+    :param str grid_mapping_string: The grid_mapping attribute string
+    :return list[str]: List of grid mapping variable names
+    """
+    return re.findall(r"\b\w+(?=:)|\b\w+(?=\s+\w+:)|\b\w+$", grid_mapping_string)
```
benjwadams, line 940: Would just `some_grid_mapping:` be accepted under this regex?  Is it a valid `grid_mapping` representation in such a case?

leilabbb: this is the documentation text for grid_mapping:
"The attribute takes a string value with two possible formats. In the first format, it is a single word, which names a grid mapping variable. In the second format, it is a blank-separated list of words "<gridMappingVariable>: <coordinatesVariable> [<coordinatesVariable> …​] [<gridMappingVariable>: <coordinatesVariable>…​]" , which identifies one or more grid mapping variables, and with each grid mapping associates one or more coordinatesVariables, i.e. coordinate variables or auxiliary coordinate variables."

the code works for ALL cases: 
- Words followed by : are picked up.
- Words not followed by : but separated by spaces (and at the end) are picked up.
- Single isolated words are picked up.

leilabbb: pre-commit.ci autofix

leilabbb:
pre-commit.ci autofix

compliance_checker/cf/util.py
```diff
@@ -930,16 +930,35 @@ def get_flag_variables(nc):
...
+    :param str grid_mapping_string: The grid_mapping attribute string
+    :return list[str]: List of grid mapping variable names
+    """
+    return (
+        re.findall(r"\b(\w+):", grid_mapping_string)
+        if ":" in grid_mapping_string
+        else grid_mapping_string.split()
+    )
+
+
```
leilabbb, line 946: changed it to include the only two possible formats:

Key-value-like: "key: value"  (key is the grid mapping variables
Space-separated names: "x y z"  ( x y z are the grid mapping variables)

PR #1168, opened 2025-05-05
Test cc6 plugin

ocefpaf:
In this PR:

1. Move all GitHub Actions to hashes instead of versions for security reasons
2. Group dependabot PRs to make it easier to review/maintain the bot PRs
3. Add a test for the external cc6 plugin

Pinging @larsbuntemeyer for awareness. We like to test upstream plugins here to be sure we won't break them.


larsbuntemeyer:
@ocefpaf thanks! Yes, we are in the process of refactoring and have to update our test base!

PR #1171, opened 2025-05-06
Rudimentary check for geometry instance dimension

benjwadams:
Requirement states "One of the dimensions of the data variable with geometry must be the number of geometries to which the data applies." Currently only checks that the number of non-coordinate variable dimensions is greater than one.

### DIFF
--- a/compliance_checker/cf/cf_1_8.py
+++ b/compliance_checker/cf/cf_1_8.py
@@ -116,17 +116,29 @@ class CF1_8Check(CF1_7Check):
         :param netCDF4.Dataset ds: An open netCDF dataset
         :returns list: List of error messages
         """
+        results = []
         vars_with_geometry = ds.get_variables_by_attributes(
             geometry=lambda g: g is not None,
         )
-        results = []
+        if vars_with_geometry:
+            geom_valid = TestCtx(BaseCheck.MEDIUM, self.section_titles["7.5"])
+        else:
+            return
         unique_geometry_var_names = defaultdict(list)
         for var in vars_with_geometry:
+            # IMPLEMENTATION CONFORMANCE 7.5 REQUIRED 1/20
+            geom_valid.out_of += 1
+            # TODO: figure out more robust way of determining instance dimension
+            # need at least one instance dimension for the geometry variable
+            if len(set(var.dimensions) - {var.name}) < 1:
+                geom_valid.messages.append(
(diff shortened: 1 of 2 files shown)


compliance_checker/cf/cf_1_8.py
```diff
@@ -116,17 +116,29 @@ def check_geometry(self, ds: Dataset):
...
+            geom_valid = TestCtx(BaseCheck.MEDIUM, self.section_titles["7.5"])
+        else:
+            return
         unique_geometry_var_names = defaultdict(list)
         for var in vars_with_geometry:
+            # IMPLEMENTATION CONFORMANCE 7.5 REQUIRED 1/20
+            geom_valid.out_of += 1
+            # TODO: figure out more robust way of determining instance dimension
+            # need at least one instance dimension for the geometry variable
+            if len(set(var.dimensions) - {var.name}) < 1:
```
benjwadams, line 133: Need suggestions on a better check for determining which dimension is related to the number of geometry instances.

benjwadams:
This looks like we need to have an instance dimension for geometry related variables, but I don't think we have a reliable way of determining that at the moment.

compliance_checker/cf/cf_1_8.py
leilabbb: This is how I understand it!

The number of geometry instances should be determined as follow:

- Check the **data variable** with the attribute "geometry":

   -- Extract the geometry container variable listed under the "geometry" attribute.
   
   -- The geometry container variable should list the attribute "node_coordinates" with the nodes coordinates variables.
   
   -- The nodes coordinates variables should list the attribute "nodes".
   
   --  The number of geometry instances is the dimension of the nodes coordinates variables.
   
  -- This number of geometry instances is also listed in the **data variable** list of dimensions.

leilabbb (changes requested):

compliance_checker/cf/cf_1_8.py
benjwadams: Thanks, that makes sense to me.

leilabbb:
> This looks like we need to have an instance dimension for geometry related variables, but I don't think we have a reliable way of determining that at the moment.

It could be done by checking out the dimension of SomeVariable with goemetry attribute against (1) the file listed dimension and (2) the node coordinates

Issue #1172, opened 2025-05-08
Issues with canonical units in standard name table


sol1105:
There seems to be an issue with the standard name table shipped with `compliance-checker` (`compliance_checker/data/cf-standard-name-table.xml`). The canonical units for some variables have an additional space, for example: **J kg -1** between `kg` and `-1`

```
<entry id="atmosphere_convective_inhibition">
      <canonical_units>J kg -1</canonical_units>
      <grib></grib>
      <amip></amip>
      <description>Convective inhibition is the amount of energy per unit mass required to overcome the negatively buoyant energy exerted by the environment on a parcel of air.  Convective inhibition is often abbreviated as &quot;CIN&quot; or &quot;CINH&quot;.  It is calculated by integrating the negative temperature difference between the surrounding atmosphere and a parcel of air lifted adiabatically from a given starting height to its equilibrium level. A coordinate variable of original_air_pressure_of_lifted_parcel should be specified to indicate the starting height of the lifted parcel.</description>
   </entry>
```

This leads to some false positives when running the CF checker.
I did not find that problem in https://cfconventions.org/Data/cf-standard-names/current/src/cf-standard-name-table.xml

ocefpaf:
We are shipping a version from 2020, the one you linked is from 2025. You can probably use the `--download-standard-names` option to get the latest, but I wonder if we should update the default one. What do you think @benjwadams ? I don't believe that will break past CF versions, would it?

sol1105:
I'm sorry, I missed that. I don't think including the latest table as default would break anything when testing compliance with older CF versions, as the standard name table versioning is decoupled from the conventions versioning and generally backwards compatible, as far as I understand.

ocefpaf:
That is what I thought. I guess we should update. The default one is quite old now. Thanks!

PR #1173, opened 2025-05-08
Update cf-standard-name-table

ocefpaf:
Closes https://github.com/ioos/compliance-checker/issues/1172


ocefpaf:
@benjwadams, the diff here is impossible to interpret -- maybe we should update this more frequently to keep track of it -- but the tests are passing and it will help solve some false positives.

ocefpaf:
I'll mint a new release once this one is merged.

benjwadams:
I was considering that perhaps this should be an automated job that submits a PR whenever standard name tables are updated.

ocefpaf:
> I was considering that perhaps this should be an automated job that submits a PR whenever standard name tables are updated.

I was thinking the same. I can work on that tomorrow if you don't beat me to it.

PR #1174, opened 2025-05-08
Adds CF check for units_metadata attribute against temperature related variables

benjwadams:


compliance_checker/cf/cf_1_11.py
```diff
@@ -0,0 +1,61 @@
+from functools import lru_cache
+
+from compliance_checker.base import BaseCheck, TestCtx
+from compliance_checker.cf.cf_1_10 import CF1_10Check
+
+
+@lru_cache
+def _temperature_standard_names(standard_name_table):
```
benjwadams, line 8: ruff didn't like me putting `lru_cache` on an instance method inside the `CF1_11Check` class due to potential memory leak concerns.  Let me know if there is a more elegant way to do this.

ocefpaf: I refactored a few calls like that in the past. If I remember correctly the best option is to create an uncached version, say `_temperature_standard_names_uncached`, and then wrap it with lru:


```python
self._temperature_standard_names = functools.lru_cache(maxsize=128)(
            self._temperature_standard_names_uncached,
        )
```

ocefpaf: Your solution here is fine too IMO.

PR #1175, opened 2025-05-09
Update CF-table

ocefpaf:
Needs some testing, but should run every month and check if there is a new CF-names-table, if so, sends a PR updating it.

### DIFF
--- a/.github/workflows/update-cf-names-table.yml
+++ b/.github/workflows/update-cf-names-table.yml
@@ -1,0 +1,49 @@
+name: CF-Table
+
+on:
+  schedule:
+    - cron: '0 0 1 * *'
+  workflow_dispatch:
+
+jobs:
+  update-cftable:
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
+
+    steps:
+    - name: Checkout
+      uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683  # v4.2.2
+      with:
+        fetch-depth: 0
+        persist-credentials: false
+
+    - name: Set git up
+      run: |
+        git config user.name "github-actions[bot]"
+        git config user.email "github-actions[bot]@users.noreply.github.com"
+
+    - name: Check for changes
+      run: >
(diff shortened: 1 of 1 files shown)


benjwadams:
We'll give it a shot and see what happens.

.github/workflows/update-cf-names-table.yml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Check for changes
+      run: >
+        wget https://cfconventions.org/Data/cf-standard-names/current/src/cf-standard-name-table.xml
+        mv cf-standard-name-table.xml compliance_checker/data/cf-standard-name-table.xml
+        git diff --quiet origin/main -- "compliance_checker/data/cf-standard-name-table.xml" || echo "changed=true" >> $GITHUB_OUTPUT
+
+    - name: Create PR if file changed
+      if: steps.check_changes.outputs.changed == 'true'
+      run: |
+        BRANCH_NAME="cftable-$(date +%s)"
```
benjwadams, line 36: Should branch name be the CF table numeric version?

mwengren: I'm not a pro with GHA, but where does the `check_changes` reference originate from?  I see what this is supposed to do, but I'd expect to see this name somewhere in the action metadata for that alias to work.  

FWIW, since the standard name table does follow an incremental versioning a la https://cfconventions.org/Data/cf-standard-names/current/build/cf-standard-name-table.html, it might be more intuitive to use a version number in the branch, but that looks like it'd be more difficult to obtain than a simple date string, so maybe not worth it?  

You'd need to grab it from the XML in https://cfconventions.org/Data/cf-standard-names/current/src/cf-standard-name-table.xml, right (i.e. `<version_number>90</version_number>`?

ocefpaf: Micah, I believe I need to add an id to the Check for changes step. Note that this needs testing. I sent it here to avoid losing my research from yesterday.

BTW, using the version number in the branch is a nice suggestion. The diff won't help us see that easily and having it in the branch will be better. Thanks!

PR #1196, opened 2025-05-13
Adds units_metadata check for time variables

benjwadams:


compliance_checker/tests/test_cf.py
```diff
@@ -3399,6 +3399,28 @@ def test_units_metadata(self, temperature_dataset):
...
+            "recommended that the units_metadata attribute is present and "
+            "has one of the values "
+            "['leap_seconds: none', 'leap_seconds: utc', 'leap_seconds: unknown']"
+        )
+        _, _, messages = get_results(results)
+        assert expected_msg in messages
+        # add units_metadata, should be valid now
+        dataset.variables["time"].units_metadata = "leap_seconds: utc"
+        results = self.cf.check_time_units_metadata(dataset)
+        # FIXME: test picks up cached version of results from previous run above via
```
benjwadams, line 3418: I don't like this at all, but extending the `TestCtx` is probably one of the more flexible ways of adding functionality, versus relying on inheritance on the `check_*` instance methods.

If you have suggestions on how this might be improved, please let me know.

ocefpaf: I guess it would take a major refactor, or a better coding mind then mine, to suggest something better. This is fine. It is working an documented. +1 from me.

Issue #949, opened 2022-10-03
Improve CF Conformance


jamesdoyle21:
5 Coordinate Systems and Domain
Requirements:

 

- [ ] All of a variable’s dimensions that are latitude, longitude, vertical, or time dimensions must have corresponding coordinate variables.

- [ ]  A coordinate variable must have values that are strictly monotonic (increasing or decreasing).

 

- [ ] A coordinate variable must not have the _FillValue or missing_value attributes.

 

- [ ] The type of the coordinates attribute is a string whose value is a blank separated list of variable names. All specified variable names must exist in the file.

- [ ]  The dimensions of each auxiliary coordinate must be a subset of the dimensions of the variable they are attached to, with two exceptions. First, a label variable of type char will have a trailing dimension for the maximum string length. Second, a ragged array (Chapter 9, Discrete sampling geometries and Appendix H) uses special, more indirect, methods to connect the data and coordinates.

jenseva:
Thanks for your work on the CF checker! It is a very helpful tool that I use often. 

I am in the process of writing new files in the unusual format of CF 1.8 with the discrete geometry ragged array of trajectory profiles Appendix H.6.3. 

[https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#_ragged_array_representation_of_trajectory_profiles](https://cfconventions.org/Data/cf-conventions/cf-conventions-1.8/cf-conventions.html#_ragged_array_representation_of_trajectory_profiles)

I get an error from the checker on my test file although as far as I can tell from the CF 1.8 docs it is to spec. Here is a sample error:

§5 Coordinate Systems | dimensions for auxiliary coordinate variable time (profile) are not a subset of dimensions for variable salinity (obs)

Upon finding this issue here, I am hoping this is a known issue with the checker. Can I simply ignore this particular error when it comes to CF 1.8 Ragged array of trajectory profiles (H.6.3)?

benjwadams:
Sorry for the very late reply.  Probably yes.  We're going to work through getting the rest of ragged array dimensions working.

leilabbb:
All necessary changes and fixes have been included in the latest pull request to meet the checklist requirements (see above: all boxes are checked).

Recommend closing this issue after pull request #1198 is merged.

PR #1198, opened 2025-05-15
Implementation for CF1-8 Section 5: Coordinate Systems

leilabbb:
Related to issue #949 
 
Listing Requirement Implementation for CF1-8 Section 5 : 

New Implementations:

1. Checks that the `coordinates` attribute is a space-separated list of valid variable names, and all listed variables exist in the dataset. Function: check_coordinates_attribute_format()
	
2. Checks that a coordinate variable must not have the _FillValue or missing_value attributes. Function: check_invalid_coordinate_attr()

3. Checks that spatial/temporal dimensions (time, lat, lon, height)  used in geophysical variables have proper coordinate variables. Function: check_spatiotemporal_dims_have_coordinate_vars()

Existing Implementations:

4. Satisfying CF 1.8 Chapter 5, paragraph 3: Dimensions of auxiliary coordinates must be a subset of the variable they’re attached to (with two documented exceptions)
Function: check_aux_coordinates()  

5. Checks that data in coordinate variables is either monotonically  increasing or decreasing. Function: check_coordinate_variables_strict_monotonicity()

### DIFF
--- a/compliance_checker/cf/cf_1_6.py
+++ b/compliance_checker/cf/cf_1_6.py
@@ -2214,7 +2214,7 @@ class CF1_6Check(CFNCCheck):
 
         ret_val.append(check_coords_attrs_format.to_result())
 
-    return ret_val
+        return ret_val
 
     # IMPLEMENTATION Section 5 Coordinate Systems and Domain
     def check_spatiotemporal_dims_have_coordinate_vars(self, ds):
@@ -2264,7 +2264,7 @@ class CF1_6Check(CFNCCheck):
 
         ret_val.append(check_spatiotemporal_dims_coords.to_result())
 
-    return ret_val
+        return ret_val
 
     # IMPLEMENTATION Section 5 Coordinate Systems and Domain
     def check_invalid_coordinate_attr(self, ds):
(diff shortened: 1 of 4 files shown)


leilabbb:
pre-commit.ci autofix

leilabbb:
"pre-commit.ci autofix"

compliance_checker/cf/cf_1_6.py
```diff
@@ -2172,6 +2172,131 @@ def check_aux_coordinates(self, ds):
...
+            # Check that it is a proper string
+            check_coords_attrs_format.assert_true(
+                isinstance(coord_attr, str),
+                f"The 'coordinates' attribute of variable '{var_name}' must be a string.",
+            )
+
+            # Check for unexpected characters (e.g., commas or brackets)
+            if isinstance(coord_attr, str):
+                has_invalid_chars = any(
+                    char in coord_attr for char in [",", "[", "]", "(", ")"]
```
benjwadams, line 2206: Could use a `set` here to avoid needing to make multiple passes over the string.  Unlikely to have a significant performance impact for such a short string, so it's not necessary.

PR #1204, opened 2025-06-03
Make NaN a valid FillValue

Karinon:
We have data where `nan` is used as the FillValue and is marked as an error by the compliance checker.

### DIFF
--- a/compliance_checker/cf/cf_1_6.py
+++ b/compliance_checker/cf/cf_1_6.py
@@ -420,7 +420,10 @@ class CF1_6Check(CFNCCheck):
 
             if hasattr(variable, "_FillValue") and hasattr(variable, "missing_value"):
                 total = total + 1
-                if variable._FillValue != variable.missing_value:
+                if not (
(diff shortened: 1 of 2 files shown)


Karinon:
Just for my understanding regarding the failing integration test: Does it fail because neracoos is down?

ocefpaf:
pre-commit.ci autofix

ocefpaf (approved):
@benjwadams I'll leave to you for a second opinion before we merge this one. After that I can mint a bugfix release.

benjwadams:
Don't worry about the integration tests, links occasionally are unavailable or rotten and your PR had nothing to do with it.

It looks from the discussion that CF backed away from disallowing NaN as a _FillValue or missing_value, and there's nothing in the conventions document forbidding use of NaN.  So it think it's most sensible and least surprising to allow its use for the aforementioned attributes.

ocefpaf:
@benjwadams I'm minting a bugfix release. Is that OK or do you want to hold it and wait for a few more PRs before we do this?

PR #1205, opened 2025-06-04
add requirements to an existing functions for Section 7.5 Geometries

leilabbb:
modified in cf_1_8.py check_geometry() to:
- include comments to show where the requirement in the codes were applied.
- implement conformance 7.5 missing requirements:

## Requirement 16/20: When node_count is missing and >1 node, type must be point

Modified in test_cf.py test_coordinate_grid_mapping_geometry_ds_pass() to:
 - added few testing arguments to set up the mocked dataset(geometry_ds) that works with the check_geometry() function.
    -- geom_geometry_type,
    -- geom_node_coordinates,
    -- expected_msg

-  the @pytest.fixture geometry_ds()  and @pytest.mark.parametrize() were also modified to work with the above

### DIFF
--- a/compliance_checker/tests/test_cf.py
+++ b/compliance_checker/tests/test_cf.py
@@ -2838,70 +2838,130 @@ class TestCF1_8(BaseTestCase):
         dataset.createDimension("instance", 1)
         fake_data = dataset.createVariable("someData", "f8", ("instance", "time"))
         fake_data.geometry = "geometry"
-        x = dataset.createVariable("x", "f8", ("point_count",))
+        x = dataset.createVariable("x", "f8", ("point_count", "instance"))
         y = dataset.createVariable("y", "f8", ("point_count",))
+        z = dataset.createVariable("z", "f8", ("instance",))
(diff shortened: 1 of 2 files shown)


compliance_checker/tests/test_cf.py
```diff
@@ -2945,7 +3097,7 @@ def test_polygon_geometry(self):
         dataset.variables["interior_ring"][:] = flip_ring_bits
         results = self.cf.check_geometry(dataset)
         # There should be messages regarding improper polygon order
-        assert results[0].value[0] < results[0].value[1]
+        assert results[0].value[0] > results[0].value[1]
```
benjwadams, line 3100: `.value[0]` should be score and `value[1]` should be the out of possible points, so the former should not exceed the latter.

benjwadams (changes requested):

compliance_checker/tests/test_cf.py
leilabbb: a fix is applied to the new commit

PR #1208, opened 2025-06-06
3.5: Fix dealing with wrong type in flags

Karinon:
One of our users provided data which use strings instead of arrays as the datatype for flag values. This caused a crash of the compliance-checker. This PR aims to avoid the crash.

```
 	byte A4O_flag_bright(lat, lon) ; 
 		A4O_flag_bright:flag_meanings = "A4O_flag_bright" ; 
 		A4O_flag_bright:flag_values = "0U 1U" ; 
 		A4O_flag_bright:flag_masks = "1U" ; 
 		A4O_flag_bright:long_name = "Level_2 bright mask for TOA reflectance (at 490) > 0.3, that includes possible clouds and sea ice" ; 
```

### DIFF
--- a/compliance_checker/cf/cf_1_6.py
+++ b/compliance_checker/cf/cf_1_6.py
@@ -420,13 +420,7 @@ class CF1_6Check(CFNCCheck):
 
             if hasattr(variable, "_FillValue") and hasattr(variable, "missing_value"):
                 total = total + 1
-                if not (
-                    variable._FillValue == variable.missing_value
-                    or (
-                        np.isnan(variable._FillValue)
-                        and np.isnan(variable.missing_value)
-                    )
-                ):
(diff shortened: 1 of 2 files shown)


compliance_checker/cf/cf_1_6.py
```diff
@@ -420,13 +420,7 @@ def check_fill_value_equal_missing_value(self, ds):
...
             if hasattr(variable, "_FillValue") and hasattr(variable, "missing_value"):
                 total = total + 1
-                if not (
-                    variable._FillValue == variable.missing_value
-                    or (
-                        np.isnan(variable._FillValue)
-                        and np.isnan(variable.missing_value)
-                    )
-                ):
+                if variable._FillValue != variable.missing_value:
```
ocefpaf, line 423: Do you mean to rever this change?

Issue #1183, opened 2025-05-12
Update Ch. 4.4 text on reference time vs. UDUNIT


jamesdoyle21:
CF 1.11
[Issue #405](https://github.com/cf-convention/cf-conventions/issues/405): Update ch. 4.4 text on reference time vs. UDUNIT

Notes from Ben:
"In short, there are other forms of reference time besides "<time_units> since <reference_date_time>" where "since" isn't used.  This is now recommended to use "since", but previous regexes dealing with reference time might not work if they rely on "since" being there.

Not to mention that I think the `cftime` library relies on this assumption too."

Anything to add Ben?

PR #1213, opened 2025-06-25
Cf reference time since

benjwadams:
Implements #1183


benjwadams:
Oops, needs squash -- we'll do when we merge.

PR #1231, opened 2025-08-28
xrlint

abkfenris:
A very initial pass at implementing some ACDD  checks using XRLint.

I chatted with @ocefpaf about [XRLint](https://bcdev.github.io/xrlint/) and it sounded like there may have already been interest in using it in the compliance checker. 

So far I’ve implemented checks for the existence of all the highly/reccomended/suggested attributes defined in the existing acdd check suite, and included the descriptions as suggestions on failures.

I’ve also implemented the global 1.3 specific attribute checks, such as if datasets contain the correct convention, the dates are ISO format, the metadata link is http, and that there are no blanks in the ID.


ocefpaf:
This is awesome! There are many advantages for us to slowly move to an xarray data model vs raw `netCDF.Dataset`:

- lightweight installs: xarray datasets can have a variety of backends, netcdf files included, and many are non-compiled code;
- take advantage of xarray's optimizations: dask, etc;
- reduce the code complexity in the long run.

I believe that our main challenge will be CF. We can probably start with a subset of the latest CF version. IMO it is not worth implement all past versions as they'll be legacy and supported by previous cc code.

abkfenris:
Well, I did start with ACDD for something simpler to get started with. 

From a quick poke around the CF checks, while they're definitely more complex, but since they build upon each other, older versions are just smaller subsets of rules (with some updated for new values, similar to how I've done the attribute checks here), so I think that would make it relatively easy to support different versions by starting at the base and creating rule sets/plugin configs along the way.

abkfenris:
Unless there are any objections, I'm going to try to upstream at least ACDD and CF test suites to XRLint.

Issue #1230, opened 2025-08-25
Docs require some cleanups and upload is disabled


ocefpaf:
Some modules were removed, but the docs build is still trying to find them. We should regenerated them. The GHA also needs some permission tweaking to get them to upload again.

benjwadams:
Closing this as fixed.  If not fully addressed, please reopen with details on how to tweak permissions properly.

PR #1235, opened 2025-09-02
fix docs permission

ocefpaf:
Address #1230


ocefpaf:
I'm planning a docs review soon. There are many places we can improve, warnings to fix, etc. But let's merge this one as-is first b/c it addresses the broken publication and fix a removed module that is sill in the docs.

PR #1292, opened 2026-04-17
Fix check_cell_boundaries_interval: vectorize and support polygonal bounds

JanStreffing:
## Problem

`CF1_7Check.check_cell_boundaries_interval` (7.1 Cell Boundaries, Recommendation 1/2) has four bugs that interact badly on CF-compliant unstructured / polygonal output (FESOM2, ICON, MPAS, …):

1. **Per-element netCDF4 slicing in a Python loop** — `for ii in range(len(variable[:]))` with `variable[ii]` / `boundary_variable[ii][0]` per iteration. Each access is a separate `netCDF4.Variable.__getitem__` → compressed chunk decode. On a 126 858-node dataset this never finishes (`>1h`, then killed).
2. **`Result` created *inside* the inner loop** → `ret_val.append(Result(…))` per vertex → N `Result` objects per coord (confirmed via `faulthandler.dump_traceback`).
3. **Hard-coded `bv[ii][0]` / `bv[ii][1]` indexing** — CF §7.1 permits bounds shape `(N, nvertices)` for polygonal cells, but the check only ever inspects the first two vertices. On `(N, 2)` intervals with descending values the `abs`-based ordering proxy silently passes broken data; on `(N, nvertices>2)` the check is effectively a no-op.
4. **No shape guard.**

## Fix

Replace the loop with a vectorized bounding-box test:

```python
centers = np.ma.asanyarray(variable[:])
bnds    = np.ma.asanyarray(boundary_variable[:])
bmin    = bnds.min(axis=-1)
bmax    = bnds.max(axis=-1)
# (longitude-wrap unwrap for cells with span > 180°, see below)
tol     = atol + rtol * max(|bmin|, |bmax|)   # if either side is float
bad     = (centers < bmin - tol) | (centers > bmax + tol)
ret_val.append(Result(..., not bad.any(), ..., [f"{n_bad} point(s) ..."]))
```

- Works for interval bounds `(N, 2)` **and** polygonal/unstructured bounds `(N, nvertices)`. Bounding-box is weaker than point-in-polygon but catches gross violations and is cheap; the stronger test can be added later.
- Float tolerance (`atol=1e-8`, `rtol=1e-5`, mirrors `np.isclose`) filters out float32 vs float64 precision noise between coord and bounds variables.
- Handles descending bounds correctly via `min/max`.
- **Longitude-wrap aware**: for coords with `units=degrees_east` or `standard_name=longitude`, cells with raw vertex span > 180° are unwrapped relative to their own center (vertices more than ±180° from the center are shifted by ∓360°) before the bounding box is computed. Antimeridian-crossing cells no longer false-positive.
- One `Result` per coordinate variable.
- Reads each variable once via `np.ma.asanyarray(variable[:])` — no per-element netCDF I/O.

## Benchmark

126 858-node FESOM2 unstructured NetCDF with `lat_bnds(nod2, 18)` and `lon_bnds(nod2, 18)`:

| Version | Wall time | Findings |
|--------|-----------|----------|
| Before | `>1h` (killed) | — |
| After  | `~3s` | 1 genuine North-Pole lat mesh violation (real), 0 longitude false positives |

## Tests

- Synthetic interval bounds `(5, 2)`: good file → no flag; bad file (one center outside) → flagged. ✓
- Synthetic longitude wrap: center `+179.5°` inside a cell with vertices `[179°, -179°]` → no flag. ✓
- Synthetic longitude non-wrap: center `+20°` outside cell `[0°, 10°]` → flagged. ✓
- Real FESOM2 unstructured: 1 genuine mesh quirk flagged, 68 antimeridian false positives gone.

## Reviewer notes

- No test-suite additions here (existing tests all target `(N, 2)` and still pass). Happy to add `(N, nvertices>2)`, descending-bounds, float-precision and longitude-wrap fixtures if the maintainers want them in the same PR.
- The function is marked "Recommendations (1/2)" (`BaseCheck.MEDIUM`, not blocking) — semantic output surfaces as advisory.

### DIFF
--- a/compliance_checker/tests/test_cf.py
+++ b/compliance_checker/tests/test_cf.py
@@ -2029,7 +2029,7 @@ class TestCF1_7(BaseTestCase):
         should lie within, or on the boundary, of the cells specified by the
         associated boundary variable.
 
-        The check emits one Result per coordinate variable: pass iff every
+        The check emits one Result per coordinate variable: pass if every
         center lies within the bounding box of its cell's vertices.
         """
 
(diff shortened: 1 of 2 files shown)


ocefpaf:
@JanStreffing can you provide a test case for the new logic?

JanStreffing:
@ocefpaf happy to - just to make sure I cover what you're after: you mean pytest cases in `compliance_checker/tests/` that build tiny synthetic datasets in-test and assert on the `Result` from `CF1_7Check.check_cell_boundaries_interval` (covering `(N, 2)` intervals, `(N, nvert>2)` polygons, descending bounds, float32/float64 precision, longitude wrap)? Not a committed real-world `.nc`, given the existing fixtures are all ≲ 24 KB. If you are after the latter, I've hosted it here: 

https://raw.githubusercontent.com/JanStreffing/compliance-checker/issue-1293-sample/samples/sample_unstructured.nc (2.6 MB, FESOM2 CORE2 mesh, 126 858 nodes, lat_bnds/lon_bnds shape (nod2, 18))

ocefpaf:
>  you mean pytest cases in compliance_checker/tests/ that build tiny synthetic datasets in-test and assert on the Result from CF1_7Check.check_cell_boundaries_interval (covering (N, 2) intervals, (N, nvert>2) polygons, descending bounds, float32/float64 precision, longitude wrap)? 

That is preferable, but if that is too hard, we do have exceptions for the alternative small nc file.

JanStreffing:
@ocefpaf tests added in 610741e — six in-memory `MockTimeSeries` cases covering interval pass/fail, descending bounds, `(N, nvertices>2)` polygons, float32/float64 precision, and longitude wrap. Also updates the existing `test_check_cell_boundaries_interval` assertion from `(1, 2)` → `(0, 1)` to match the new one-Result-per-coord semantics (this was the CI failure).

ocefpaf (commented):
LGTM. Thanks! I would like to get @benjwadams eyes here before merging though b/c this is a part of the code that I did not write.

ocefpaf:
@JanStreffing do you mind testing this against the file in https://github.com/ioos/compliance-checker/issues/1296:

https://data.ceda.ac.uk/badc/cmip6/data/CMIP6/CMIP/MOHC/UKESM1-0-LL/historical/r1i1p1f2/Omon/tos/gn/v20190627/tos_Omon_UKESM1-0-LL_historical_r1i1p1f2_gn_185001-194912.nc

I'm getting an array broadcast error.

BTW, do you mind if we merge #1299 first, and mint a bugfix release for the "status quo" before merging this one? I can handle the rebasing to avoid taxing you with our decision, but I believe we need a minor bugfix release first and then a a minor release with this extended check.

compliance_checker/tests/test_cf.py
```diff
@@ -1988,37 +1988,162 @@ def test_check_cell_boundaries(self):
...
+
     def test_check_cell_boundaries_interval(self):
         """
         7.1 Cell Boundaries
         Recommendations: (1/2)
         The points specified by a coordinate or auxiliary coordinate variable
         should lie within, or on the boundary, of the cells specified by the
         associated boundary variable.
+
+        The check emits one Result per coordinate variable: pass iff every
```
ocefpaf:
```suggestion
        The check emits one Result per coordinate variable: pass if every
```

compliance_checker/tests/test_cf.py
```diff
@@ -2145,6 +2145,72 @@ def test_check_cell_boundaries_interval_longitude_wrap(self):
         assert (score, out_of) == (0, 1)
         assert any("1 point(s)" in m for m in messages)
 
+    def test_check_cell_boundaries_interval_2d_curvilinear(self):
+        """2-D curvilinear lat/lon (e.g. tripolar ocean): centers ``(J, I)``
+        with bounds ``(J, I, 4)``. Includes one antimeridian-crossing cell
+        so the longitude-wrap branch runs and the broadcast must work for
+        2-D shapes (regression test for the UKESM1-0-LL ``tos`` failure
+        reported on PR #1292)."""
+        # 3x4 curvilinear grid; cell [1, 0] spans the antimeridian.
```
ocefpaf: You don't have to do this if you are too busy -- we can do that later -- but I would prefer if these were separated tests, for example `test_check_cell_boundaries_interval_2d_curvilinear_3x4_curv(self)`, etc,. That would make future debugging and fixes a bit easier to implement.

JanStreffing:
@ocefpaf good catch. The broadcast error was the `[:, None]` reshape in the longitude-wrap branch. It works for 1-D coords (`(N,)` against bounds `(N, k)`), but on 2-D curvilinear lat/lon `(J, I)` it produces `(J, 1, I)` which won't broadcast against bounds `(J, I, k)`. Pushed `7fe7e3a` with:

- both `[:, None]` uses swapped to `[..., None]` so the reshape generalises to any number of leading axes
- tighter early shape guard: `bnds.ndim != centers.ndim + 1 or bnds.shape[:-1] != centers.shape`
- a 2-D regression test (3×4 grid with one antimeridian-crossing cell, the exact UKESM1-0-LL `tos` failure mode)
- your `iff` → `if` docstring suggestion

Verified on the UKESM file. No broadcast error. The §7.1 check now runs through and reports 85 real longitude points outside their cell bbox on the tripolar grid, which is a true CF finding rather than a check bug.

Happy with the merge plan. Please go ahead with #1299 first and rebase this on top, thanks for handling that.

ocefpaf:
@JanStreffing hopefully I did not mangled your PR when resolving the merge conflict. Let me know if you are OK with the results here before we merge it.

ocefpaf:
> @JanStreffing hopefully I did not mangled your PR when resolving the merge conflict. Let me know if you are OK with the results here before we merge it.

Spoke too soon! I'm never using GH interface ever again. Do you mind force pushing your previous change from before the merge conflict. I'll handle the fixes later for you.

JanStreffing:
Done, branch is back at `7f9e5c4`. All yours.

ocefpaf (approved):
@leilabbb and @benjwadams this PR looks great and generalizes the section 7.1 checks. We tested against some existing files and @JanStreffing made an awesome working of expanding the existing tests.

Please let me know what you think. After merging this one I would like to mint a v6.1.0 release to get this feature out.

benjwadams:
This looks good overall, but we need to add further checks for the edge cases where the bounds are not square-ish (pardon my topology nomenclature), e.g. not rectangular, cube-like, etc.  We have `shapely` as a library but it looks like most of the the supported operations are in the Cartesian plane.   SciPy and CGAL look like they have support for >= 3 dimensions, so I'm looking at both of them right now.

mo-gill:
In reference to this [comment](https://github.com/ioos/compliance-checker/issues/1296#issuecomment-4354663837) i have tested and seem to be getting the same expected results over here. Thanks for having a look at this!

benjwadams (changes requested):

benjwadams:
> This looks good overall, but we need to add further checks for the edge cases where the bounds are not square-ish (pardon my topology nomenclature), e.g. not rectangular, cube-like, etc. We have `shapely` as a library but it looks like most of the the supported operations are in the Cartesian plane. SciPy and CGAL look like they have support for >= 3 dimensions, so I'm looking at both of them right now.

I just saw that this is replacing the simple interval case after re-reviewing the code.  We'll probably make a separate issue n-dimensional bounds.  `shapely.intersection` could presently handle 2D polygon bounds accurately.

In an extreme and rather degenerate example, we could consider the bounds where n=4 and the bounds has no area (so it looks like a "line segment"), but the point `x` lies outside the bounds:

``` 
x /   
 /    
/     
```

Currently this would be accepted by the bounding box test, but obviously does not intersect the bounds.  This is a pretty extreme example, but could still occur with valid polygons with area wherein the points lie outside of the bounds.

benjwadams:
I'm going to merge and make follow up PRs for the `(N, M, 4)` and `(..., M, X)` vertices.

benjwadams (approved):

Issue #1298, opened 2026-04-27
Another possible bug in `check_cell_boundaries_interval` at 0E


mo-gill:
**Version of compliance checker running:*
1.7

**Describe the checker this affects:**
IOOS checker `check_cell_boundaries_interval`

**Attach a minimal CDL or NetCDF file which is able to reproduce the issue**
https://data.ceda.ac.uk/badc/cmip6/data/CMIP6/CMIP/MOHC/UKESM1-0-LL/piControl/r1i1p1f2/Amon/uas/gn/v20190410/uas_Amon_UKESM1-0-LL_piControl_r1i1p1f2_gn_196001-204912.nc

**To Reproduce:**
Run compliance-checker on the above file using `v1.7`,  whilst ensuring you have `netCDF4` version `1.7.4.` in the same environment.

**Describe the issue below:**
When checking cell bounds, the cell crossing 0E is being flagged as erroneous due to a misunderstanding around the bounds within the code. 

```
--------------------------------------------------------------------------------
                         IOOS Compliance Checker Report                         
                                 Version 5.4.2                                  
                     Report generated 2026-04-23T13:34:23Z                      
                                     cf:1.7                                     
http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html
--------------------------------------------------------------------------------
                               Corrective Actions                               
uas_Amon_UKESM1-0-LL_piControl_r1i1p1f2_gn_196001-204912.nc has 1 potential issue


                                    Warnings                                    
--------------------------------------------------------------------------------
§7.1 Cell Boundaries
* The points specified by the coordinate variable lon (0.0) lie outside the boundary of the cell specified by the associated boundary variable lon_bnds ([-0.9375  0.9375])
* The points specified by the coordinate variable lon (0.0) lie outside the boundary of the cell specified by the associated boundary variable lon_bnds ([-0.9375  0.9375])
* The points specified by the coordinate variable lon (0.0) lie outside the boundary of the cell specified by the associated boundary variable lon_bnds ([-0.9375  0.9375])
* The points specified by the coordinate variable lon (0.0) lie outside the boundary of the cell specified by the associated boundary variable lon_bnds ([-0.9375  0.9375])
* The points specified by the coordinate variable lon (0.0) lie outside the boundary of the cell specified by the associated boundary variable lon_bnds ([-0.9375  0.9375])
[...]

Issue #1296, opened 2026-04-21
Possible bug in `v1.7` `check_cell_boundaries_interval`


mo-gill:
**Version of compliance checker running:**
1.7

**Describe the checker this affects:**
IOOS checker `check_cell_boundaries_interval`

**To Reproduce:**
Check data with a three dimensional bounds variables.
e.g. [UKESM1 tos ocean data](https://data.ceda.ac.uk/badc/cmip6/data/CMIP6/CMIP/MOHC/UKESM1-0-LL/historical/r1i1p1f2/Omon/tos/gn/v20190627/tos_Omon_UKESM1-0-LL_historical_r1i1p1f2_gn_185001-194912.nc)

**Describe the issue below:**
It seems that this function only checks cell boundaries on first level due to indentation level being one too large on line [380](https://github.com/ioos/compliance-checker/blob/main/compliance_checker/cf/cf_1_7.py#L380). This means that only the first coordinate is checked.

If this issue is corrected manually, the following output is obtained:
```
-------------------------------------------------------------------------------
                         IOOS Compliance Checker Report                         
                                 Version 5.4.2                                  
                     Report generated 2026-04-21T13:13:34Z                      
                                     cf:1.7                                     
http://cfconventions.org/Data/cf-conventions/cf-conventions-1.7/cf-conventions.html
--------------------------------------------------------------------------------
                               Corrective Actions                               
tos_Omon_UKESM1-0-LL_historical_r1i1p1f2_gn_185001-194912.nc has 3 potential issues


                                    Warnings                                    
--------------------------------------------------------------------------------
§4.1 Latitude Coordinate
* latitude variable 'vertices_latitude' should define standard_name='latitude' or axis='Y'

§4.2 Longitude Coordinate
* longitude variable 'vertices_longitude' should define standard_name='longitude' or axis='X'

§7.1 Cell Boundaries
* The Boundary variables 'vertices_latitude' should not have the attributes: '['units', 'missing_value', '_FillValue']'
* The Boundary variables 'vertices_longitude' should not have the attributes: '['units', 'missing_value', '_FillValue']'
WARNING: The following exceptions occurred during the cf:1.7 checker (possibly indicate compliance checker issues):
cf:1.7.check_cell_boundaries_interval: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

[...]

PR #967, opened 2022-10-26
7.1 cell boundaries

leilabbb:
Related to issue #952

PR #1299, opened 2026-04-27
Debug  check_cell_boundaries_interval

ocefpaf:
Should help understand #1298 and #1296.

### DIFF
--- a/compliance_checker/cf/cf_1_7.py
+++ b/compliance_checker/cf/cf_1_7.py
@@ -364,9 +364,9 @@ class CF1_7Check(CF1_6Check):
                 if not min(boundary_variable[k]) <= variable[k] <= max(boundary_variable[k]):
                     valid = False
                     reasoning.append(
-                        f"The points specified by the coordinate variable {variable_name} ({variable[k]})"
-                        " lie outside the boundary of the cell specified by the "
-                        f"associated boundary variable {boundary_variable_name} ({boundary_variable[k]})",
+                        f"The points specified by the coordinate variable {variable_name} ({variable[k]}) "
+                        "lie outside the boundary of the cell specified by the "
+                        f"associated boundary variable {boundary_variable_name} ({boundary_variable[k]}).",
                     )
 
                 result = Result(
--- a/compliance_checker/cf/cf_1_7.py
+++ b/compliance_checker/cf/cf_1_7.py
@@ -379,7 +379,7 @@ class CF1_7Check(CF1_6Check):
                     reasoning,
                 )
                 ret_val.append(result)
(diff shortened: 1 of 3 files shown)


ocefpaf:
@leilabbb as the author of https://github.com/ioos/compliance-checker/pull/967, do you mind taking a look?

compliance_checker/cf/cf_1_7.py
```diff
@@ -360,15 +360,14 @@ def check_cell_boundaries_interval(self, ds):
             variable = ds.variables[variable_name]
             boundary_variable = ds.variables[boundary_variable_name]
 
-            for ii in range(len(variable[:])):
-                if abs(boundary_variable[ii][1]) >= abs(boundary_variable[ii][0]):
```
ocefpaf, line 364: I could not find if we should check the bounds themselves in https://cfconventions.org/Data/cf-conventions/cf-conventions-1.13/cf-conventions.html#cell-boundaries. I wonder if it is OK for them to be reversed in the spec.

ocefpaf:
Note that we only do 1D checks, but the spec is extended to 2D.

compliance_checker/cf/cf_1_7.py
```diff
@@ -360,15 +360,14 @@ def check_cell_boundaries_interval(self, ds):
...
-                    if not ((abs(variable[ii]) >= abs(boundary_variable[ii][0])) and (abs(variable[ii]) <= abs(boundary_variable[ii][1]))):
-                        valid = False
-                        reasoning.append(
-                            f"The points specified by the coordinate variable {variable_name} ({variable[ii]})"
-                            " lie outside the boundary of the cell specified by the "
-                            f"associated boundary variable {boundary_variable_name} ({boundary_variable[ii]})",
-                        )
+            for k in range(len(variable[:])):
+                if not min(boundary_variable[k]) <= variable[k] <= max(boundary_variable[k]):
```
ocefpaf, line 367: Should probably add a check to ensure that `boundary_variable` has only 2 values. While the spec supports 4, we did not implement that yet.

benjwadams (approved):

ocefpaf:
Just to avoid an accidental merge. We should focus on https://github.com/ioos/compliance-checker/pull/1292 first.

ocefpaf:
We will release a MICRO bugfix release that fixes the current 1-D coords for section 7.1. See the discussion in https://github.com/ioos/compliance-checker/pull/1292#issuecomment-4354189922 for the plan on adding a more general solution into a  MINOR release to avoid breaking people's expectation of the functionality.

PR #1303, opened 2026-05-11
Properly list available modules

ocefpaf:
Let's make it easier for downstream libraries to inspect what is available instead of skipping the import check.


compliance_checker/cf/cf.py
```diff
@@ -24,9 +24,36 @@
...
+    "grid_mapping_dict16",
+    "grid_mapping_dict17",
+    "horizontal_datum_names17",
+    "prime_meridian_names17",
+    "CF1_6Check",
+    "CF1_7Check",
+    "CF1_8Check",
+    "CF1_9Check",
+    "CF1_10Check",
+    "CF1_11Check",
```
ocefpaf: I wonder if we are using all these from the `cf.py`, if not, we could simplify the library a bit and clean those for the next release. It is a breaking change, but an easy one to fix, just an import origin change.

ocefpaf:
@benjwadams I first wanted to just properly list the imports in `cf.py`, but it turns out all of them were redundant with the `__init__.py`, or unused, or should be in the `__init__.py` anyway. Let me know what you think of this re-organization, the goal is to make it easier for downstream libraries.

PS: Note that while this is a simply change, it is a breaking one. I'd like to mint a new MINOR release after this is merged.

benjwadams (approved):

PR #1307, opened 2026-05-11
Various docs fixes and change web-app to WASM site

ocefpaf:
In this PR:

- Fix various URL redirects in the docs;
- Point to the new WASM web-app instead of the deprecated  web-app;
-  Update FAQ with the web-app versioning info;
- Build `cf_feature_parity_matrix` as proper docs so we can link to it.


ocefpaf:
The https://mmisw.org/ont/ioos/platform endpoint is timing out now 😩

ocefpaf:
Merging this... Hopefully that URL will be back next time we build the docs.

Issue #1304, opened 2026-05-11
Update CF Classifications to Match Test Classification Scheme


jamesdoyle21:
From May 11 Meeting notes:

"Data standard classifications:

- IOOS metadata profile(https://ioos.github.io/ioos-metadata/ioos-metadata-profile-v1-2.html): 

required, recommended, and optional.

- CF: Required, Highly-recommended, Suggested

- ACDD: Highly-recommended, Recommended, Suggested

- NCEI: Required, Recommended, Suggested

Example: “Conventions: Conventions must contain the string "IOOS 1.2" is required by the IOOS Metadata Profile, highly recommended for the web app, and high_priority for the package."

jamesdoyle21:
@mwengren @benjwadams

benjwadams:
May be eligible for closure -- Compliance Checker Web (NodeJS version) still has these headers, but we'll be moving to the WASM version very soon.

PR #1309, opened 2026-05-12
Changes CF error display headers

benjwadams:
Changes CF error display headers to the following
- 3: Required
- 2: Recommended
- 1: Suggested

### DIFF
--- a/compliance_checker/cf/cf_1_6.py
+++ b/compliance_checker/cf/cf_1_6.py
@@ -33,7 +33,7 @@ class CF1_6Check(CFNCCheck):
     _cc_spec_version = "1.6"
     _cc_description = "Climate and Forecast Conventions (CF)"
     _cc_url = "http://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html"
-    _cc_display_headers = {3: "Errors", 2: "Warnings", 1: "Info"}
+    _cc_display_headers = {3: "Required", 2: "Recommended", 1: "Suggested"}
     appendix_a = appendix_a_base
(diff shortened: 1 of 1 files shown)


benjwadams:
Implements #1304

benjwadams:
@ocefpaf, not directly related to this, but I had some trouble running the pre-commit tests for `add-trailing-comma`.  It kept trying to run under the system Python (3.9) and erroring out even though I had an activated virtualenv running Python 3.11 series.  It's possible I'm omitting something, so I don't want to make a separate issue quite yet.

ocefpaf:
> @ocefpaf, not directly related to this, but I had some trouble running the pre-commit tests for `add-trailing-comma`. It kept trying to run under the system Python (3.9) and erroring out even though I had an activated virtualenv running Python 3.11 series. It's possible I'm omitting something, so I don't want to make a separate issue quite yet.

Python 3.9 reached EoL and compliance-checker is >=3.10 now. You can skip the checks with the option `--no-verify` when commiting.

ocefpaf:
Regarding this PR, after Monday's meeting I found @ksauby original issue and the PR that fixed it: https://github.com/ioos/compliance-checker/pull/1262

If they are still experiexing the issue it is probably b/c the web-app at https://compliance.ioos.us/  was not updated yet.

Here we also change the nomenclature for CF, but that is not really specified in the CF documents and we had `Errors`, `Warnings`, and `Info` for so long that I'm not so sure we should change now.

TL;DR I don't think we should change it for CF here, unless they explicitly define a language for these in the docs.

PR #1320, opened 2026-06-10
Fix CF-1.9 check_domain_variables crashes on malformed domain variables and ragged arrays (#1318)

### LINKED ISSUE #1318: CF-1.9 check_domain_variables crashes on the malformed inputs it should flag (str.join TypeError, ragged unpack ValueError, VariableReferenceError.__str__ RecursionError) — silently no result under run_all
The CF-1.9 `check_domain_variables` check (section 5.8, inherited by CF-1.10 and CF-1.11) crashes with an unhandled exception on exactly the malformed inputs it is supposed to flag: a domain variable whose `dimensions` attribute names a missing dimension raises `TypeError`, and a ragged-array dataset with a 1-D numeric id variable raises `ValueError`. Because `suite.run_all` catches per-check exceptions (`suite.py:406-410`), the CF 5.8 domain check then silently contributes no result to the report — the user sees nothing wrong. There are three interlocking defects (a `__str__` infinite recursion in the error sentinel, the `str.join` TypeError, and a return-value/unpack mismatch); they were introduced together with the CF 5.8 domain-variable feature work and are most sensibly fixed together, so they are grouped in one issue.

The existing `test_domain` (`compliance_checker/tests/test_cf.py:3594`) only exercises a bad `coordinates` attribute, never a bad `dimensions` attribute and never the ragged branch, which is why none of this is caught by the test suite.

## Environment

- ioos/compliance-checker commit `b3fa4d490fa4badf46ab79b2a57b11d465ef754c` (editable install, reports `6.1.1.dev31+gb3fa4d490`)
- Windows 11 Pro
- Python 3.14.5
- netCDF4 1.7.4, numpy 2.4.6, cf-units 3.3.1, cftime 1.6.5

All reproductions below use in-memory `netCDF4.Dataset` objects — no `.nc`/`.cdl` files or `ncgen` needed. Save each snippet as `repro.py` and run `python repro.py` in an environment with compliance-checker installed. (Tracebacks below are verbatim from this machine with local path prefixes abbreviated to `...`.)

---

## Sub-defect 1 (root cause): `VariableReferenceError.__str__` infinite-recurses

`compliance_checker/cf/util.py:2225-2226`:

```python
    def __str__(self):
        return f"Cannot find variable named {self.name} in dataset {self}.dataset_path"
```

The f-string interpolates `{self}`, which re-invokes `__str__` — infinite recursion. The `.dataset_path` sits *outside* the brace, so it is a literal suffix; this was clearly meant to be `{self.dataset_path}`.

### Reproduction

```python
from compliance_checker.cf.util import VariableReferenceError

print(str(VariableReferenceError("foo")))
```

### Observed output

```
  File "repro.py", line 3, in <module>
    print(str(VariableReferenceError("foo")))
          ~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\compliance_checker\cf\util.py", line 2226, in __str__
    return f"Cannot find variable named {self.name} in dataset {self}.dataset_path"
                                                               ^^^^^^
  File "...\compliance_checker\cf\util.py", line 2226, in __str__
    return f"Cannot find variable named {self.name} in dataset {self}.dataset_path"
                                                               ^^^^^^
  File "...\compliance_checker\cf\util.py", line 2226, in __str__
    return f"Cannot find variable named {self.name} in dataset {self}.dataset_path"
                                                               ^^^^^^
  [Previous line repeated 996 more times]
RecursionError: maximum recursion depth exceeded
```

### Expected

```
Cannot find variable named foo in dataset None
```

### Why it matters

Any attempt to log, print, or interpolate this sentinel blows up. The shipped code does exactly that: `cf_1_9.py:149` interpolates the error object directly into an f-string (`f"Referenced coordinate variable {coord_var} does not exist in file"`), so once sub-defect 3 below is fixed, that ragged-branch path would hit this RecursionError. It also means the *obvious* one-line fix for sub-defect 2 (`", ".join(str(e) for e in dim_errors)`) would still crash — which is why these belong in one fix.

### Proposed fix

```diff
-        return f"Cannot find variable named {self.name} in dataset {self}.dataset_path"
+        return f"Cannot find variable named {self.name} in dataset {self.dataset_path}"
```

---

## Sub-defect 2: `check_domain_variables` raises `TypeError` when a domain variable's `dimensions` attribute names a missing dimension

`compliance_checker/cf/cf_1_9.py:189-190` joins a list of `VariableReferenceError` *objects*, but `str.join` requires `str` elements:

```python
                if dim_errors:
                    errors_str = ", ".join(dim_errors)
```

`dim_errors` comes from `reference_attr_variables(ds, dim_nc_attr, " ", "dimensions")` (`cf/util.py:2525-2531`), which appends `VariableReferenceError(var_name)` objects, not strings. The sibling *coordinates* branch nine lines later does it correctly (`cf_1_9.py:199`: `", ".join(err.name for err in domain_coord_var_errors)`).

### Reproduction

```python
from netCDF4 import Dataset

from compliance_checker.cf.cf_1_9 import CF1_9Check

ds = Dataset("d.nc", "w", diskless=True, persist=False, memory=1 << 20)
ds.createDimension("time", 3)
# Scalar domain variable whose `dimensions` attribute names a dimension
# that does not exist in the file -- exactly what the check should flag.
domain = ds.createVariable("domain", "c", ())
domain.long_name = "Domain variable"
domain.setncattr("dimensions", "nope")

check = CF1_9Check()
check.setup(ds)
results = check.check_domain_variables(ds)
print(results[0].msgs)
```

### Observed output

```
  File "repro.py", line 15, in <module>
    results = check.check_domain_variables(ds)
  File "...\compliance_checker\cf\cf_1_9.py", line 190, in check_domain_variables
    errors_str = ", ".join(dim_errors)
TypeError: sequence item 0: expected str instance, VariableReferenceError found
```

### Expected

A failing `Result` whose message lists the missing dimension, in the same style the coordinates branch already produces (cf. `test_cf.py:3616`):

```
Could not find the following dimensions referenced in dimensions attribute from domain variable domain: nope
```

This is the commonly-exercised non-ragged branch of the check, and the trigger is precisely the malformed dataset that CF 1.9 section 5.8 ("Domain variables") says must be flagged: per the CF conformance requirements for 5.8, the `dimensions` attribute of a domain variable is a blank-separated list of dimension names *that must exist in the file*. Under `run_all` the exception is swallowed at `suite.py:406-410` and the check silently produces no result.

### Proposed fix

Mirror the coordinates branch:

```diff
-                    errors_str = ", ".join(dim_errors)
+                    errors_str = ", ".join(err.name for err in dim_errors)
```

---

## Sub-defect 3: `resolve_ragged_array_dimension` returns a bare string but its caller unpacks two values

`compliance_checker/cf/util.py:1168-1182` ends with `return ragged_type` (a single string, `"sample_dimension"` or `"instance_dimension"`), but its only caller, `cf_1_9.py:132`, unpacks two values:

```python
            ragged_array_dim_variable, ragged_attr_name = cfutil.resolve_ragged_array_dimension(ds)
```

Unpacking a 16/18-character string into two targets raises `ValueError`. Additionally, the guard at `util.py:1180` (`if ragged_variable is None:`) is dead code — `get_variables_by_attributes` returns a list (`[]` when empty), never `None` — so the intended "Could not find a ragged array related variable" `ValueError` can never fire.

### Reproduction

```python
import numpy as np
from netCDF4 import Dataset

from compliance_checker.cf.cf_1_9 import CF1_9Check

# Valid contiguous ragged-array timeSeries: 1-D numeric id variable with
# cf_role, count variable carrying sample_dimension.
ds = Dataset("ragged.nc", "w", diskless=True, persist=False, memory=1 << 20)
ds.featureType = "timeSeries"
ds.createDimension("station", 2)
ds.createDimension("obs", 5)

station = ds.createVariable("station", "i4", ("station",))
station.cf_role = "timeseries_id"

row_size = ds.createVariable("row_size", "i4", ("station",))
row_size.sample_dimension = "obs"
row_size.long_name = "number of obs per station"

temp = ds.createVariable("temperature", "f4", ("obs",))
temp.standard_name = "air_temperature"
temp.units = "K"
temp.coordinates = "time lat lon"
temp[:] = np.arange(5)

check = CF1_9Check()
check.setup(ds)
results = check.check_domain_variables(ds)
print(results)
```

### Observed output

```
  File "repro.py", line 28, in <module>
    results = check.check_domain_variables(ds)
  File "...\compliance_checker\cf\cf_1_9.py", line 132, in check_domain_variables
    ragged_array_dim_variable, ragged_attr_name = cfutil.resolve_ragged_array_dimension(ds)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: too many values to unpack (expected 2)
```

### Expected

The check runs and returns `Result` objects for the ragged-array domain checks.

### Reachability caveat (honest scoping)

This branch is gated by `is_dataset_valid_ragged_array_repr_featureType` (`util.py:1081`), which returns `False` when the `cf_role` id variable is the conventional 2-D `char(station, name_strlen)` string array (its `len(dimensions) != 1`). That is why the project's own `indexed_ragged_domain` CDL fixture does not trip it. The crash is reached when the id variable is 1-D — e.g. a numeric station id as above, or a 1-D string (vlen) id. So this is a real but less-frequently-exercised path than sub-defect 2.

### Proposed fix

```diff
-    if ragged_variable is None:
+    if not ragged_variable:
         raise ValueError("Could not find a ragged array related variable")
-    return ragged_type
+    return ragged_variable[0], ragged_type
```

(Note that after this fix, the `coord_var` f-string interpolation at `cf_1_9.py:149` becomes reachable, which is where sub-defect 1's `__str__` fix matters.)

---

## Test coverage

No existing test needs *updating* — `test_domain` (`test_cf.py:3594`) simply never exercises these paths (it covers a missing `long_name`, bad `coordinates`, and bad `cell_measures`, all with an existing `time` dimension in the `dimensions` attribute). Suggested additions alongside the fix:

- a domain variable with `dimensions = "nope"` (missing dimension) asserting the failure message rather than a crash;
- a ragged timeSeries dataset with a 1-D numeric `cf_role` id variable;
- a direct `str(VariableReferenceError("foo"))` assertion.

## Related

This code was introduced with the CF 5.8 domain-variable feature work (#949, follow-up in #951); neither PR nor any open/closed issue reports these crashes (searched issue tracker for `check_domain_variables`, `VariableReferenceError`, `resolve_ragged_array_dimension`, RecursionError/TypeError/ValueError keywords).

Happy to submit a PR for this if useful.

consigcody94:
# Fix CF-1.9 `check_domain_variables` crashes on malformed domain variables and ragged arrays

## Summary

The CF-1.9 (and inherited 1.10/1.11) `check_domain_variables` check crashes with an unhandled exception instead of reporting a failed `Result` in two situations, and the exception class it relies on cannot even be stringified. This PR fixes all three interlocking defects so the check degrades into failed results/messages the way it was designed to. Under `run_all` the crashes are swallowed, so today the check silently produces no result for exactly the malformed inputs it exists to flag.

Fixes #1318

## Root cause

`VariableReferenceError.__str__` interpolates `{self}` instead of `{self.dataset_path}`, recursing into itself (`RecursionError`); the domain `dimensions`-attribute branch joins the `VariableReferenceError` objects themselves (`", ".join(dim_errors)` → `TypeError`); and `resolve_ragged_array_dimension` returns only the attribute-name string while its sole caller unpacks `(variable, attribute_name)` (`ValueError: too many values to unpack`).

## What the fix changes

- `compliance_checker/cf/util.py` — `VariableReferenceError.__str__`: `{self}.dataset_path` → `{self.dataset_path}` (the `.dataset_path` was a literal outside the brace).
- `compliance_checker/cf/cf_1_9.py` — `", ".join(dim_errors)` → `", ".join(err.name for err in dim_errors)`, matching the sibling coordinates branch a few lines below.
- `compliance_checker/cf/util.py` — `resolve_ragged_array_dimension` now returns `(ragged_variable[0], ragged_type)`, exactly what `check_domain_variables` unpacks. The dead `if ragged_variable is None` guard (`get_variables_by_attributes` returns a list, never `None`) is replaced with an emptiness check so the intended `ValueError("Could not find a ragged array related variable")` can actually fire.

Three regression tests added in `compliance_checker/tests/test_cf.py`:

- `TestCF1_9::test_domain_invalid_dimensions` — a domain variable whose `dimensions` attribute names a missing dimension produces a failed `Result` with the expected message (previously `TypeError`).
- `TestCFUtil::test_resolve_ragged_array_dimension` — contiguous and indexed ragged datasets resolve to `(variable, attribute name)` (previously `ValueError` on unpack), and a dataset without ragged-array variables raises the intended `ValueError`.
- `TestCFUtil::test_variable_reference_error_str` — `str(VariableReferenceError(...))` returns the expected string (previously `RecursionError`).

Note: with this fix the ragged-array branch gets past `resolve_ragged_array_dimension`, but `check_domain_variables` line ~145 still iterates the `(references, errors)` tuple returned by `reference_attr_variables(ds, dim_name, " ", "dimensions")` rather than the references list — a separate latent defect left out of this PR to keep it minimal.

## Test evidence

Regression tests with the source fix temporarily reverted (`git stash` of `cf/util.py` + `cf/cf_1_9.py`, tests kept):

```
compliance_checker/tests/test_cf.py::TestCF1_9::test_domain_invalid_dimensions FAILED
    compliance_checker\cf\cf_1_9.py:190: TypeError: sequence item 0: expected str instance, VariableReferenceError found
compliance_checker/tests/test_cf.py::TestCFUtil::test_resolve_ragged_array_dimension FAILED
    ValueError: too many values to unpack (expected 2)
compliance_checker/tests/test_cf.py::TestCFUtil::test_variable_reference_error_str FAILED
    compliance_checker\cf\util.py:2226: RecursionError: maximum recursion depth exceeded
3 failed in 4.44s
```

With the fix applied:

```
compliance_checker/tests/test_cf.py::TestCF1_9::test_domain_invalid_dimensions PASSED
compliance_checker/tests/test_cf.py::TestCFUtil::test_resolve_ragged_array_dimension PASSED
compliance_checker/tests/test_cf.py::TestCFUtil::test_variable_reference_error_str PASSED
3 passed in 1.33s
```

Honest caveat on local suite coverage: the development box has no `ncgen` (netCDF C tools), so the `.cdl`-fixture-based tests (most of `test_cf.py`, `test_acdd.py`, etc.) cannot run locally — including the fixture portion of the existing `test_domain`. The new tests use only in-memory `MockNetCDF`/`MockTimeSeries`/`MockRaggedArrayRepr` datasets and need no fixtures. The fixture-free subset (`test_base`, `test_util`, `test_protocols`, `test_suite`) was run on this branch and on `main` as a baseline: identical results — `7 failed, 18 passed, 8 skipped, 1 xfailed`, where all 7 failures are the pre-existing `ncgen` `FileNotFoundError` environment issue on both runs. Full fixture suite should be exercised by CI.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/compliance_checker/cf/cf_1_9.py
+++ b/compliance_checker/cf/cf_1_9.py
@@ -187,7 +187,7 @@ class CF1_9Check(CF1_8Check):
                     "dimensions",
                 )
                 if dim_errors:
-                    errors_str = ", ".join(dim_errors)
+                    errors_str = ", ".join(err.name for err in dim_errors)
                     domain_valid.messages.append(
                         f"Could not find the following dimensions referenced in dimensions attribute from domain variable {domain_var.name}: {errors_str}",
                     )
--- a/compliance_checker/cf/util.py
+++ b/compliance_checker/cf/util.py
@@ -1177,9 +1177,9 @@ def resolve_ragged_array_dimension(ds: Dataset):
             instance_dimension=lambda s: isinstance(s, str),
         )
         ragged_type = "instance_dimension"
-    if ragged_variable is None:
+    if not ragged_variable:
         raise ValueError("Could not find a ragged array related variable")
-    return ragged_type
+    return ragged_variable[0], ragged_type
 
 
 def is_variable_valid_ragged_array_repr_featureType(nc, variable: str) -> bool:
@@ -2223,7 +2223,7 @@ class VariableReferenceError(Exception):
         self.dataset_path = dataset.filepath() if dataset is not None else None
 
     def __str__(self):
-        return f"Cannot find variable named {self.name} in dataset {self}.dataset_path"
+        return f"Cannot find variable named {self.name} in dataset {self.dataset_path}"
 
 
 class StandardNameTable:
--- a/compliance_checker/tests/test_cf.py
+++ b/compliance_checker/tests/test_cf.py
@@ -3647,6 +3647,19 @@ class TestCF1_9(BaseTestCase):
         results = self.cf.check_domain_variables(dataset)
         assert results[0].value[0] == results[0].value[1]
 
+    def test_domain_invalid_dimensions(self):
+        # TEST CONFORMANCE 5.8 REQUIRED 2/4
+        # a domain variable with a dimensions attribute referencing a
+        # nonexistent dimension should fail the check, not crash it
+        dataset = MockTimeSeries()
+        domain_var = dataset.createVariable("domain", "c", ())
+        domain_var.long_name = "Domain variable"
+        domain_var.coordinates = "lon lat depth"
+        domain_var.setncattr("dimensions", "xyxz")
+        results = self.cf.check_domain_variables(dataset)
+        assert results[0].value[0] != results[0].value[1]
+        assert "Could not find the following dimensions referenced in dimensions attribute from domain variable domain: xyxz" in results[0].msgs
+
 
 class TestCF1_11(BaseTestCase):
     def setup_method(self):
@@ -4114,3 +4127,27 @@ class TestCFUtil(BaseTestCase):
                 "trajectoryprofile",
             )
         )
+
+    def test_resolve_ragged_array_dimension(self):
+        # must return both the variable bearing the ragged array
+        # attribute and the attribute name, as unpacked by
+        # CF1_9Check.check_domain_variables
+        nc = MockRaggedArrayRepr("timeseries", "contiguous")
+        ragged_variable, ragged_type = cfutil.resolve_ragged_array_dimension(nc)
+        assert ragged_variable.name == "counter_var"
+        assert ragged_type == "sample_dimension"
+
+        nc = MockRaggedArrayRepr("timeseries", "indexed")
+        ragged_variable, ragged_type = cfutil.resolve_ragged_array_dimension(nc)
+        assert ragged_variable.name == "index_var"
+        assert ragged_type == "instance_dimension"
+
+        # no ragged array related variables present
+        nc = MockTimeSeries()
+        with pytest.raises(ValueError):
+            cfutil.resolve_ragged_array_dimension(nc)
+
+    def test_variable_reference_error_str(self):
+        # str() must not recurse into __str__ via {self}
+        err = cfutil.VariableReferenceError("foo")
+        assert str(err) == "Cannot find variable named foo in dataset None"


benjwadams (approved):

ocefpaf:
@benjwadams this is a bot account that will be able to run CIs and access to tokens on GitHub Actions after the first PR is merged. I would be more careful with such contributions given that most, if not all, supply chain attacks of 2026 happened due to holes in GitHub Actions.

With that said, we do run zizmor here, and does provide a layer of protection. Still, I would only merge PRs that have a human behind it.

benjwadams:
I've looked at the changed files and the code itself introduces no changed or added dependencies.

This fixes at least one issue with some code I wrote where there evidently was a crash in some cases.

In PRs I've seen recently from outside contributors I need to explicitly approve workflows to run.

ocefpaf:
> In PRs I've seen recently from outside contributors I need to explicitly approve workflows to run.

The first one, next PRs are automatically approved.

benjwadams:
Every PR I've seen recently from a non-core maintainer has had "Approve workflows to run" in the PR gating the GHA runs:

https://docs.github.com/en/actions/how-tos/manage-workflow-runs/approve-runs-from-forks

lukecampbell:
I'm with @ocefpaf on this one. The account looks entirely fictitious and it's hard to understand the rationale behind the activity. If it's a GENUINELY useful feature there's nothing preventing a human contributor from making a similar PR. There's nothing outwardly nefarious about the change but the flurry of activity seems extremely suspicious. One theory I could see is that the account is testing NOAA/IOOS projects' CI configuration by making innocuous PRs to scan for vulnerabilities in our configuration or security policies.

srstsavage:
It's an org level setting with per-repo override (settings/actions):

image

benjwadams:
cc @mwengren

PR #1321, opened 2026-06-10
Fix is_netcdf matching any path ending in the letters "nc" (#1319)

### LINKED ISSUE #1319: is_netcdf() matches any filename ending in the letters "nc" (myfunc, data.sync) — endswith("nc") missing the dot
`is_netcdf()` in `compliance_checker/protocols/netcdf.py` uses `url.endswith("nc")` — missing the dot — so any path whose name merely ends in the two letters "nc" is classified as a netCDF file: `is_netcdf('myfunc')`, `is_netcdf('data.sync')`, and `is_netcdf('stats.inc')` all return `True` without the file even existing.

## Environment

- ioos/compliance-checker commit `b3fa4d490fa4badf46ab79b2a57b11d465ef754c` (editable install, reports `6.1.1.dev31+gb3fa4d490`)
- Windows 11 Pro
- Python 3.14.5
- netCDF4 1.7.4, numpy 2.4.6

## Reproduction

Save as `repro.py` and run `python repro.py` in an environment with compliance-checker installed:

```python
from compliance_checker.protocols.netcdf import is_netcdf

for name in ("myfunc", "data.sync", "stats.inc", "foo.nc"):
    print(f"is_netcdf({name!r}) = {is_netcdf(name)}")
```

## Observed output

```
is_netcdf('myfunc') = True
is_netcdf('data.sync') = True
is_netcdf('stats.inc') = True
is_netcdf('foo.nc') = True
```

## Expected

Only the `.nc` *extension* should short-circuit to `True`; everything else should fall through to the magic-number sniffing below it (which correctly identifies real netCDF-classic/HDF5 content regardless of filename):

```
is_netcdf('myfunc') = False
is_netcdf('data.sync') = False
is_netcdf('stats.inc') = False
is_netcdf('foo.nc') = True
```

## Root cause

`compliance_checker/protocols/netcdf.py:23-25`:

```python
    # If it's a known extension, give it a shot
    if url.endswith("nc"):
        return True
```

The comment says "known extension", but the check matches a bare two-letter suffix, not the `.nc` extension.

## Impact

Via `CheckSuite.load_local_dataset` (`suite.py:853-858`), a misclassified file is handed straight to `netCDF4.Dataset(ds_str)` and fails with an opaque low-level `OSError` instead of reaching the `GenericFile` fallback two lines below. Low severity — it only bites users whose input filenames happen to end in "nc" — but the one-character fix removes a confusing failure mode and makes the extension short-circuit match its own comment.

## Proposed fix

```diff
-    if url.endswith("nc"):
+    if url.endswith(".nc"):
         return True
```

No existing tests encode the old behavior; a small parametrized test covering `"myfunc"`/`"data.sync"` vs `"foo.nc"` could accompany the fix.

Happy to submit a PR for this if useful.

consigcody94:
# Fix `is_netcdf` matching any path ending in the letters "nc"

## Summary

`protocols/netcdf.py::is_netcdf` uses `url.endswith("nc")` as its extension shortcut, so any path whose name merely ends in the two letters "nc" — `myfunc`, `data.sync`, `stats.inc` — is classified as a netCDF file. Via `suite.load_local_dataset` such files are handed straight to `netCDF4.Dataset`, which fails with an opaque `OSError` instead of letting them fall through to the magic-number content checks and the `GenericFile` fallback. This PR requires the full `.nc` extension.

Fixes #1319

## Root cause

The extension check is missing the dot: `endswith("nc")` matches a 2-letter suffix, not a file extension.

## What the fix changes

- `compliance_checker/protocols/netcdf.py`: `url.endswith("nc")` → `url.endswith(".nc")`.

The function has no other extension cases to update: `.nc4` (and anything else) never matched the old 2-letter check either — those paths are unaffected and still fall through to the `is_classic_netcdf`/`is_hdf5` magic-number sniffing, which keeps working for real netCDF files regardless of name.

Regression test added in `compliance_checker/tests/test_protocols.py` (parametrized, in the style of `test_as_zarr`): `myfunc`, `data.sync`, `stats.inc` → `False`; `foo.nc` → `True`. No real files are needed — the non-`.nc` names fall through to `open()`, whose `OSError`/`FileNotFoundError` path correctly returns `False`.

## Test evidence

With the fix temporarily reverted (`git stash` of `protocols/netcdf.py`, test kept):

```
FAILED compliance_checker/tests/test_protocols.py::test_is_netcdf_extension[myfunc-False]
FAILED compliance_checker/tests/test_protocols.py::test_is_netcdf_extension[data.sync-False]
FAILED compliance_checker/tests/test_protocols.py::test_is_netcdf_extension[stats.inc-False]
3 failed, 1 passed, 12 deselected in 3.45s
```

With the fix applied:

```
compliance_checker/tests/test_protocols.py::test_is_netcdf_extension[myfunc-False] PASSED
compliance_checker/tests/test_protocols.py::test_is_netcdf_extension[data.sync-False] PASSED
compliance_checker/tests/test_protocols.py::test_is_netcdf_extension[stats.inc-False] PASSED
compliance_checker/tests/test_protocols.py::test_is_netcdf_extension[foo.nc-True] PASSED
4 passed, 12 deselected in 0.41s
```

Suite check: the development box has no `ncgen` (netCDF C tools), so the `.cdl`-fixture-based tests cannot run locally. The fixture-free subset (`test_base`, `test_util`, `test_protocols`, `test_suite`) was run on this branch vs. a `main` baseline: `7 failed, 22 passed, 8 skipped, 1 xfailed` vs. baseline `7 failed, 18 passed, 8 skipped, 1 xfailed` — the +4 passes are the new parametrized cases; all 7 failures are the same pre-existing `ncgen` `FileNotFoundError` environment issue on both runs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/compliance_checker/protocols/netcdf.py
+++ b/compliance_checker/protocols/netcdf.py
@@ -21,7 +21,7 @@ def is_netcdf(url):
         return False
 
     # If it's a known extension, give it a shot
-    if url.endswith("nc"):
+    if url.endswith(".nc"):
         return True
 
     try:
--- a/compliance_checker/tests/test_protocols.py
+++ b/compliance_checker/tests/test_protocols.py
@@ -8,7 +8,7 @@ import platform
 
 import pytest
 
-from compliance_checker.protocols import zarr
+from compliance_checker.protocols import netcdf, zarr
 from compliance_checker.suite import CheckSuite
 
 from .conftest import datadir
@@ -19,6 +19,24 @@ pytestmark = [pytest.mark.integration]
 @pytest.fixture
 def cs():
     return CheckSuite()
+
+
+extension_io = [
+    ("myfunc", False),
+    ("data.sync", False),
+    ("stats.inc", False),
+    ("foo.nc", True),
+]
+
+
+@pytest.mark.parametrize("url_in, expected", extension_io)
+def test_is_netcdf_extension(url_in, expected):
+    """
+    Test that only paths with a proper .nc extension are assumed to be netCDF
+    files; other paths ending in the letters "nc" must fall through to
+    inspecting the file contents.
+    """
+    assert netcdf.is_netcdf(url_in) == expected
 
 
 @pytest.mark.vcr()


benjwadams:
This looks good overall.  ERDDAP does have some nonstandard netCDF file extensions such as .ncCF and .ncCFMA, but those are captured by separate logic.

PR #1323, opened 2026-06-17
cf §7.2: support multiple cell_measures entries per CF 1.11

JanStreffing:
The existing `_cell_measures_core` regex in `cf/cf_1_6.py` matched the entire attribute as a single `measure: var` pair via `regex.match`, so any file with two measures was flagged as formatted incorrectly even though CF 1.11 §7.2 explicitly allows multiple:

```
cell_measures = 'area: cell_area volume: cell_volume'
```

CMIP7's ocean branded variables (`hfx`, `hfy`, `masscello`, `sfx`, `sfy`, etc.) ship the registered `cell_measures = 'area: areacello volume: volcello'` verbatim, and every one of those files currently fails §7.2 even though the value is canonical.

This PR refactors `_cell_measures_core` to use `regex.finditer` over `measure: var` pairs, walk each pair, and reject if the regex doesn't cover the full attribute (so malformed inputs like `area:` with no var or random trailing junk still fail). It also enforces the §7.2 rule that each measure type may appear at most once (a second `area:` entry is a failure).

Verified locally on a CMIP7 file with `cell_measures = 'area: areacello volume: volcello'`. The §7.2 false-positive is gone. The only remaining failure is the downstream "var not in dataset" check, which is correctly handled by `CF1_7Check`'s `external_variables` path.

### DIFF
--- a/compliance_checker/cf/cf_1_6.py
+++ b/compliance_checker/cf/cf_1_6.py
@@ -2796,79 +2796,95 @@ class CF1_6Check(CFNCCheck):
 
     def _cell_measures_core(self, ds, var, external_set, variable_template):
         # IMPLEMENTATION CONFORMANCE REQUIRED 1/2
+        # CF 1.11 §7.2: cell_measures may carry MULTIPLE measure entries
+        # ("area: a volume: v"), and each entry must be of a different type.
+        # Walk every "type: var" pair instead of trying to match the whole
+        # string as a single pair.
         reasoning = []
-        search_str = r"^(?P<measure_type>area|volume):\s+(?P<cell_measure_var_name>\w+)$"
-        search_res = regex.match(search_str, var.cell_measures)
-        if not search_res:
+        pair_re = r"(?P<measure_type>area|volume):\s+(?P<cell_measure_var_name>\w+)"
+        matches = list(regex.finditer(pair_re, var.cell_measures))
+        # Reject if the parsed pairs don't account for the whole attribute
+        # (catches malformed entries like ``area:`` with no variable or
+        # ``area: foo bogus volume: bar``). Compare on whitespace-normalised
+        # forms so the check accepts any consistent spacing variant.
+        canonical_from_pairs = " ".join(f"{m.group('measure_type')}: {m.group('cell_measure_var_name')}" for m in matches)
+        canonical_input = " ".join(var.cell_measures.split())
+        if not matches or canonical_from_pairs != canonical_input:
             valid = False
             reasoning.append(
                 f"The cell_measures attribute for variable {var.name} "
                 "is formatted incorrectly. It should take the "
-                "form of either 'area: cell_var' or "
-                "'volume: cell_var' where cell_var is an existing name of "
-                "a variable describing the cell measures.",
+                "form of 'area: cell_var', 'volume: cell_var', or "
+                "multiple such pairs separated by spaces (e.g. "
+                "'area: a volume: v'), where each cell_var is an "
+                "existing variable describing the cell measures.",
             )
-        else:
-            valid = True
-            cell_measure_var_name = search_res.group("cell_measure_var_name")
+            return Result(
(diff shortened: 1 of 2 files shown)


JanStreffing:
Failing doc build seems to be IOOS CI infrastructure flake, right?

ocefpaf:
Looks like link check is failing to access this site due to ssl certs:

```
(generated/compliance_checker.cf.CF1_10Check: line    8) broken    https://docs.unidata.ucar.edu/netcdf-c/current/attribute_conventions.html - HTTPSConnectionPool(host='docs.unidata.ucar.edu', port=443): Max retries exceeded with url: /netcdf-c/current/attribute_conventions.html (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1082)')))
```

We can ignore that for now.

JanStreffing:
@ocefpaf both this PR and #1324 are small mechanical fixes for CMIP7 file checks (§7.2 cell_measures multi-pair here, §7.3.4 cell_methods names without coord vars on #1324). CI is green for both apart from the link-check SSL issue you already flagged here. Could you queue them for a review pass when you have a moment? Karl Taylor pointed at the §7.3.4 spec text in WCRP-CMIP/WCRP-universe#194 if you want context on #1324.

ocefpaf (approved):
LGTM. Thanks!

@benjwadams if you have some spare cycles to take a look at this one let me know. Otherwise I'll merge in ~1 week.

compliance_checker/cf/cf_1_6.py
```diff
@@ -2796,79 +2796,95 @@ def check_cell_boundaries(self, ds):
 
     def _cell_measures_core(self, ds, var, external_set, variable_template):
         # IMPLEMENTATION CONFORMANCE REQUIRED 1/2
+        # CF 1.11 §7.2: cell_measures may carry MULTIPLE measure entries
```
benjwadams: We don't really want to mix and match commentary from separate CF versions if possible and it looks like this behavior goes way back to early CF versions as well:
[CF 1.6 docs on cell measures](https://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#cell-measures) mentions:

> This is a string attribute comprising a list of blank-separated pairs of words of the form "measure: name". For the moment, "area" and "volume" are the only defined measures, but others may be supported in future.

I'd update to mention a pertinent section of CF 1.6 in the comment, or eliminate the reference to CF 1.11 in the comment entirely.

benjwadams (changes requested):

JanStreffing:
The CF 1.11 reference appeared four times in the diff, and one of them was a bigger problem than the version number.

https://github.com/ioos/compliance-checker/blob/a4ecd49c6c5116e7664ba37005b33717a34b7c82/compliance_checker/cf/cf_1_6.py#L2833-L2834

That was punctuated as a quotation from the convention, and the sentence is not in it. Searching the full CF 1.6 and CF 1.11 documents, "Multiple measures may be specified" occurs zero times in both, as does "of different types". The §7.2 paragraph is near identical across the two versions and says only:

> This is a string attribute comprising a list of blank-separated pairs of words of the form "measure: name". For the moment, "area" and "volume" are the only defined measures, but others may be supported in future.

The multiple-pairs behaviour is therefore CF 1.6, as you said, and the comment now cites that sentence. The one-entry-per-type rule has no basis in the convention text, so it is now labelled as this checker's reading rather than a quotation.

I kept the rule. Before this PR the whole-attribute match rejected `area: a area: b`, and a pair walk without the type check would start accepting it, which is a larger behaviour change than the one under review. If you would rather the PR only implemented what §7.2 states outright, the `seen_types` block and its test come out in one commit.

All four comments are updated in becb3ff. Comments only, no behaviour change, and the two cell_measures tests pass.

benjwadams (approved):

PR #1325, opened 2026-06-22
Fix linkcheck docs

ocefpaf:
Testing a non-conda packages solution to use PyPI's certifi instead of the system one.


ocefpaf:
Failures here should go away when https://github.com/Unidata/netcdf4-python/pull/1469 lands.

Issue #1350, opened 2026-08-25
Dealing with datatype int64


mphemming:
**Version of compliance checker running:**

Version: 5.4.2

**Describe the checker this affects:**

check_data_types

**Attach a minimal CDL or NetCDF file which is able to reproduce the issue**

[example.nc.txt](https://github.com/user-attachments/files/31399339/example.nc.txt)

[example.zip](https://github.com/user-attachments/files/31445743/example.zip)

**To Reproduce:**

compliance-checker -t=cf:1.6 example.nc

**Describe the issue below:**

We have been skipping `check_data_types` for files that include int64 variables. These files need to include int64 variables as the values can be very large.  It would be ideal not to skip this check as we would like to check other variables in the file. 

>                                      Errors                                     
> --------------------------------------------------------------------------------
> §2.2 Data Types
> * The variable QC_accept_code failed because the datatype is int64
> * The variable QC_reject_code failed because the datatype is int64

Is there a better solution for this use case?

ocefpaf:
I was unable to convert that file to a netCDF.

```
Running Compliance Checker on the datasets from: ['example.cdl']
netCDF-4 file could not be generated from cdl file with message:
Undefined name (line 144): TEMP_quality_control

Trying to create netCDF Classic file instead.
netCDF Classic file could not be generated from cdl filewith message:
Undefined name (line 144): TEMP_quality_control
```

You can zip a netcdf and attach it here. BTW, what is the error you are seeing? That will help us narrow it down.

ocefpaf:
I'm assuming the issue is related to this part in the code:

https://github.com/ioos/compliance-checker/blob/326ab907b12bb6bd3d93d284fcd021f5bee3c416/compliance_checker/cf/cf_1_6.py#L39-L46

@benjwadams is there a reason to not allow all integers types there? I could not find anything in CF 1.6 that limits this. Note that future CF implementations will inherit these values.

mphemming:
> I was unable to convert that file to a netCDF.
> 
> ```
> Running Compliance Checker on the datasets from: ['example.cdl']
> netCDF-4 file could not be generated from cdl file with message:
> Undefined name (line 144): TEMP_quality_control
> 
> Trying to create netCDF Classic file instead.
> netCDF Classic file could not be generated from cdl filewith message:
> Undefined name (line 144): TEMP_quality_control
> ```
> 
> You can zip a netcdf and attach it here. BTW, what is the error you are seeing? That will help us narrow it down.

[...]

PR #1351, opened 2026-08-25
Better subtype testing

ocefpaf:
Not ready for review yet. Should help with #1350.

### DIFF
--- a/.github/workflows/default-tests.yml
+++ b/.github/workflows/default-tests.yml
@@ -6,6 +6,7 @@ on:
 
 jobs:
   run:
+    name: "Default Tests (${{ matrix.os }} ${{ matrix.python-version }} cf_units=${{ matrix.cf_units }})"
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
--- a/.github/workflows/default-tests.yml
+++ b/.github/workflows/default-tests.yml
@@ -19,8 +19,10 @@ jobs:
             cf_units: true
           - os: windows-latest
             python-version: "3.14"
+            cf_units: true
           - os: macos-latest
             python-version: "3.14"
+            cf_units: true
       fail-fast: false
     defaults:
       run:
--- a/compliance_checker/cf/cf_1_6.py
+++ b/compliance_checker/cf/cf_1_6.py
@@ -36,15 +36,6 @@ class CF1_6Check(CFNCCheck):
     _cc_display_headers = {3: "Errors", 2: "Warnings", 1: "Info"}
     appendix_a = appendix_a_base
     appendix_d_parametric_coords = dimless_vertical_coordinates_1_6
-    _allowed_numeric_var_types = {
-        np.character,
-        np.bytes_,  # "|S1" dtype, byte array used as string
-        np.int8,
-        np.int16,
-        np.int32,
-        np.float32,
-        np.float64,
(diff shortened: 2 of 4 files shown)


ocefpaf:
@benjwadams I could not find why larger sizes, 64, 128, where forbidden. I added a subtype logic instead of checking for every variation of a specific `dtype`, and re-enabled a commented out test that should suffice.

Do you recall why we stopped at 32-bits dtype?

benjwadams:
> @benjwadams I could not find why larger sizes, 64, 128, where forbidden. I added a subtype logic instead of checking for every variation of a specific `dtype`, and re-enabled a commented out test that should suffice.
> 
> Do you recall why we stopped at 32-bits dtype?

For CF 1.6, likely because of the accepted netCDF datatypes and their corresponding numbers of bits:

https://cfconventions.org/cf-conventions/v1.6.0/cf-conventions.html#_data_types

Even 1.13 only goes up to 64 bits, whether float (double) or signed/unsigned int64:
https://cfconventions.org/Data/cf-conventions/cf-conventions-1.13/cf-conventions.html#_data_types

ocefpaf:
In this case I wonder if it was more of a limitation of the time than a standard. IMO we should be flexible here as those limitations are no longer true. What do you think? Or should we limit to the first one that accepts 64?

ocefpaf:
@benjwadams checking chapter 2.2 I can see that the text changed to include double precision, etc in CF-1.9.  However, before that nothing really states that it was not allowed, more like we could not due to netcdf own limitation.

What do think? We should implement this back to 1.6 or only 1.9 forward?

benjwadams:
Leave as is, `int` has a specific meaning in NUG, namely a signed 32-bit integer.

ocefpaf:
> Leave as is, `int` has a specific meaning in NUG, namely a signed 32-bit integer.

I understand that and they use the `NC_` prefix to distinguish that , but the "meaning" was a limitation of the time, not a standard. Reading CF I can't really interpret that it should be a limiting standard. After the CDF5 and netCDF-4 format showed up those types were allowed even in NUG:

```
Remarks
* These types are available only for CDF5 (NC_CDF5) and netCDF-4 format (NC_NETCDF4) files. All the unsigned ints and the 64-bit ints are for CDF5 or netCDF-4 files only.
+ These types are available only for netCDF-4 (NC_NETCDF4) files.
```

TL;DR if someone has netcdf-3 data, they are following NUG already b/c of the limitation, if they are on a modern netcdf-like  format, there is nothing preventing them from using these data types according to my interpretation of CF.

https://docs.unidata.ucar.edu/nug/current/md_types.html

benjwadams:
I'm not clear on the intent of the CF authors when they changed in 1.9:

https://github.com/cf-convention/cf-conventions/issues/243

CF Trac is now offline, here is the most recent archived version I could find of the link in the above GitHub issue.

https://web.archive.org/web/20210117093247/https://cf-trac.llnl.gov/trac/ticket/166

ocefpaf:
I'll ask folks at the end of month during the CF workshop. IMO they clarified an ambiguity and we should allow those dtypes in CF <1.9. If not, at least for CF-1.8 as that one is still the reference implementation (or bump the recommendation to CF-1.9 as the new reference.)

PR #1355, opened 2026-09-10
Improve testing on GHA

ocefpaf:
In this PR:
- Replaced cc6 (deprecated) with the wcrp.
- Removed unnecessary default shell from all GHA that did not require it.
- Add linter check with the safer and faster `prek` \*
- Use only wheels in tests when possible thanks tio cf-units using ABI3.
- The only limitation we have and still requires conda packages are the `ncgen` in the default tests. There I added a script to parse the dependencies from the `pyproject.toml` file to keep things in one place.

\*  @MathewBiddle can you disable pre-commit-ci for us here, or IOOS org entirely? I'll get to all repos ASAP.


.github/workflows/default-tests.yml
```diff
@@ -6,16 +6,13 @@ on:
 
 jobs:
   run:
+    name: "Default Tests (${{ matrix.os }} ${{ matrix.python-version }}"
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        python-version: [ "3.10", "3.11", "3.12", "3.13", "3.14" ]
+        python-version: [ "3.11", "3.14" ]
```
benjwadams, line 13: What's the reason 3.12 and 3.13 were removed?

benjwadams (approved):

.github/workflows/default-tests.yml
ocefpaf: Python is now on a fast pace release with much better support for stability on both pure Python and compiled extensions (ABI3). There is no need to test them all, we can get by with oldest and latest only.

PR #1360, opened 2026-09-23
Handle multiple cf_roles

benjwadams:
Handles up to two cf_roles if featureType global attribute timeSeriesProfile or trajectoryProfile.
If featureType is one of timeseries_id, profile_id, and trajectory_id, one cf_role is allowed.


benjwadams:
pre-commit.ci autofix

benjwadams:
Addresses #1284.  Interestingly this portion of CF section 9.5 does not appear in the conformance docs.

compliance_checker/cf/cf_1_6.py
```diff
@@ -3676,14 +3546,29 @@ def check_cf_role(self, ds):
...
+        max_cf_role_variable_count = {
+            "timeseries": 1,
+            "profile": 1,
+            "trajectory": 1,
+            "timeseriesprofile": 2,
+            "trajectoryprofile": 2,
+        }.get(feature_type)
+
+        # CF does not specify a maximum for point or unknown feature types.
+        if max_cf_role_variable_count is not None:
```
benjwadams: Open question: What should be done if one of the mentioned `featureType`s isn't used?
I defaulted to not checking maximum number of `cf_role`s is not present, but I don't see any particular guidance from the CF manual.

compliance_checker/cf/cf_1_11.py
```diff
@@ -172,18 +172,6 @@ def check_bounds_inherit_attributes(self, ds):
             results.append(bounds_bi_ctx.to_result())
         return results
 
-    def check_single_cf_role(self, ds):
```
benjwadams: Original limits are defined in CF 1.6 -- this deleted implementation doesn't cover the two `cf_role` case, so I moved into the CF 1.6 module.

